PR #1, opened 2014-07-04
7.0: fix most flake8 errors

gurneyalex:
I did not work on the **unported** modules
There's still an unused local variable but I don't think this is too bad.


pedrobaeza:
Hi, Alexandre,

I see very needed to rename at last tr_barcode\* family modules to barcode*, because tr_ is the preffix for the author company, which was strictly forbidden on OCA rules.

Regards.

gurneyalex:
@pedrobaeza I'm very reluctant to do so in 6.1 / 7.0 as it would break existing instances. I'm all for it in the course of the porting of these modules to 8.0. 

In any case, this is not the point of this PR (which aims at getting a green Travis status for the branch); could you please file an issue about this so that a proper migration path can be proposed to our users?

barcode_link/__init__.py
```diff
@@ -19,6 +19,6 @@
 #
 #################################################################################
 
-from . import barcode_link
+from . import barcode_link  # noqa
```
bwrsandman, line 22: With the community pep8 checks, it should not be necessary to add `# noqa` here, `__init__.py` files don't check for unused imports.

I generally disapprove of adding `# noqa`

product_multi_ean/__init__.py
```diff
@@ -19,4 +19,4 @@
 #
 ##############################################################################
 
-from . import product_ean
+from . import product_ean   # noqa
```
bwrsandman, line 22: Should not be necessary

tr_barcode/__init__.py
```diff
@@ -17,9 +17,10 @@
 #    You should have received a copy of the GNU Affero General Public License
 #    along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
-#/#############################################################################
+##############################################################################
+
+from . import tr_barcode  # noqa
+from . import res_config  # noqa
+from . import wizard      # noqa
```
bwrsandman, line 24: Should not be necessary

tr_barcode/wizard/__init__.py
```diff
@@ -19,5 +19,6 @@
 #
 ##############################################################################
 
-from . import tr_barcode_wizard
+from . import tr_barcode_wizard  # noqa
```
bwrsandman, line 22: Should not be necessary

tr_barcode_config/__init__.py
```diff
@@ -19,6 +19,6 @@
 #
 #################################################################################
 
-from . import barcode_config
+from . import barcode_config  # noqa
```
bwrsandman, line 22: Should not be necessary

tr_barcode_config/barcode/__init__.py
```diff
@@ -19,6 +19,6 @@
 #
 #################################################################################
 
-from . import barcode_osv
+from . import barcode_osv  # noqa
```
bwrsandman, line 22: Should not be necessary

tr_barcode_field/__init__.py
```diff
@@ -19,6 +19,6 @@
 #
 #################################################################################
 
-from . import barcode_field
+from . import barcode_field  # noqa
```
bwrsandman, line 22: Should not be necessary

tr_barcode_on_picking/__init__.py
```diff
@@ -19,8 +19,7 @@
 #
 #################################################################################
 
-from . import picking
-from . import res_config
+from . import picking     # noqa
+from . import res_config  # noqa
```
bwrsandman, line 23: Should not be necessary

tr_barcode_on_prodlots/__init__.py
```diff
@@ -19,8 +19,7 @@
 #
 #################################################################################
 
-from . import prodlot
-from . import res_config
+from . import prodlot      # noqa
+from . import res_config   # noqa
```
bwrsandman, line 23: Should not be necessary

tr_barcode_on_product/__init__.py
```diff
@@ -19,8 +19,7 @@
 #
 #################################################################################
 
-from . import product
-from . import res_config
+from . import product     # noqa
+from . import res_config  # noqa
```
bwrsandman, line 23: Should not be necessary

tr_barcode_on_tracking/__init__.py
```diff
@@ -19,8 +19,7 @@
 #
 #################################################################################
 
-from . import tracking
-from . import res_config
+from . import tracking    # noqa
+from . import res_config  # noqa
```
bwrsandman, line 23: Should not be necessary

tr_barcode/tr_barcode.py
```diff
@@ -96,7 +101,7 @@ def get_image(self, value, width, height, hr, code='QR'):
             fd, temp_path = mkstemp(suffix='.png')
             qrCode = QR(data=value)
             qrCode.encode(temp_path)
-            fdesc = open(qrCode.filename,"rb")
+            fdesc = open(qrCode.filename, "rb")
             data = base64.encodestring(fdesc.read())
             fdesc.close()
```
bwrsandman, line 106: It is better to use 

``` python
with open(qrCode.filename, "rb") as fdesc:
    data = base64.encodestring(fdesc.read())
```

tr_barcode/wizard/tr_barcode_wizard.py
```diff
@@ -39,10 +39,12 @@ class tr_barcode_wizard(orm.TransientModel):
     """ wizard for barcode """
     _name = "tr.barcode.wizard"
     _description = "Barcode Wizard for generic use"
+
     def _get_val(self, cr, uid, context=None):
         if not context:
             context = {}
-        if not context.get('active_model',False) or not context.get('active_id',False):
+        if not context.get('active_model', False) \
+           or not context.get('active_id', False):
```
bwrsandman, line 47: `, False` is redunant, and could be fit on one line as 

``` python
if not context.get('active_model') or not context.get('active_id'):
```

Also, note that the max line length is set to 120

tr_barcode_on_picking/picking.py
```diff
@@ -21,7 +21,8 @@
 
 from openerp.addons.tr_barcode_config.barcode import barcode_osv
 
+
 class stock_picking(barcode_osv.barcode_osv):
     _inherit = 'stock.picking'
 
-# vim:expandtab:smartindent:tabstop=4:softtabstop=4:shiftwidth=4:
\ No newline at end of file
+# vim:expandtab:smartindent:tabstop=4:softtabstop=4:shiftwidth=4:
```
bwrsandman, line 28: I would just go ahead and remove the vim line, TBH

gurneyalex:
Thanks for the review @bwrsandman. I won't be able to work on this PR until august unless something bad happens and I have to shorten my vacations. Anyone willing to enhance this is welcome to further work on this branch and propose another PR.

tr_barcode/tr_barcode.py
gurneyalex, line 106: both are strictly equivalent unless an exception is raised during read(), in which case the context manager ensures fdesc is closed, whereas the other version leaves this to the garbage collector, which may delay the closing to later in case you're running on Jython, or maybe Pypy. 

My goal in this PR is _not_ to fix all the doubious code in the addons (the use of mkstemp just above the line you mention is such an example, and it will cause a crash on Windows, I think). Just to make the minimal changes in the code for flake8 to be happy and Travis to produce green builds... and also be able to merge these changes to master and 6.1 to avoid doing this tedious work more than needed.

barcode_link/__init__.py
gurneyalex, line 22: ok I tested this with raw flake8, so this was needed to remove the warning. Please undo the change if you dislike it.

pedrobaeza:
Hi, @gurneyalex,

You're right. Let's stay as is until version 8, though I think we will find the same migration problems as in this version.

Thanks for your work. Regards.

yvaucher:
:+1:

bwrsandman:
@yvaucher there is still work to do on this PR, can you revert the merge, please

yvaucher:
I'll do the changes in a new PR

PR #6, opened 2014-11-05
Correctly use import of addons product

mdietrichc2c:
During the PEP8 validation, the import was renamed from 'product' to 'addon_product', but the call added a layer; this creates an issue "no module called 'product'" when the function _check_ean_key was called.

### DIFF
--- a/barcode_link/__init__.py
+++ b/barcode_link/__init__.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-#################################################################################
+###############################################################################
 #
 #    OpenERP, Open Source Management Solution
 #    Copyright (C) 2011 Julius Network Solutions SARL <contact@julius.fr>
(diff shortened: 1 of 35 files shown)


pedrobaeza:
I prefer to put: `from openerp.addons.product.product import check_ean`, which is clearer.

mdietrichc2c:
@pedrobaeza I actually think addon_product is more readable, as you definitely know you're calling the method from addons, instead of a local one, just by reading _check_ean_key().
@gurneyalex , do you have a preference?

pedrobaeza:
OK, I have no strong opinion about this, but in general, I don't like renaming nothing

lepistone:
I also slightly prefer the direct import without renaming like @pedrobaeza suggests, but It is a detail for me and I :+1: in any case once green. Thanks!

nbessi:
:+1:  for me too

nbessi:
I will fix the flake8 lint stuff now

nbessi:
It is green we can merge now

mdietrichc2c:
:+1: thanks @nbessi

jgrandguillaume:
:+1:

PR #8, opened 2014-11-17
First release of stock_inventory_barcode

tremlin:
The module proposed for merging adds a barcode input field to the stock inventory form. This allows to add products to the inventory list by a barcode scanner. The adding mechanism is encapsulated in an abstract class that can be used with other models.

### DIFF
--- a/stock_inventory_barcode/__init__.py
+++ b/stock_inventory_barcode/__init__.py
@@ -1,0 +1,24 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+#
+#    OpenERP, Open Source Management Solution
+#    Copyright (C) 2010-2013 OpenERP s.a. (<http://openerp.com>).
+#    Copyright (C) 2014 initOS GmbH & Co. KG (<http://www.initos.com>).
+#    Author Katja Matthes <katja.matthes at initos.com>
+#
+#    This program is free software: you can redistribute it and/or modify
+#    it under the terms of the GNU Affero General Public License as
+#    published by the Free Software Foundation, either version 3 of the
+#    License, or (at your option) any later version.
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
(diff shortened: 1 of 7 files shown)


stock_inventory_barcode/barcode_input.py
```diff
@@ -0,0 +1,103 @@
...
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+from openerp.osv import orm, fields
+from openerp.tools.translate import _
+
+
+class abtract_product_barcode_input(orm.AbstractModel):
+    _name = 'product.product.barcode.input'
```
bealdav: Maybe a good idea to have 'abstract' string in the name of the model. Thanks for this contrib

tremlin:
Sorry for my lag. I've just renamed the class to `abstract.product.barcode.input`

stock_inventory_barcode/__init__.py
```diff
@@ -0,0 +1,24 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+#
+#    OpenERP, Open Source Management Solution
+#    Copyright (C) 2010-2013 OpenERP s.a. (<http://openerp.com>).
```
alexis-via, line 5: Don't put Copyright OpenERP s.a. when it's not the case (and I don't think it is the case here !

alexis-via:
Nice initiative. I like this kind of simple and pragmatic approach.

I don't understand what is the advantage of having an abstract class... but why not ! I tried to use it, but it didn't refresh the quantity on the lines (although the code seems to work nice and the onchange returns good value).... but maybe it's me that doesn't know how to use the stock in v7 any more...

This kind of important feature should be in OCA, not outside, so :+1:

alexis-via:
Another question: I wonder how this module would work with 10k products or more (i.e. 10k inventory lines). I wonder if an on_change that returns 10k lines for the one2many field works fine...

lasley:
Is this PR abandoned?

alexis-via:
@lasley A similar module with the same name has been merged in v8, see https://github.com/OCA/stock-logistics-barcode/tree/8.0/stock_inventory_barcode

lasley:
Thanks for pointing that out @alexis-via.  

I am closing this because it has been superseded by [a module in v8](https://github.com/OCA/stock-logistics-barcode/tree/8.0/stock_inventory_barcode). Per [OCA Guidelines](https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#please-respect-a-few-basic-rules), modules may not duplicate functionality.

PR #12, opened 2015-03-03
fix travis (flake8)

gurneyalex:


lepistone:
:+1:

pedrobaeza:
```
./tr_barcode/tr_barcode.py:32:5: F821 undefined name '_logger'
```

tr_barcode/tr_barcode.py
```diff
@@ -33,6 +32,9 @@
     _logger.warning("ERROR IMPORTING REPORT LAB")
```
yvaucher, line 33: Erm here is a case for which this rule is wrong unless it is possible to define logger just before `try .. except`

pedrobaeza, line 33: Put it just right after except (duplicating it)

yvaucher, line 33: Or we could ignore E402 in such case

lepistone, line 33: no need, the try-except blocks are ignored already. Just put `_logger` after the last top-level import.

yvaucher:
:+1:

PR #17, opened 2015-09-01
8.0 product barcode generator

damdam-s:
add module  "product_barcode_generator" in this OCA project thanks to https://github.com/yelizariev and https://github.com/julius-network-solutions


lmignon:
@damdam-s Thank you for your contribution. Can you fix your module to follow the OCA guidelines? https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md (pep8, module description, ...)

damdam-s:
@lmignon of course. I'm currently working on it :)

damdam-s:
@lmignon fixed but travis doesn't launch ...

product_barcode_generator/README.rst
```diff
@@ -0,0 +1,58 @@
...
+You will have to define the company default value (6 firsts number of EAN13) then the 6 next number the sequence.
+
+The 13rd is the key of the EAN13, this will be automatically computed.
+
+Usage
+=====
+
+To use this module, you need to:
+
+* go to ...
```
lmignon: ... ???? :smirk:

product_barcode_generator/__openerp__.py
```diff
@@ -0,0 +1,42 @@
...
+
+{
+    "name": 'Product barcode generator',
+    "license": "AGPL-3",
+    "category": "Stock Management",
+    "version": '8.0.1.0.0',
+    "author": 'Julius Network Solutions, Odoo Community Association (OCA)',
+    "website": 'http://www.julius.fr/',
+    "depends": [
+        'base',
```
lmignon: `base` is not required here. You only depend of `product`

product_barcode_generator/product.py
```diff
@@ -0,0 +1,128 @@
...
+from openerp.osv import fields, orm
+from openerp.tools.translate import _
+
+
+def isodd(x):
+    return bool(x % 2)
+
+
+class product_category(orm.Model):
+    _inherit = 'product.category'
```
lmignon: IMO, new module should follow the OCA guidelines (For ex: Use CamelCase for class name) https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md

product_barcode_generator/product.py
```diff
@@ -0,0 +1,128 @@
...
+
+    _columns = {
+        'ean_sequence_id': fields.many2one('ir.sequence', 'Ean Sequence'),
```
lmignon: IMO New module  in 8.0 should use the new API...

product_barcode_generator/product.py
```diff
@@ -0,0 +1,128 @@
...
+                       [product.id],
+                       {'ean13': ean13},
+                       context=context)
+        return True
+
+    def copy(self, cr, uid, id, default=None, context=None):
+        if default is None:
+            default = {}
+        if context is None:
+            context = {}
```
lmignon: Why do you need to initialize the context?

product_barcode_generator/product.py
```diff
@@ -0,0 +1,128 @@
...
+
+    def _get_ean_key(self, code):
+        sum = 0
+        for i in range(12):
+            if isodd(i):
+                sum += 3 * int(code[i])
+            else:
+                sum += int(code[i])
+        key = (10 - sum % 10) % 10
+        return str(key)
```
lmignon: I prefer

```
return '%d' % key
```

product_barcode_generator/product.py
```diff
@@ -0,0 +1,128 @@
...
+
+    def _generate_ean13_value(self, cr, uid, product, context=None):
+        ean13 = False
+        if context is None:
```
lmignon: If you don't use the context, you don't need to initialize it.

product_barcode_generator/product.py
```diff
@@ -0,0 +1,128 @@
...
+            context = {}
+        ean = self._get_ean_next_code(cr, uid, product, context=context)
+        if not ean:
+            return None
+        key = self._get_ean_key(ean)
+        ean13 = ean + key
+        return ean13
+
+    def generate_ean13(self, cr, uid, ids, context=None):
+        if context is None:
```
lmignon: see previous comment

product_barcode_generator/product.py
```diff
@@ -0,0 +1,128 @@
...
+class product_product(orm.Model):
+    _inherit = 'product.product'
+
+    _columns = {
+        'ean_sequence_id': fields.many2one('ir.sequence', 'Ean Sequence'),
+    }
+
+    def _get_ean_next_code(self, cr, uid, product, context=None):
+        if context is None:
+            context = {}
```
lmignon: If you don't use the context, you don't need to initialize it.

product_barcode_generator/product.py
```diff
@@ -0,0 +1,128 @@
...
+        sequence_obj = self.pool.get('ir.sequence')
```
lmignon:
```
self.pool['ir.sequence']
```

product_barcode_generator/product.py
```diff
@@ -0,0 +1,128 @@
...
+        ean = ''
+        if product.ean_sequence_id:
```
lmignon: You should explain the logic below in the module description.

product_barcode_generator/product_view.xml
```diff
@@ -0,0 +1,37 @@
...
+        
+		<record id="view_product_simple_ean_form" model="ir.ui.view">
+			<field name="name">product.product.form</field>
+			<field name="model">product.product</field>
+			<field name="inherit_id" ref="product.product_normal_form_view"/>
+			<field name="arch" type="xml">
+				<field name="ean13" position="after">
+                    <field name="ean_sequence_id" domain="[('barcode_sequence', '=', 1)]"
+                        context="{'default_name': name and 'EAN ' + name or 'EAN', 'default_barcode_sequence': 1}"
+                        class="oe_edit_only"/>
```
lmignon: Why oe_edit_only? it's strange having to enter in edit mode to see that the product is linked to a specific sequence.

lmignon:
@damdam-s Thank you for the contribution. IMO, the description should be improved to explain that the sequence to use to generate the ean13 can be specified at 3 levels: (product, category, company).
When testing the module on runbot I've been disappointed to not see the 'generate ean' button on the product form in 'Warehouse -> Products -> Products'. Indeed, the button is only available on the view of product variants  (Warehouse -> Configuration -> Products -> Product Variants) and not on the product template.
IMO, the main issues are:
- the module should use the new api,
- the description must be improved.
- If you edit a product with an ean13 and you change the sequence the button stay hidden.

damdam-s:
@lmignon Thank you for the review. I agree with you for the 2 first points (new api and description's improvement).
However for the third point, let me explain what the method which generates ean13 work. If there is already an ean13 on the product, the method doesn't change it. So if you want to have a new ean13, you can manually empty the field and the button will appears.
I can change this behaviour but i'm not the original author of this module, i just had the permission from the real authors to push it in OCA repository. So maybe we may not have to change the behaviour from previous version (even if those versions are not hosted in OCA repository) that can be used by other people ? 
I understand what you mean in your comment and understand it can be confusing when the sequence changes.

Do you think that I should empty ean13 field when sequence changes ?
I'm open for further discussion.

lmignon:
@damdam-s an other solution is to avoid to display the sequence if ean13 is not empty.

damdam-s:
@lmignon thanks for this simple solution. fixed

product_barcode_generator/models/product.py
```diff
@@ -0,0 +1,104 @@
...
+        if not ean:
+            return None
+        key = self._get_ean_key(ean)
+        ean13 = ean + key
+        return ean13
+
+    @api.multi
+    def generate_ean13(self):
+        product_ids = self
+        for product in product_ids:
```
lmignon: for product in self:

product_barcode_generator/models/product.py
```diff
@@ -0,0 +1,104 @@
...
+
+    ean_sequence_id = fields.Many2one('ir.sequence', string='Ean sequence')
+
+
+class ProductProduct(models.Model):
+    _inherit = 'product.product'
+
+    ean_sequence_id = fields.Many2one('ir.sequence', string='Ean sequence')
+
+    @api.model
```
lmignon, line 42: Why not 

```
@api.one
def _get_ean_next_code(self):
```

??

product_barcode_generator/models/product.py
```diff
@@ -0,0 +1,104 @@
...
+        sum = 0
+        for i in range(12):
+            if isodd(i):
+                sum += 3 * int(code[i])
+            else:
+                sum += int(code[i])
+        key = (10 - sum % 10) % 10
+        return '%d' % key
+
+    @api.model
```
lmignon, line 79: `@api.one`???

lmignon:
@damdam-s Thank you for all your changes and the unittests! Your module with their tests puts the initial coverage value for stock-logistics-barcode very high. Let's see what others think about my 2 last comments since I can't give you a strong motivation to prefer `api.one` in place of `api.model` for these 2 methods.

product_barcode_generator/README.rst
```diff
@@ -0,0 +1,59 @@
...
+
+=========================
+Product barcode generator
+=========================
+
+This module will add a function which leads to an automatic generation of EAN13 for products
+
+You will have to define the company default value (6 firsts number of EAN13) then the 6 next number the sequence.
+
+The 13rd is the key of the EAN13, this will be automatically computed.
```
pedrobaeza: s/key/control digit

product_barcode_generator/models/product.py
gurneyalex, line 79: why ? api.one does not mandate that there is a single record in self, it loops on all elements in self and accumulates the results in a list. Using api.one here would change the return value of the method.

product_barcode_generator/models/product.py
pedrobaeza, line 42: It can't be `@api.one` if there's a return. It should be `@api.multi` with `self.ensure_one()` sentence and taking self as the product.

product_barcode_generator/models/product.py
```diff
@@ -0,0 +1,103 @@
...
+    _inherit = 'product.product'
+
+    ean_sequence_id = fields.Many2one('ir.sequence', string='Ean sequence')
+
+    @api.model
+    def _get_ean_next_code(self, product):
+        sequence_obj = self.env['ir.sequence']
+        ean = ''
+        if product.ean_sequence_id:
+            ean = sequence_obj.next_by_id(product.ean_sequence_id.id)
```
pedrobaeza: s/ean/ean_seq for clarity in code

product_barcode_generator/models/product.py
gurneyalex, line 42: @lmignon this changes the return type (see my other comment below)

product_barcode_generator/models/product.py
```diff
@@ -0,0 +1,103 @@
...
+
+class ProductProduct(models.Model):
+    _inherit = 'product.product'
+
+    ean_sequence_id = fields.Many2one('ir.sequence', string='Ean sequence')
+
+    @api.model
+    def _get_ean_next_code(self, product):
+        sequence_obj = self.env['ir.sequence']
+        ean = ''
```
pedrobaeza: This initialization does nothing

product_barcode_generator/models/product.py
```diff
@@ -0,0 +1,103 @@
...
+        elif product.company_id and product.company_id.ean_sequence_id:
+            ean = sequence_obj.next_by_id(
+                product.company_id.ean_sequence_id.id)
+        elif self.env.context.get('sequence_id', False):
+            ean = sequence_obj.next_by_id(self.env.context.get('sequence_id'))
+        else:
+            return None
+        if len(ean) > 12:
+            raise exceptions.Warning(
+                _("Configuration Error!"),
```
pedrobaeza: Warning only takes one argument

product_barcode_generator/models/product.py
```diff
@@ -0,0 +1,103 @@
...
+                _("There next sequence is upper than 12 characters. "
```
pedrobaeza: s/There/The

product_barcode_generator/models/product.py
```diff
@@ -0,0 +1,103 @@
...
+                  "This can't work. "
+                  "You will have to redefine the sequence or create a new one")
+                )
+        else:
+            ean = (len(ean[0:6]) == 6 and ean[0:6] or
+                   ean[0:6].ljust(6, '0')) + ean[6:].rjust(6, '0')
+        return ean
+
+    def _get_ean_key(self, code):
```
pedrobaeza: s/key/control_digit

product_barcode_generator/models/product.py
pedrobaeza, line 79: See my comment in https://github.com/OCA/stock-logistics-barcode/pull/17/files#r38534015

lmignon, line 79: @gurneyalex You'r right. I had forgotten the problem with the return values When a method is decorated with api.one.

product_barcode_generator/models/product.py
```diff
@@ -0,0 +1,103 @@
...
+                continue
+            ean13 = self._generate_ean13_value(product)
+            if not ean13:
+                continue
+            product.write({'ean13': ean13})
+        return True
+
+    @api.one
+    def copy(self, default=None):
+        if default is None:
```
pedrobaeza: For getting an even better coverage, put this in one line: `default = default if default is not None else {}`

product_barcode_generator/models/product.py
```diff
@@ -0,0 +1,103 @@
...
+    @api.model
+    def _generate_ean13_value(self, product):
+        ean = self._get_ean_next_code(product)
+        if not ean:
+            return None
+        key = self._get_ean_key(ean)
+        ean13 = ean + key
+        return ean13
+
+    @api.multi
```
pedrobaeza: Here you can put @api.one and save a loop

product_barcode_generator/models/product.py
```diff
@@ -0,0 +1,103 @@
...
+        else:
+            return None
+        if len(ean) > 12:
+            raise exceptions.Warning(
+                _("Configuration Error!"),
+                _("There next sequence is upper than 12 characters. "
+                  "This can't work. "
+                  "You will have to redefine the sequence or create a new one")
+                )
+        else:
```
pedrobaeza: There's no need for else sentence

pedrobaeza: I mean, put directly the content of the else at a higher level

pedrobaeza:
Remove product_barcode_generator/static/src/img/icon.png

product_barcode_generator/models/product.py
```diff
@@ -0,0 +1,103 @@
...
+            if product.ean13:
+                continue
+            ean13 = self._generate_ean13_value(product)
+            if not ean13:
+                continue
+            product.write({'ean13': ean13})
+        return True
+
+    @api.one
+    def copy(self, default=None):
```
lmignon: Can we achieve the same result by redefining the field ean13 with the new api?

``` python
ean13 = fields.Integer(copy=False)
```

pedrobaeza:
Why don't call _product_ean13_generator_?

damdam-s:
@pedrobaeza this module was created by Julius Network (https://github.com/julius-network-solutions/julius-openobject-addons/tree/master/product_barcode_generator) and @yelizariev (https://github.com/yelizariev/pos-addons/tree/8.0/product_barcode_generator) with the name `product_barcode_generator`. Thanks to them we can have this module in OCA because I have their permission. I think we could keep the name they gave to this module ?

lmignon:
:+1:  good job @damdam-s and thank you also to the original authors.

pedrobaeza:
I prefer to change it to reflect better its use and also to differentiate it from the original one.

gurneyalex:
:+1:

gurneyalex:
regarding the rename, since this module is going to be removed from Julius' repository, I'm _not_ favorable as this will create lots of trouble for the existing user base.

pedrobaeza:
I'm not sure they will remove it and there are 2 modules with the same name. As I said, I prefer to rename it for not confusing the user which one to use. You can rename easily the module in the existing installations with the openupgradelib.

product_barcode_generator/README.rst
```diff
@@ -0,0 +1,59 @@
...
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+=========================
+Product barcode generator
+=========================
+
+This module will add a function which leads to an automatic generation of EAN13 for products
+
+You will have to define the company default value (6 firsts number of EAN13) then the 6 next number the sequence.
```
yvaucher: s/6 firsts number/first 6 numbers/
s/6 next number/next 6 numbers/

**of** the sequence?

product_barcode_generator/README.rst
```diff
@@ -0,0 +1,59 @@
...
+
+The 13rd is the control digit of the EAN13, this will be automatically computed.
+
+The sequence to use to generate the ean13 can be specified at 3 levels (in the specified order):
+* product
```
yvaucher, line 17: Add a blank line before a list

product_barcode_generator/__openerp__.py
```diff
@@ -0,0 +1,41 @@
...
+#    GNU General Public License for more details.
+#
+#    You should have received a copy of the GNU General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+###############################################################################
+
+{
+    "name": 'Product barcode generator',
+    "license": "AGPL-3",
```
yvaucher: Please change this to GPL-3 to respect license set by Julius Network

product_barcode_generator/README.rst
```diff
@@ -0,0 +1,59 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
```
yvaucher: ~~Julius module was GPL-3, please change the license logo accordingly~~

guewen:
> You can rename easily the module in the existing installations with the openupgradelib.

We don't have the same idea of "easily". I'm not favorable for the renaming neither.

product_barcode_generator/models/product.py
```diff
@@ -0,0 +1,95 @@
...
+        elif self.env.context.get('sequence_id', False):
+            ean = sequence_obj.next_by_id(self.env.context.get('sequence_id'))
+        else:
+            return None
+        ean = (len(ean[0:6]) == 6 and ean[0:6] or
+               ean[0:6].ljust(6, '0')) + ean[6:].rjust(6, '0')
+        if len(ean) > 12:
+            raise exceptions.Warning(
+                _("Configuration Error!"
+                  "The next sequence is upper than 12 characters. "
```
yvaucher: `upper than` to replace. Upper means higher you might want to use `longer than`

pedrobaeza:
Sorry, but I don't agree to put a module in OCA with a non-proper name if it comes from outside. As I have to deal with that in my modules coming from OdooMRP, you also should.

product_barcode_generator/models/product.py
```diff
@@ -0,0 +1,95 @@
...
+                  "This can't work. "
```
yvaucher: We could be more explicit saying, `It is not valid for an EAN13 needing 12 characters, the 13 being used as a control digit`

guewen:
On what basis do you claim it is a _non-proper name_? It describes what it does and does not contains a company name or whatever which would be considered invalid. Also, I don't see any other module with this name as you pretend.

product_barcode_generator/tests/test_product_barcode_generator.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+class TestProductBarcodeGenerator(common.TransactionCase):
+
+    def setUp(self):
+        super(TestProductBarcodeGenerator, self).setUp()
+        self.sequence1 = self.env.ref(
+            'product_barcode_generator.seq_ean13_sequence')
+        if not self.sequence1.barcode_sequence:
+            self.sequence1.write({'barcode_sequence': True})
```
yvaucher: You can write

```
self.sequence1.barcode_sequence = True
```

pedrobaeza:
It only generates EAN13, not other barcodes. That's my claim.

product_barcode_generator/tests/test_sequence_configured_on_categ.py
```diff
@@ -0,0 +1,47 @@
...
+        self.product_demo.write({'ean13': False})
+        self.product_demo.generate_ean13()
+        self.assertEqual(bool(self.product_demo.ean13), False)
+
+    def test_generate_ean_sequence_on_categ_with_sequence(self):
+        self.product_demo.write({'ean13': False})
+        self.product_demo.categ_id.write(
+            {'ean_sequence_id': self.sequence1.id})
+        self.product_demo.generate_ean13()
+        self.assertEqual(bool(self.product_demo.ean13), True)
```
yvaucher: assertTrue ?

product_barcode_generator/tests/test_sequence_configured_on_categ.py
```diff
@@ -0,0 +1,47 @@
...
+        if not self.sequence1.barcode_sequence:
+            self.sequence1.write({'barcode_sequence': True})
+
+        self.product_demo = self.env.ref('product.product_product_6')
+        self.product_obj = self.env['product.product']
+
+    def test_generate_ean_sequence_on_categ_no_sequence(self):
+        self.product_demo.write({'ean13': False})
+        self.product_demo.generate_ean13()
+        self.assertEqual(bool(self.product_demo.ean13), False)
```
yvaucher: assert False ?

product_barcode_generator/views/product_view.xml
```diff
@@ -0,0 +1,38 @@
...
+            <field name="arch" type="xml">
+                <field name="type" position="after">
+                    <field name="ean_sequence_id" domain="[('barcode_sequence', '=', 1)]"
+                        context="{'default_name': name and 'EAN ' + name or 'EAN', 'default_barcode_sequence': 1}"/>
+                    <newline/>
+                </field>
+            </field>
+        </record>
+        
+		<record id="view_product_simple_ean_form" model="ir.ui.view">
```
yvaucher: beware of the tabs

please replace them by spaces

yvaucher:
@damdam-s almost good some nitpicking to fix

yvaucher:
@juliusfr and @mathieuvatel

Hello there the juluis team, just a ping to let you know your module is going into OCA and I have 2 questions for you.
1. Are you willing that we rename it on OCA not to conflict with you own or is it ok to have it on OCA?
2. License is GPL3 you want to keep it that way or you prefer AGPL (I believe this was licensed GPL for historical reasons) we can let it unchanged and it will be perfectly good.

product_barcode_generator/README.rst
yvaucher: Forget this, with your tests in AGPL I guess it makes the module AGPL

yvaucher:
And one more thing: translations are missing

yvaucher:
@juliusfr ping (did a typo in preview comment)

damdam-s:
@yvaucher fixed. I changed licence to GPL3 while waiting an answer from @juliusfr and/or @mathieuvatel

mathieuvatel:
@yvaucher GPL3 is ok for us.
For naming, we will remove it from our repository for odoo 8.0 and future version

gurneyalex:
@mathieuvatel do you object to the module being renamed in the OCA?

mathieuvatel:
@gurneyalex  no you can do what you want for the naming of this module. Just tell me the new name please.

damdam-s:
So what about the module name ? Do we change it as suggested by @pedrobaeza ?
And what about the licence GPL3 or AGPL3 ? cc @yvaucher

damdam-s:
up

yvaucher:
@damdam-s What I understand is that we can keep GPL3 and the same module name so :+1:

PR #24, opened 2015-10-27
refactor generate_ean13: do not return None

lepistone:
Those "return" used to be "continue" inside a loop. Now it looks better
to have one return in the end. We want to return True because that's
what odoo expects from an action method.


pedrobaeza:
:+1:

damdam-s:
:+1:

rlizana:
:+1:

rvalyi:
:+1:

PR #26, opened 2015-12-23
8.0 add module stock_inventory_barcode

alexis-via:
See the module description for more info.

This is a REWRITE of the module proposed by initOS for v7.0 in this PR: https://github.com/OCA/stock-logistics-barcode/pull/8 (PR dated november 2014 and not merged as of today)
I rewrote this module instead of porting it to v8 for 2 reasons:
1) I think it's not a good idea to update the qty via on on_change that updates the inventory lines... as you can have several thousands inventory lines and I fear it would have bad perfs
2) You should be able to enter the new quantity with the keyboard and not just scan each item of the same product one by one (which can take a lot of time if you have a big qty of items, instead of just counting the items and entering the qty)

I kept the name of the module because it has exactly the same goal and the same use case. If you think it's a problem to keep the same name for the module, I'm OK to change the name of this module.

### DIFF
--- a/stock_inventory_barcode/README.rst
+++ b/stock_inventory_barcode/README.rst
@@ -1,0 +1,88 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+=======================
+Stock Inventory Barcode
+=======================
+
+The interface provided by Odoo in the official *stock* module to make an
+inventory has 2 limitations:
+
+* it is not easy/fast to use when you have several thousand products
+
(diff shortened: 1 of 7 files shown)


brother-bernard:
:+1: using this code in production

tremlin:
I think by the newest code guidelines the version should be `8.0.1.0.0` instead of `8.0.0.1.0` and short license headers should be used. It is up to the one who could merge this if a change is necessary here. Code review otherwise :+1:

gurneyalex:
@alexis-via @tremlin how difficult would it be to migrate from the implementation in #8 to this one for someone moving from 7.0 to 8.0?

gurneyalex:
Sorry this was a stupid question (only transient models, and I had 8 -> 9 in mind, not 7 -> 8). 

:+1:

PR #28, opened 2016-03-15
[8.0] [ADD] product_multi_ean

pedrobaeza:
# 
# Multiple EAN13 on products

It allows to have multiple EAN13 on products.
# Usage

A list of EAN13 is available for each product with a priority, so a main EAN13
code is defined.


pedrobaeza:
@andrius-preimantas, here you have the module with tests.

product_multi_ean/models/product_product.py
```diff
@@ -0,0 +1,85 @@
...
+    @api.multi
+    @api.constrains('name')
+    def _check_duplicates(self):
+        for record in self:
+            eans = self.search(
+                [('id', '!=', record.id), ('name', '=', record.name)])
+            if eans:
+                raise UserError(
+                    _('The EAN13 Barcode "%s" already exists for product '
+                      '"%s"!') % (record.name, eans[0].product_id.name))
```
guewen: don't need to be so exclamative :)

product_multi_ean/models/product_product.py
```diff
@@ -0,0 +1,85 @@
...
+from openerp.addons.product import product as addons_product
+
+
+class ProductEan13(models.Model):
+    _name = 'product.ean13'
+    _description = "List of EAN13 for a product."
+    _order = 'sequence'
+
+    name = fields.Char(string='EAN13', size=13, required=True)
+    sequence = fields.Integer(string='Sequence', default=0)
```
guewen, line 18: There was a default value for the sequence that automatically increment it, I think it was necessary for users to keep a "main" ean13, if they all have a default value of 0, the ean13 may change haphazardly (or at least change the `_order` to `sequence, id`).

product_multi_ean/models/product_product.py
```diff
@@ -0,0 +1,85 @@
...
+        return {
+            'product_id': self.id,
+            'name': self.ean13,
+        }
+
+    @api.model
+    def search(self, domain, *args, **kwargs):
+        if filter(lambda x: x[0] == 'ean13', domain):
+            ean_operator = filter(lambda x: x[0] == 'ean13', domain)[0][1]
+            ean_value = filter(lambda x: x[0] == 'ean13', domain)[0][2]
```
guewen, line 80: A `for` loop would be more readable and more efficient instead of repeating a `filter` (yes I know it was done like that before the migration, so if you don't want to change, I won't mind).

product_multi_ean/models/product_product.py
andrius-preimantas: :-1: 
I'd argue being explicit here is an advantage. Imagine importing your barcodes from csv and receiving error message like "EAN13 Barcode already exists!". Then you'd have to manually find duplicate EANs in your file which is a bummer :-)

guewen:
Thanks for the migration and for adding tests!

product_multi_ean/models/product_product.py
guewen: > I'd argue being explicit here is an advantage. Imagine importing your barcodes from csv and receiving error message like "EAN13 Barcode already exists!". Then you'd have to manually find duplicate EANs in your file which is a bummer :-)

That's not my point, no need to put an exclamation mark in an error message, that's all :smile:  (it's a bit like the machine yelling at you "you are an idiot!")

andrius-preimantas: My bad :-)

pedrobaeza:
HI, @guewen, thanks for the review. I have changed the test and added the order as sequence. About the search, I agree it's not the more readable code, but I kept as is before and test ensures that it works.

guewen:
thanks! :+1:

tremlin:
I reviewed the code :+1: We've also got it running in out test environment for some weeks (with deactivated unique constraint due to not well-maintained customer data). Thanks for the migration.

PR #32, opened 2016-07-17
[PORT] Migrate product_barcode_generator to 9.0

yelizariev:


product_barcode_generator/views/res_company_view.xml
```diff
@@ -7,7 +7,7 @@
 			<field name="model">res.company</field>
 			<field name="inherit_id" ref="base.view_company_form"/>
 			<field name="arch" type="xml">
-				<page string="Configuration" position="inside">
+				<page name="configuration" position="inside">
```
pedrobaeza: Don't remove string attribute

product_barcode_generator/__openerp__.py
```diff
@@ -21,10 +21,10 @@
 
 {
     "name": 'Product barcode generator',
-    "license": "GPL-3",
+    "license": "AGPL-3",
```
pedrobaeza: You should check if the license can be changed

product_barcode_generator/data/ean_sequence.xml
```diff
@@ -3,13 +3,13 @@
     <data noupdate="1">
     
         <!-- Product Sequences -->
-        <record id="seq_type_ean13_sequence" model="ir.sequence.type">
-            <field name="name">EAN13 code</field>
-            <field name="code">product.ean13.code</field>
-        </record>
+        <!--<record id="seq_type_ean13_sequence" model="ir.sequence.type">-->
```
pedrobaeza: Remove this instead of commenting

product_barcode_generator/models/product.py
```diff
@@ -36,21 +36,20 @@ class ProductCategory(models.Model):
 class ProductProduct(models.Model):
     _inherit = 'product.product'
 
-    ean13 = fields.Char(copy=False)
     ean_sequence_id = fields.Many2one('ir.sequence', string='Ean sequence')
```
pedrobaeza, line 39: What's the meaning of the sequence here? Shouldn't be enough at company level (although I miss the possibility of putting more than one sequence there)?

product_barcode_generator/views/res_company_view.xml
yelizariev: Why? It's used for inheritance, the view doesn't add new Page

yelizariev: You know, that it's impossible to inherit via string in 9.0

product_barcode_generator/models/product.py
yelizariev, line 39: Check _get_ean_next_code -- it allows to use different sequence for the product. But it's strange to have a separate sequence on a product level. 
Anyway, this PR is intended to port module rather than improve its functionality

pedrobaeza, line 39: Each time we migrate a module, we can (and should) improve the functionality, because this is the best chance of rethinking things.

product_barcode_generator/views/res_company_view.xml
pedrobaeza: Sorry, I though it was a new page.  Sunday mistakes...

product_barcode_generator/models/product.py
yelizariev, line 39: I still believe that OCA should speed PR merging up someway. One of ideas is creating new issues instead of waiting until every concern is resolved

yelizariev:
Why does runbot make build without tests?
@gurneyalex

pedrobaeza:
Runbot is not executing tests because we have already Travis for that, and this way, we make the builds lighter, faster, and cheaper (Travis infrastructure is free, Runbot one is not).

moylop260:
Could you check the lint build of travis?
https://travis-ci.org/OCA/stock-logistics-barcode/jobs/145529529

lasley:
@yelizariev - Can you take a look at the Travis build please? I'm 👍  other than that.

yelizariev:
I need advice how to fix long line error in author attribute:

```
"author": 'IT-Projects LLC, Julius Network Solutions, Odoo Community Association (OCA)',
```

travis:

```
 product_barcode_generator/__openerp__.py:27:80: E501 line too long (92 > 79 characters)
```

pedrobaeza:
Put:

```
    "author": 'IT-Projects LLC, '
              'Julius Network Solutions, '
              'Odoo Community Association (OCA)',
```

yelizariev:
@pedrobaeza thank you!

product_barcode_generator/models/product.py
pedrobaeza: OK, let's stay as is for now.

product_barcode_generator/views/res_company_view.xml
```diff
@@ -2,21 +2,21 @@
 <openerp>
```
pedrobaeza: Put `<odoo>` and remove `<data>` tag

yelizariev: Can someone make a script and fix modules (in all repositories) at once, instead of manual fixing each time?

lasley (approved):

lasley:
Thanks @yelizariev

legalsylvain:
Hi @yelizariev. 

I just discovered this module you're porting. I realized recently an other module, named 'barcodes_generator_product' (and the same for partners). See #49.
I didn't know the module product_barcode_generator because when I began to make my module, I look for an existing OCA module but based on barcode.rule. (on odoo-code-search). And 'product_barcode_generator' do not extend barcode.rule. The module provides abstract feature to generate barcodes for any models. (product and partner are implemented for the time being, but we can imagine to generate barcodes for stock location, for stock picking, ...)
I took a look on this module and my PoV is that barcodes_generator_product provides same features, but in a more generic way. (barcode rule pattern).
I suppose that barcodes_generator_product could supersed product_barcode_generator for version 9.0 +, but I'd like to have your PoV about that point. Do the new module feat with your needs ? Maybe I missed some features that are not in the module you're porting. If yes, feel free to ask improvment on my PR.

Kind regards.

yelizariev:
@legalsylvain I am far from this module and not able to compare it with yours. Let me rely on your opinion at this moment. We would return to this module or suggest improvements for your modules in the future if needed.

rafaelbn:
Hi,

We would like yo use this module https://github.com/OCA/stock-logistics-barcode/tree/9.0 or this PR for EAN-13. @yelizariev did you check if this PR is duplicated? If so maybe is better to close this PR

Let me know please

Thanks!

legalsylvain:
Hi. Well, it's not decided for the time being, but I propose to replace this module (product_barcode_generator) by barcode_generator_product.
- barcode_generator_product is available in 9.0 [link](https://github.com/OCA/stock-logistics-barcode/tree/9.0/barcodes_generator_product)
- barcode_generator_product is being ported in 10.0 [PR](https://github.com/OCA/stock-logistics-barcode/pull/66#issuecomment-272137663) by @lasley 
- I'll port this module for V8.0 in a few weeks.

From my PoV, the new module is more generic, allow more features. But as I wrote it, I would prefer to let community compare and accept my proposal.

Any external point of view is so welcome. @rafaelbn,  could you benchmark ?

Kind regards.

lasley:
My thoughts point towards the abstract module too. Having already been able to add a new barcode type to an entity, I think this is a total win 👍

legalsylvain:
Ok. I propose to close this PR. If you're agree, please accept too this #74. (and #73).

kind regards.

pedrobaeza:
Closing this one as deprecated and following the other approach. Sorry for the time lost, @yelizariev, but we all think than the new approach is more consistent with Odoo itself than to continue with this one.

PR #34, opened 2016-07-22
[8.0][ADD][stock_disable_barcode_interface] Disable and enable the barcode interface

tarteo:
This module makes it possible to disable and enable (configurable) the barcode interface system wide. Maybe it's a good idea to also add this configuration on operation (stock.picking.type) but I need someone to confirm this.

I'm not sure if this is the right repository.

### DIFF
--- a/stock_disable_barcode_interface/README.rst
+++ b/stock_disable_barcode_interface/README.rst
@@ -1,0 +1,59 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+=========================
+Disable Barcode Interface
+=========================
+
+This module makes it possible to disable and enable (configurable) the barcode interface.
+
+Usage
+=====
+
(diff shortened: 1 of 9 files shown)


lasley:
@tarteo - Thanks for the submission! This is the best repo I can think of for it as well.

👍  but would you mind squashing your commits into one please?

lasley:
Also regarding adding it on the picking type, you're meaning instead of a global config to allow a configuration for whether the interface is activated/deactivated at the picking type instead? 

I'm typically all for granularity, but I can't personally think of a use case. As I'm thinking about it, making this a user permission may be the best bet if we want to get granular.

tarteo:
@lasley Thanks for your comment. I've squashed the commits together.

lasley:
Thanks @tarteo - 👍

susport (commented):
I am trying the runbot but the admin user does not see the check Enable barcode interface.I imagine it's because of the permission of group group_enable_barcode. Thank

tarteo:
@susport It's the last checkbox of Settings -> Configuration -> Warehouse
checkbox

lasley (approved):

pedrobaeza (approved):

PR #36, opened 2016-08-16
[ADD] base_gtin: Support abstract EAN14 verification

lasley:
This module allows for barcode verifications using an abstract model + adds EAN14 (GTIN14) validation in the v9 barcode rule system.

I will also be creating a subsequent module based on this to deprecate https://github.com/OCA/product-attribute/tree/9.0/product_gtin

Still left to do:
- [x] JS implementation

ReadMe: 
# 
# Base GTIN

This module provides core logic for handling of GTIN barcode nomenclatures
# Configuration

To configure this module, you need to:
# . Set a default barcode type in Company => Configuration
# Known issues / Roadmap
- EAN-11 was half-way implemented in original `product_gtin`, but it is not implemented
  here due to lack of documentation corresponding to validations & test samples


lmignon:
@lasley In your description you plan to deprecate the `product_gtin` addon in favor of your refactored one. In your last commit you change the licence from AGPL to LGPL. Therefore I'll be against the deprecation of the original `product_gtin` and the inclusion of yours in place of the original one in OCA.

base_gtin/static/src/js/barcode_parser.js
```diff
@@ -0,0 +1,56 @@
...
+    var BarcodeParser = require('barcodes.BarcodeParser');
+    
+    BarcodeParser.include({
+        
+        /* It performs EAN-14 checksum operation
+         * :param barcode: Barcode string to validate, including check digit
+         * :return: ``int`` representing calculated checksum
+         **/
+        ean14_checksum: function(barcode) {
+            if (barcode.length !== 14) {
```
lmignon, line 16: We should avoid to duplicate the same logic in JS and Python. IMO, we should call the method provided by `barcode.nomenclature` model via XMLRPC to do the validation.

lasley, line 16: This is reimplemented in js because [barcodes also reimplements](https://github.com/odoo/odoo/blob/9.0/addons/barcodes/static/src/js/barcode_parser.js#L44). Using XMLRPC would disallow the use of this code in systems that are offline, such as POS.

lasley:
@lmignon this is a complete rewrite- none of the code in `product_gtin` was salvageable for v9. Odoo SA basically deprecated `product_gtin` themselves with `barcodes`, outside of the if statement allowing ean14 as a valid option. There was not even ean14 validation included in the original `product_gtin`, it was simply a passthru method allowing any barcode.

I don't mind releasing under AGPL, but I'm somewhat confused as to why this is a problem if I wrote it all.

lasley:
@lmignon I updated the license to AGPL, but I would still like to know the reasoning behind this please. Just trying the understand why it makes a difference, as LGPL provides more rights to the user & all derivatives already licensed as AGPL are compatible from what I understand. With this being my code, I was under the impression I was able to choose my license.

lmignon:
@lasley https://github.com/OCA/product-attribute/blob/8.0/product_gtin provides the same functionality. Even if it's no more usable in v9 the fact is that this functionality is already covered by an OCA module. You are right that he ean14 is not supported by `product_gtin`but a placeholder already exists to add the logic and the original one can be easily extended to support this new kind of ean. https://github.com/OCA/product-attribute/blob/9.0/product_gtin/product_gtin.py#L129
You describe your module as a replacement of `product_gtin`that provides the basis to validate eans. Therefore, If I need to migrate one of my clients using `product_gtin` to V9 I should use your module and reimplement the same logic as the one already implemented into the AGPL `product_gtin` IOW relicence exising code if your module is released under LGPL.
OTH if your module becomes the replacement of `product_gtin` in V9 and is released under LGPL it's no more possible to provide the same functionalities by migrating `product_gtin`to v9 since both provides the same functionalities.

lasley:
Understood, thank you for the explanation @lmignon - I just like to know why I'm making changes so that I can try to avoid mistakes in the future 😄 

Sweet so now that is handled, this is good for review I do believe.

PR #38, opened 2016-08-23
[9.0][ADD] web_stock_picking: Mobile & barcode compliant stock picking interface

lasley:
Moving from OCA/web#397

I'm currently toying with the idea of breaking this into two modules:
- web_stock - Base stock interface for mobile
- web_stock_picking - Stock picking functionality for it

This would also allow a `web_stock_inventory` and possibly others. 
# 
# Web Stock Picking

This module provides a web front end for stock pickings.

It is designed to be particularly useful with a mobile device and barcode
scanner.
# Installation
- Install Module
# Configuration
- You must set a Barcode Nomenclature on any Stock Picking Type that will
  utilize the barcode integration
# Usage
- Go to Inventory => Mobile/Barcode
- Choose a picking, or refine search
- Utilize picking form in similar fashion as backend form

Picking form also contains a barcode listener interface, and will increment/decrement
quantities based on scanned barcodes. It currently supports the following nomenclature
types:
- Product
- Lot

.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
   :alt: Try me on Runbot
   :target: https://runbot.odoo-community.org/runbot/154/9.0
# Known issues / Roadmap
- Support for lot scanning is at the unit level only. The lots themselves are not bound
  to the picking in any way
- Implement proper errors instead of alerts - requires keybindings for enter/cancel
- Implement commented out view actions
- It would be nice to also allow for a BarcodeScanner device, instead of just the
  element events
- Button injection in stock.picking.type Kanban fails silently (have also tried jQuery
  template injection)
# Bug Tracker

Bugs are tracked on `GitHub Issues
<https://github.com/OCA/{project_repo}/issues>`_. In case of trouble, please
check there if your issue has already been reported. If you spotted it first,
help us smashing it by providing a detailed and welcomed feedback.
# Credits
## Images
- Odoo Community Association: `Icon <https://github.com/OCA/maintainer-tools/blob/master/template/module/static/description/icon.svg>`_.
## Contributors
- Dave Lasley dave@laslabs.com
## Maintainer

.. image:: https://odoo-community.org/logo.png
   :alt: Odoo Community Association
   :target: https://odoo-community.org

This module is maintained by the OCA.

OCA, or the Odoo Community Association, is a nonprofit organization whose
mission is to support the collaborative development of Odoo features and
promote its widespread use.

To contribute to this module, please visit https://odoo-community.org.


lmignon:
@lasley Do you know this one? https://github.com/OCA/stock-logistics-workflow/tree/8.0/stock_scanner

lasley:
Interesting, I did not! Am I correct in my understanding that this is utilizing the pre-existing hardware proxy, but centralizing and providing workflows based upon it? 

Seems like that's going to help a lot both here and in OCA/server-tools#514 - thanks @lmignon !

lmignon:
@lasley stock_scanner comes with a python script 'https://github.com/OCA/stock-logistics-workflow/blob/8.0/stock_scanner/hardware/sentinel.py' that plays the role of an ncurses client. Once started the terminal based UI is generated by the server based on the scenarii definitions. We use this client with  the motorola MC3x00 from which we connect to the server via SSH. A great improvement could be to provide a new client/web controller that generates the UI in the same way as the sentinel script but in html5... The approach is really flexible and powerful.

lasley:
Interesting - I have a client that is looking at a device similar to an MC3 series, just Android based.

Ok so this has me intrigued and perplexed. I'm going to study the code this week and formulate a real response 😄

gurneyalex:
@lasley can you rebase? I've updated the travis config, which should allow travis2docker to work.

lasley:
@gurneyalex - This PR is broken & I am likely to abandon in favor of the new sentinel strategy we devised during OCA days, so may not make for a good tester. 

I will, however, restart some ones that have been causing issues & will get back to you in OCA/runbot-addons#105 with results.

lasley:
Ok so circling back around to this - I have played around extensively with the `stock_scanner` & `stock_scanner_web` now, and I need something more opinionated for my requirements I think.

We are going to resume this & finalize in the next few weeks. Current requirements have us using RFID and hopefully scanning 50-100 units in less than a second, so there may be some updates required to the barcodes JS too.

web_stock/static/src/js/barcode.js
```diff
@@ -0,0 +1,192 @@
...
+                });
+                this.barcodeParser.load().then(this.onParserLoad);
+            }
+            this.$target.on('barcode_scanned', this.on_barcode_scanned);
+        },
+
+        onParserLoad: function() {
+            this.loaded = true;
+            console.log('Barcode parser initialized');
+        }
```
angelmoya, line 54: comma is missing

web_stock/static/src/js/barcode.js
```diff
@@ -0,0 +1,192 @@
...
+
+        on_barcode_scanned: function(barcode) {
+            console.log(barcode);
+            var self = this;
+            // Call hook method possibly implemented by subclass
+            this.preOnchangeHook(barcode).then(function(proceed) {
```
angelmoya, line 60: Hi @lasley I'm trying to test it, but I don't understand everything... I don't know why, I habe a picking, and product with barcode, I read this barcode and in this point proceed is false. Do you know that could be happing?
I also trying to migrate it to odoo 10.

angelmoya: Also If I force proceed=true after next if, then scan product works right

angelmoya:
@lasley are you still working on it?

web_stock/controllers/main.py
```diff
@@ -0,0 +1,226 @@
...
+            return request.env['stock.picking'].browse(picking_id)
+        else:
+            return picking_id
+
+    def __get_tpl_defaults(self, **kwargs):
+        return {
+            'errors': [],
+            'error_fields': kwargs.get('error_fields', []),
+            'db_info': json.dumps(db_info()),
+            'barcode_nomenclature': request.env.ref(
```
angelmoya, line 223: When I try to set lots, this says that barcode.nomenclature(1,) is not serializable. I'm trying to set id, and update input

lasley:
@angelmoya - I've been going back and forth on this honestly. IMO the real solution is to get our damned scenarios to work, I think the path I'm going here is doomed for being too limited. I'm really looking forward to seeing what your thoughts are with the HTML specific scenario, and will help out in any way I can.

angelmoya:
OK @lasley I will focus my effort on #55 now

rousseldenis:
Closing this as work will be done in further versions. See https://github.com/OCA/stock-logistics-barcode/pull/95#issuecomment-413135626

PR #40, opened 2016-10-03
[9.0] Add stock scanner module

Garamotte:
Migrated stock_scanner to v9.0.
The module was added in the stock-logistics-workflow repo on v8.0, but we finally think that it's better in this one, so I move it for the 9.0 migration.

The only difference between 8.0 and 9.0 is a menu's XML ID.
Changes from PR https://github.com/OCA/stock-logistics-workflow/pull/263 (on stock-logistics-workflow 8.0 branch) are included in this one.

PR #43, opened 2016-10-03
[9.0] Add stock_scanner_receipt module

Garamotte:
This PR is based on https://github.com/OCA/stock-logistics-barcode/pull/40

It adds a scenario module : Receipts


jgrandguillaume:
Hi,

Great work ! Good to split one module per scenario. Just a little remark:

I saw you have a common readme for all scenarios modules. It would be nice to have instead a explanation of the scenario it-self.

Regards,

Joël

jgrandguillaume:
:+1:  After discussion :)

JordiBForgeFlow (approved):
:+1: Functional tests

stock_scanner_receipt/__init__.py
```diff
@@ -0,0 +1,27 @@
+# -*- coding: utf-8 -*-
+##############################################################################
```
lasley: Please use new header

stock_scanner_receipt/__init__.py
```diff
@@ -0,0 +1,27 @@
...
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+
+
+# vim:expandtab:smartindent:tabstop=4:softtabstop=4:shiftwidth=4:
```
lasley: Remove vim stuff

stock_scanner_receipt/data/Receipt/scanner_scenario_step_receipt_done.py
```diff
@@ -0,0 +1,21 @@
+# flake8: noqa
```
lasley, line 1: Why no flake?

lasley (commented):
Minor comments inline

stock_scanner_receipt/data/Receipt/scanner_scenario_step_receipt_done.py
Garamotte: As explained on another PR, flake8 is disabled in these files, because some variables are only available at runtime (this code is evaluated during the scenario).

stock_scanner_receipt/__init__.py
Garamotte: **init**.py file emptied.

lasley (approved):
Thanks for the updates

PR #46, opened 2016-10-06
[MIG] Migrate stock_scanner to v10.0

Garamotte:
The module didn't need any modification to technically work on v10.

I just applied new naming :
- In XML files : openerp/data > odoo
- In python files : import openerp > import odoo
- Renamed the manifest from `__openerp__.py` to `__manifest__.py`

### DIFF
--- a/stock_scanner/__manifest__.py
+++ b/stock_scanner/__manifest__.py
@@ -5,7 +5,7 @@
 {
     'name': 'Stock Scanner',
     'summary': 'Allows managing barcode readers with simple scenarios',
-    'version': '9.0.1.0.0',
+    'version': '10.0.1.0.0',
     'category': 'Generic Modules/Inventory Control',
     'website': 'https://odoo-community.org/',
     'author': 'SYLEAM,'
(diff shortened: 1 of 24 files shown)


Garamotte:
Runbot green, but it tested on the Odoo's 7.0 branch, ans no other module than base was installed ?! (according to the test instance's log)

@gurneyalex : The runbot says "Server branch cannot be determined exactly. Please use naming convention '9.0-my-branch' to build with '9.0' server branch."
Is it possible that it searches for a single digit (9.0) instead of two (10.0) ?

pedrobaeza:
Runbot should be now working. Please restart build.

Garamotte:
@pedrobaeza : All 20 "testing" builds on runbot seem to be blocked.
I think you should reset their state to "pending" tu unblock the runbot.

pedrobaeza:
I'm afraid we have to wait. Maybe it's stuck, and we have to restart server... Let's see tomorrow. Now there are 115 pending builds.

alexis-via (approved):
Thanks for this port to v10 !

pedrobaeza (approved):

PR #47, opened 2016-10-10
[WIP][9.0] ADD stock_scanner_web

angelmoya:
WIP create module to use stock_scanner scenarios on web interface

This is a prototype module, it could be finish without too much effort.

To test only install module, create scenario o use demo data, and create an scanner hardware. It's better documentated on stock_scanner module.

To play scenario on web go to:
http://127.0.0.1:8069/scanner_call/hard1
hard1 is code for your hardware.

At this moment the module is based on simple template and controller, to finish I think it could be better to improve controler to get easier template.

It's needed to get last stage on scenarios, if you are playing scenario and go out, when you return it would be nice be on last stage.

Other thing to think is about split original scanner_module on twice modules, one stock_scanner_base and stock_scanner_ssh.

The web don't has css, it's needed to improve template.

Finally I think it's not need to add javascript, because scenarios stages only has one field to set, then focus could be set on this input, and enter key could do the submit.


stock_scanner_web/controllers/main.py
```diff
@@ -0,0 +1,41 @@
...
+                     terminal_number='',
+                     action='',
+                     message=False,
+                     type='http',
+                     auth='public',
+                     website=True):
+        scanner_hardware = request.env()['scanner.hardware'].sudo()
+        values = {}
+        try:
+            action = int(action)
```
Garamotte, line 26: I don't understand why you try to cast action to integer, as it always should be a string (see in this method : https://github.com/OCA/stock-logistics-barcode/blob/9.0/stock_scanner/models/scanner_hardware.py#L218)

stock_scanner_web/views/web_templates.xml
```diff
@@ -0,0 +1,160 @@
...
+                            </p>
+                            <p t-if="code in ['M','E']">
+
+                                <h3>
+                                    <span t-esc="result[0][1:]"/>
+                                </h3>
+                                <t t-foreach="result[1:]" t-as="line">
+                                    <p t-esc="line"/>
+                                </t>
+                                <a id="btn-go" t-attf-href="/scanner_call/{{terminal_number}}/action/0" class="btn">
```
Garamotte, line 81: You forgot to test the value : a default value set to True should call the 'back' action, instead of 'action'.

stock_scanner_web/views/web_templates.xml
```diff
@@ -0,0 +1,160 @@
...
+                                    <span t-esc="result[0][1:]"/>
+                                </h3>
+                                <t t-foreach="result[1:]" t-as="line">
+                                    <p t-esc="line"/>
+                                </t>
+
+                                <form method="get"
+                                    t-att-action="'/scanner_call/%s/action' % (terminal_number)" class="search_class form-inline">
+                                    <div class="input-group">
+                                        <input type="number" name="message" class="search-query form-control"/>
```
Garamotte, line 124: The float step doesn't allow float values, only integers (tested on Firefox 49).

Garamotte (commented):
Great work.
Some comments added inline.

You should also clean/remove files from OCA template.

JordiBForgeFlow:
We have continued @angelmoya  work here https://github.com/OCA/stock-logistics-barcode/pull/55. Can we close this one?

JordiBForgeFlow:
Since @angelmoya had just one commit we saw no good reason to maintain the commit history

pedrobaeza:
You need to keep his commit and thus his attribution.

PR #48, opened 2016-10-13
[FIX] Fixed scenario loading when overriding

Garamotte:
When overriding a scenario, we can add or modify transitions between steps that are not created in the current scenario. In this case, the id of the step was not in the local dictionnary.
This is fixed by calling env.ref(xml_id) to retrieve the step's id.

### DIFF
--- a/stock_scanner/load_scenario.py
+++ b/stock_scanner/load_scenario.py
@@ -189,7 +189,11 @@ def import_scenario(env, module, scenario_xml, mode, directory, filename):
         transition_values = {}
         for key, item in node.items():
             if key in ['to_id', 'from_id']:
-                item = resid[get_xml_id(_('step'), module, {'id': item})]
+                step_xml_id = get_xml_id(_('step'), module, {'id': item})
+                if step_xml_id in resid:
+                    item = resid[step_xml_id]
+                else:
+                    item = env.ref(step_xml_id).id
 
             transition_values[key] = item
 


stock_scanner/load_scenario.py
```diff
@@ -189,7 +189,11 @@ def import_scenario(env, module, scenario_xml, mode, directory, filename):
         transition_values = {}
         for key, item in node.items():
             if key in ['to_id', 'from_id']:
-                item = resid[get_xml_id(_('step'), module, {'id': item})]
+                step_xml_id = get_xml_id(_('step'), module, {'id': item})
+                if step_xml_id in resid:
```
pedrobaeza, line 193: You can do:

```
item = resid.get(step_xml_id, env.ref(step_xml_id).id)
```

Garamotte: This will call env.ref each time, even if the value is already in the `resid` dictionnary (which is the most common case).

pedrobaeza: No, it won't. It will only call in case of not being found in the dictionary. Check https://www.tutorialspoint.com/python/dictionary_get.htm

Garamotte: The default value will be returned if the key is not in the dictionnary.
But the `env.ref()` method will be called before, to generate this default value.

pedrobaeza: OK, I see. Thanks for the explanation.

pedrobaeza (approved):

lasley (approved):

lasley:
The coverage drop saddens me, but I won't block on a fraction of a percent

PR #49, opened 2016-11-08
[9.0] add module barcodes generator

legalsylvain:
This module add a new module based on barcode.rule model, to generate barcodes for any models ( mixin model.)
With that module, you can generate barcodes for partners and product. (.product / .template)

See the whole description in the readme file : 

https://github.com/OCA/stock-logistics-barcode/blob/9.0_ADD_module_barcodes_generator_2/barcodes_generator/README.rst

The PR has been initially done against OCA / pos repository.
https://github.com/OCA/pos/pull/130
@pedrobaeza renamed done + module moved. Thanks for your advice.

@pedrobaeza : I was thinking about to split my module in 3. 

* barcodes_generator_abstract (depends on base) : provide mixin model.
* barcodes_generator_product (depends on barcodes_generator_abstract + product)
* barcodes_generator_partner (depends on barcodes_generator_abstract + point_of_sale, due to Odoo bad design)

What to you think about that option ? 

Thanks for your review.


legalsylvain:
Hi @gurneyalex, runbot fails because pyBarcode python lib is not installed on Runbot instance. Could you install it ? (ref .travis.yml file)
If you're interested by this module, your review is welcome !
Kind regards.

gurneyalex:
@legalsylvain installed.

gurneyalex (approved):

legalsylvain:
@gurneyalex : what about spliiting module ? do you think it's a good idea. (to avoid pos dependency for product generation)

gurneyalex:
@legalsylvain hmm yes this would be nice actually.

barcodes_generator_abstract/README.rst
```diff
@@ -0,0 +1,149 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+===========================================
+Generate Barcodes for any Models (Abstract)
+===========================================
+
+This module expends Odoo functionnality, allowing user to generate barcode
```
lasley: s/functionnality/functionality

barcodes_generator_abstract/README.rst
```diff
@@ -0,0 +1,149 @@
...
+      internal code defined in your company)
+    * automaticaly by a sequence, if you want to let Odoo to increment a
+      sequence. (typical case of a customer number incrementation)
+
+* Generate a barcode, based on the defined pattern and the barcode base
+
+Installation
+============
+
+This module use an extra python librairy named 'pyBarcode' you should install
```
lasley: s/librairy/library

barcodes_generator_abstract/README.rst
```diff
@@ -0,0 +1,149 @@
...
+to make barcode generation works properly.
+
+sudo pip install pyBarcode
```
lasley: Maybe make this a code block?

barcodes_generator_abstract/models/barcode_generate_mixin.py
```diff
@@ -0,0 +1,86 @@
...
+from openerp import models, fields, api, exceptions, _
+
+from .barcode_rule import _GENERATE_TYPE
+
+_logger = logging.getLogger(__name__)
+
+try:
+    import barcode
+except ImportError:
+    _logger.debug("Cannot import 'barcode' python Librairy.")
```
lasley: s/Librairy/library

Also IMO this should be a warning, or at least info.

lasley (approved):
I love it, thanks @legalsylvain 

Few spelling issues, but the code looks good otherwise.

lasley:
Although tests would be nice if you're so inclined 😉

legalsylvain:
Thanks a lot for your review @lasley.
- I fixed typo / description. 
- Tests added.
- New modules "_partner" and "_product", to improve modularity.

Thanks for your review.

lasley:
Woot and test coverage is 🍏  too, nice! I'll drop another review on the changes sometime today or tomorrow just to make sure still good.

lasley:
@legalsylvain -Wow from a high level this just gained a lot more function. Is your proposal to supersede `product_barcode_generator` & #32?

barcodes_generator_abstract/__openerp__.py
```diff
@@ -0,0 +1,29 @@
...
+    'author':
+        'GRAP,'
+        'La Louve,'
+        'Odoo Community Association (OCA)',
+    'website': 'http://www.odoo-community.org',
+    'license': 'AGPL-3',
+    'depends': [
+        'barcodes',
+    ],
+    'data': [
```
alexis-via, line 21: Don't we need 
'external_dependencies': {'python': ['pyBarcode']},

alexis-via (approved):

legalsylvain:
@gurneyalex : Runbot is red but I don't understand why. 

https://runbot.odoo-community.org/runbot/build/3271805

The log is pretty crytic : 
https://runbot.odoo-community.org/runbot/static/build/3271805-9-0-ADD-module-barcodes-generato-5e78ca/logs/job_10_test_base.txt

legalsylvain:
if this module supersed the current module product_barcode_generator (uninstallable in V9.0 serie), I propose to remove the unused folder to avoid to let people trying to port this module.
Thanks for your feeback.

bealdav:
Nice job, thanks

Could you change your module names in singular form, please ?
Thanks

legalsylvain:
Hi. @bealdav. thanks for your review !

I choose the plurial form because original module is named barcodes. so, for consistency. We talk with @pedrobaeza, about the name of the module in an PR previously done against POS repository. See the whole discussion here : https://github.com/OCA/pos/pull/130#issuecomment-258639833

@pedrobaeza seems to be agree about that arg. But not a blocking point for me. I let community choose what is the best choice.

Thanks.

pedrobaeza:
Yeah, there's an exception in the guideline when Odoo itself uses plural: https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#modules

bealdav:
I don't find 'barcodes' module name in standard odoo ?
barcode

in runbot v8 at least.

pedrobaeza:
It's v9: https://github.com/odoo/odoo/tree/9.0/addons/barcodes

bealdav:
Ok you're right, exception are in guidelines and there this module in v9.
Another short error (exception) in implicit module naming rule (95 % applied).

Propagate this exception instead of restrict these effects is the choice in our guidelines.
But if tomorrow we decide to clean all names in version 54 of our erp favorite (whatever its name), it will be easier if we limit the exception now than expand it.
Consistency is not to propagate exceptions, it's to limit them, I think.

I could propose a PR to guidelines if are open to this logic.
My 2 cents

Thanks for your attention

pedrobaeza:
The rationale behind the exception propagation is to maintain a logic set of modules that are bound by complete name, and I think it's something admissible. You can see me in other repo claiming to change `intercompany_rules` to `intercompany_rule` because of there's not such binding, but it's not the case here, as the dependency chain is clear.

gurneyalex:
@legalsylvain regarding the "pretty cryptic"   log, someone with commit rights pushed new branches on the OCA repository  (https://github.com/OCA/stock-logistics-barcode/tree/9.0_ADD_module_barcodes_generator and https://github.com/OCA/stock-logistics-barcode/tree/9.0_ADD_module_barcodes_generator_2) and is looking at the runbot build logs for these branches  instead of the build for the current PR (https://runbot.odoo-community.org/runbot/build/3271805). 

#justsaying

barcodes_generator_abstract/__openerp__.py
legalsylvain: Well in fact, it's 'barcode', simply, but yes ! Thanks.
https://github.com/OCA/stock-logistics-barcode/pull/49/commits/5e78cabb5125aa2b6cf8788681d4bd1d6cc151c7#diff-3f9fcad59a91c0ecef85bbd7cde75c9c

legalsylvain:
@gurneyalex closed this PR because the branch of the PR was hosted by OCA (and not GRAP).
For that reason, Runbot failed.
I'll redo my PR, correctly.

Kind regards. 

#QuiEstLeConQui... #OupsCestMoi

PR #50, opened 2016-11-13
[7.0] [ADD] new module scan_to_inventory

legalsylvain:
Add new mobile module to make inventory with a mobile phone.

* This module provides a new web apps, available on http(s)://my_oodoo_url/scan_to_inventory/static/www/index.html that allow user to use mobile (with barcode scanner) to make inventory.

* FYI, this module works fine with the following devices
    * robust mobile phone [Samsung Xcover 3](https://www.amazon.fr/Samsung-Galaxy-Xcover-VE-Smartphone/dp/B01FH8MDMW/ref=cm_cr_arp_d_product_top?ie=UTF8)
    * Bluetooth Barcode scanner [KDC400](https://koamtac.com/kdc400-bluetooth-barcode-scanner/)

* The complete description of the module is available [here](https://github.com/grap/stock-logistics-barcode/blob/7.0_ADD_scan_to_inventory/scan_to_inventory/README.rst). (screenshots, technologies, etc...)

* You can test the feature with runbot. If you do so, please check : 
    * that the database is xxx-all. (module is not installed on xxx-base).
    *  At least a product has a barcode defined. I set one with 5400313040109 for test.
    * tip "Ctrl + Shift + M" on your firefox browser, to emulate mobile device.

**About Module name**
* I'm not sure about the module name. @pedrobaeza, @dbeal, @sebastienbeau, could you take a look ? I propose : 
* scan_to_inventory (current name)
* mobile_app_inventory
* ionic_mobile_app_inventory
* ionic_app_inventory

But i'd like to have your PoV, before making this change.

**Possible refactoring in V8+**
For the time being, all the extra libs. (ionic, ...) are embeded in this module. when I'll port this module in new version, I'd like to split this module in two.
* mobile_app_abstract, with all the generic libs and common features
* mobile_app_inventory, that provides only the inventory feature

This refactoring should prevent duplicated code for other existing module, like scan_to_purchase module, that allow user to make a purchase order, scanning product. The module works correctly in production 6 months, and is available [here](https://github.com/grap/odoo-addons-mobile/tree/7.0/scan_to_purchase).


legalsylvain:
Hi @gurneyalex : I have a little problem with this PR for runbot.
**Fact** 
- my module add a simple extra product with ean13 defined. (in scan_to_inventory/demo/product_product.xml)
- Once an existing module is installed 'product_multi_ean', it doesn't seems possible to create demo product with ean13 defined. See log at the bottom of this comment.

Travis was failing, I so wrote an exclude rule in Travis file, that fixed the problem.
but runbot is still failing. How could I fix this problem ?

Possible solution : 
- [DONE] remove barcode on demo data of 'scan_to_inventory' module.
- try to fix product_multi_ean, but I really don't know this module, and this PR is an intent to share this work with OCA, but I'll migrate to Odoo 8 in a few time, so I would prevent to take a long time for that work. 
- other ? Is it possible to set similar feature as Travis with runbot ? (exclude / multiple database)
- alternatively, reordering installation order should fix the bug. (if scan_to_inventory is installed before 'product_multi_ean', it should work.  

Thanks for your feedback.

```
openerp.modules.loading: module scan_to_inventory: loading demo/product_product.xml
openerp.sql_db: bad query: insert into "product_product" (id,product_tmpl_id,"track_production","ean13","color","track_incoming","price_extra","default_code","track_outgoing","active","valuation","price_margin",create_uid,create_date,write_uid,write_date) values (50,50,'False','5400313040109',0,'False','0.00','ORG-CHIPS-PAP','False','True','manual_periodic','1.00',1,(now() at time zone 'UTC'),1,(now() at time zone 'UTC'))
Traceback (most recent call last):
  File "/srv/openerp/instances/openerp-oca-runbot/parts/odoo-extra/runbot/static/build/3272214-50-cafaf5/openerp/sql_db.py", line 226, in execute
    res = self._obj.execute(query, params)
DataError: value "5400313040109" is out of range for type integer
LINE 1: ..._date,write_uid,write_date) values (50,50,'False','540031304...
```

legalsylvain:
@rafaelbn, @pedrobaeza : PR is now ready for review and for test. Thanks.

gurneyalex:
@legalsylvain I'm sorry, I don't quite understand what the issue is. Could you clarify?

scan_to_inventory/static/www/controllers/c_confirm_quantity.js
```diff
@@ -0,0 +1,65 @@
...
+        'new_qty': 0,
+        'sum_qty': 0,
+    };
+
+    $scope.$on(
+            '$stateChangeSuccess',
+            function(event, toState, toParams, fromState, fromParams){
+        if ($state.current.name === 'confirm_quantity') {
+            // Get Product Data
+            $scope.data.product_id = parseInt(toParams.product_id);
```
hparfr: add the 2nd arg (base)
`parseInt(toParams.product_id, 10)`

scan_to_inventory/static/www/controllers/c_confirm_quantity.js
```diff
@@ -0,0 +1,65 @@
...
+
+    // TODO
+    // Refactor me. merge add_quantity and replace_quantity code
+    $scope.add_quantity = function () {
+        StockInventoryModel.AddInventoryLine(
+                $rootScope.currentInventoryId,
+                $rootScope.currentLocationId, $scope.data.product_id,
+                $scope.data.sum_qty, 'add').then(function (res){
+            if (res.state == 'write_ok'){
+                angular.element(document.querySelector('#sound_quantity_selected'))[0].play();
```
hparfr: avoid as possible to query the DOM from controller.
Use a directive instead.

scan_to_inventory/static/www/controllers/c_confirm_quantity.js
```diff
@@ -0,0 +1,65 @@
...
+                setTimeout(function(){
```
hparfr: use $timeout instead otherwise $state.go will run outside of the digest cycle

scan_to_inventory/static/www/controllers/c_load.js
```diff
@@ -0,0 +1,55 @@
...
+                StockLocationModel.LoadLocationList().then(function(location_qty){
+                    $scope.data.location_ok = true;
+                    $scope.data.location_qty = location_qty;
+                    StockInventoryModel.LoadDraftInventoryList().then(function(inventory_qty){
+                        $scope.data.inventory_ok = true;
+                        $scope.data.inventory_qty = inventory_qty;
+                        $scope.doneMessage = $translate.instant("Loading Done");
+                        setTimeout(function(){
+                            $state.go('select_stock_inventory');
+                        }, 1000);
```
hparfr: you can avoid this callback hell by chaining promises

legalsylvain:
Hi @gurneyalex. thanks for your answer.

I broke runbot that was green with the [last very simple commit](https://github.com/OCA/stock-logistics-barcode/pull/50/commits/89647970b0a321cae0744963e5e405c7721467ca) for demo.

In a few word, the new module **scan_to_inventory** I propose in this PR, is incompatible with an existing module **product_multi_ean** ( \* ). So it breaks CI and prevents runbot to work : 
- For Travis, I set exclude / include mechanism in .travis.yml file. ([Ref](https://github.com/grap/stock-logistics-barcode/blob/89647970b0a321cae0744963e5e405c7721467ca/.travis.yml#L24)). It's OK.
- For Runbot, my question is how to fix that ? Is it possible to have defined in a config file to finally have three databases : base,all-exclude-product_multi_ean and all-exclude-scan_to_inventory. (or other solutions). I guess that is not the first this case occurs.

( \* ) : **product_multi_ean** changes product ean13 field from char to many2one (technically integer in database). So the demo data is too big. The same bug will occur if you install with demo data **product_multi_ean** and then **point_of_sale** module that have demo products with ean13 defined.

```
DataError: value "5400313040109" is out of range for type integer
```

Thanks.

scan_to_inventory/static/www/controllers/c_select_quantity.js
```diff
@@ -0,0 +1,64 @@
...
+            // Set Focus
+            angular.element(document.querySelector('#input_quantity'))[0].focus();
+            $scope.data.qty = '';
+            // Get Product Data
+            $scope.product = $rootScope.ProductListByEan13[toParams.ean13];
+        }
+    });
+
+    $scope.submit = function () {
+        if ($scope.data.qty !== null && !isNaN($scope.data.qty)){
```
hparfr: `if (!$scope.data.qty)`

scan_to_inventory/static/www/controllers/c_select_quantity.js
```diff
@@ -0,0 +1,64 @@
...
+                    $scope.errorMessage = $translate.instant("Too Small Quantity");
+                    angular.element(document.querySelector('#sound_user_error'))[0].play();
+                }
+            }else{
+                $scope.errorMessage = $translate.instant("Too Big Quantity");
+                angular.element(document.querySelector('#sound_user_error'))[0].play();
+            }
+        }
+        else{
+            $scope.errorMessage = $translate.instant("Incorrect Quantity");
```
hparfr: you can parseInt instead

scan_to_inventory/static/www/controllers/c_select_stock_location.js
```diff
@@ -0,0 +1,32 @@
...
+            '$stateChangeSuccess',
+            function(event, toState, toParams, fromState, fromParams){
+        if ($state.current.name === 'select_stock_location') {
+            $scope.data.location_list = $rootScope.LocationList;
+            $scope.data.location_qty = $rootScope.LocationList.length;
+            // Skip this screen if there is only one internal location
+            if ($rootScope.LocationList.length === 1){
+            $rootScope.currentLocationId = $rootScope.LocationList[0].id;
+            $rootScope.currentLocationName = $rootScope.LocationList[0].name;
+                $state.go('select_product_product');
```
hparfr: wrong indentation

scan_to_inventory/static/www/models/m_product_product.js
```diff
@@ -0,0 +1,18 @@
...
+        '$q', '$rootScope', 'jsonRpc',
+        function ($q, $rootScope, jsonRpc) {
+
+    return {
+        LoadProductList: function() {
+            return jsonRpc.call(
+                    'product.product', 'scan_to_inventory_load_product', []).then(function (res) {
+                $rootScope.ProductListByEan13 = res;
+                var quantity = 0;
+                for (var a in res){quantity ++;}
```
hparfr: `quantity = res.length` ?

scan_to_inventory/static/www/models/m_stock_inventory.js
```diff
@@ -0,0 +1,39 @@
...
+                    'id', 'name', 'date', 'inventory_line_qty',
+                    ]).then(function (res) {
+                $rootScope.DraftInventoryList = res.records;
+                return res.records.length;
+            });
+        },
+        CreateInventory: function(name) {
+            return jsonRpc.call(
+                    'stock.inventory', 'create_by_scan',
+                    [name]).then(function (res) {
```
hparfr: useless .then callback

scan_to_inventory/static/www/controllers/c_load.js
```diff
@@ -0,0 +1,55 @@
...
+        function ($scope, $rootScope, jsonRpc, $state, ProductProductModel, StockLocationModel, StockInventoryModel, $translate) {
+
+    $scope.data = {};
+
+    $scope.$on(
+            '$stateChangeSuccess',
+            function(event, toState, toParams, fromState, fromParams){
+        if ($state.current.name === 'load') {
+            $scope.data.product_ok = false;
+            $scope.data.product_ko = false;
```
hparfr: 1) never used
2) use ngHide instead

scan_to_inventory/static/www/controllers/c_confirm_quantity.js
```diff
@@ -0,0 +1,65 @@
...
+            $scope.data.new_qty = parseInt(toParams.new_qty);
+            $scope.data.sum_qty = $scope.data.current_qty + $scope.data.new_qty;
+        }
+    });
+
+    // TODO
+    // Refactor me. merge add_quantity and replace_quantity code
+    $scope.add_quantity = function () {
+        StockInventoryModel.AddInventoryLine(
+                $rootScope.currentInventoryId,
```
hparfr: it's better to use a service for storing the states.

hparfr (approved):
some style recommandations

gurneyalex:
@legalsylvain I don't see a way out of this. The offending module redefines the column from char to m2o which is really a bad idea as it completely breaks the db layout. OTOH changing this would mean breaking existing installations. 

I just created #52 to discuss this.

legalsylvain:
@gurneyalex : Thanks for the time taken to understand this problem. Let's talk about that point on the issue you have created.
Well, in fact, not a big deal, I'll remove again demo data to make Runbot green. the only problem is that once the PR merged, it will be not possible to test the new module, online, on V7.0 runbot.
kind regards.

scan_to_inventory/static/www/controllers/c_load.js
legalsylvain: 1. product_ok and product_ko are both used in v_load.html. (and use ng-hide / ng-show)
   There are 3 state in fact : 
2. product_ok false / product_ko false : loading 
3. product_ok True / product_ko false : OK
4. product_ok False / product_ko True : KO

Is it OK, if I just set a state $scope.data.product_state : 'loading' / 'ko' / 'ok' , and set ng-show="product_state == 'ok'", in load view ?

scan_to_inventory/static/www/controllers/c_load.js
legalsylvain: In fact, it was just for test. I'll remove it useless setTimeout.

scan_to_inventory/static/www/models/m_product_product.js
legalsylvain: Because it doesn't work. 
**log**
console_screen

**result**
log_screen

Any idea ?
(indeed, It's not very clean)

scan_to_inventory/static/www/controllers/c_select_quantity.js
legalsylvain: typo in your suggestion I guess. the line below works.

```
if ($scope.data.qty)
```

And check about numeric field is managed by the view. It should be enougth. 

```
<input type="number" ng-model="data.qty" />
```

Thanks for pointing this simplification.

scan_to_inventory/static/www/controllers/c_load.js
hparfr: While I was reading your code I didn't get why you had product_ok and product_ko at the same time. You can may be choose better names  or may be use only one variable: 
`product_loaded_correctly = true | false | undefined`
(or you can even add the value: `product_qty = number | false | undefined`)

scan_to_inventory/static/www/models/m_product_product.js
hparfr: From your log trace : res is an object with an ean13 as key and an object as value.
Size of an object does not make a lot of sense in your case. It's may be better to transfert `res` as an array.

What you want to do is : 
`Object.keys(res).length`

legalsylvain: Thanks !

rafaelbn:
Hi @JGarcia-Panach you could be interested in this PR.

legalsylvain:
To test on runbot, please review and approve before https://github.com/OCA/stock-logistics-barcode/pull/57.

scan_to_inventory/static/www/controllers/c_load.js
legalsylvain: Well, I didn't remember, but I wrote this line to let user to see that the loading was ok. (new message, displayed with ng-show inventory_ok)
What do you propose alternatively ?

hparfr:
```js
$q.all([
PPM.LoadProductList().then( qty => { $scope.data.product_ok = true; $scope.data.producty_qty = qty;}),
SLM.LoadLocationList().then(qty => { $scope.data.location_ok = true; $scope.data.location_qty = qty;}),
SIM.LoadDraftInventoryList().then(qty => { $scope.data.inventory_ok = true; $scope.data.inventory= qty;})
]).then( () => $scope.doneMessage = $translate.instant('Loading Done'), () => $scope.LoadingError())
```

legalsylvain: Sorry, maybe new question, but what is $q ?

hparfr: https://docs.angularjs.org/api/ng/service/$q

scan_to_inventory/static/www/controllers/c_select_quantity.js
legalsylvain: Done, above in the source code.

scan_to_inventory/static/www/models/m_stock_inventory.js
legalsylvain: I did not understood this part. but I guess that the first return is useless. (return jsonRpc.call).
The interest of this function is to store inventory quantity in a global var too, so we have to wait, don't we ?

legalsylvain:
Hi @hparfr,
Thanks a lot for all your comment. I took into account all of them, except one, because I didn't understood.

scan_to_inventory/static/www/models/m_stock_inventory.js
hparfr: keep `return jsonRpc.call`
remove `.then(function (res) { return res;});`

hparfr:
👍 
Thanks for this usefull app !

scan_to_inventory/static/www/models/m_stock_inventory.js
legalsylvain: Thanks ! Done.

legalsylvain:
Thanks @hparfr for your complete review. do you have any idea about my initial question about module name ?
(not a blocking point, these module do not provide model changes, so it's possible to rename it very easely between 2 major versions)
regards.

hparfr:
I like `mobile_app_inventory`. 
`Ionic` is not necessary in the name and `scan_to_inventory` doesn't says it's a mobile app.

legalsylvain:
@hparfr : renamed done. Thanks for your suggestion.

@rafaelbn, @JGarcia-Panach, could you take a moment to test ?
(just have to go to runbot and to write runbot_url/mobile_app_inventory/static/www/index.html)

Thanks !

JGarcia-Panach:
Dear @legalsylvain,
sorry I don't have a cell phone with barcode scanner or anything similar.

However I have tried your module in my Iphone SE (writing the barcode by hand) and it works perfectly.
* I can create a new inventory
* I can select the location
* I can write the barcode and it allows me to select the quantity
* If the line exists, it allows me adding to the current quantity or replacing it
* If I create a new product (in the backend) I click refresh (in the cell phone) and I can add the new product with its barcode to the inventory

Every operation worked as expected. (Everything tested with the cell phone)

So, you have my 👍 for these parts

Thanks for all your hard work!

legalsylvain:
Hi @JGarcia-Panach and @hparfr . Thanks a lot for your review.

Note @JGarcia-Panach : it's not necessary to test with a real phone. You can go to runbot with your firefox browser, and then tip "Ctrl + Maj + M". It will emulate a phone interface. (you can choose the size of the screen.)

kind regards.

PR #55, opened 2016-11-24
[9.0] stock_scanner_web

JordiBForgeFlow:
Supersedes https://github.com/OCA/stock-logistics-barcode/pull/47


JordiBForgeFlow:
Cc @eficent

JordiBForgeFlow:
Cc @eficent

pedrobaeza:
Respect commit attribution

LoisRForgeFlow:
@pedrobaeza is it fine now?

Garamotte:
@lreficent You should rebase on 9.0.
This will remove the merge commit, which is useless here, and make the history cleaner.

stock_scanner_web/views/web_templates.xml
```diff
@@ -0,0 +1,247 @@
...
+
+<openerp>
+    <data>
+
+        <template id="stock_scanner_web.assets_frontend" name="Website assets">
+            <meta name="viewport"
+                  content="width=device-width, initial-scale=1"/>
+            <!-- Include the jQuery library -->
+            <script type="text/javascript"
+                    src="https://code.jquery.com/jquery-1.11.3.min.js"/>
```
lmignon: These externals libraries should be embedded into the module since in some cases the servers do not have access to internet

stock_scanner_web/README.rst
```diff
@@ -0,0 +1,81 @@
...
+Configuration
+=============
+
+To configure this module, you need to:
+
+* Create a Scanner Hardware and add to the user. You can now do this
+directly from the user form.
+* Also, do not forget to enable "Sentinel: technical users" option in the user
+form, as it is required by the basic stock_scanner module.
+* Go to "Inventory > Configuration > Scanner Configuration > Scanner
```
lmignon: Does that means that it's not possible to use at the same time the web interface and the sentinel interface?

stock_scanner_web/controllers/main.py
```diff
@@ -0,0 +1,81 @@
...
+                else:
+                    values = {
+                        'code': 'N',
+                        'result': "You do not have any hardware allowed. " \
+                                  "Please contact your administrator."
+                    }
+                return http.request.render('stock_scanner_web.hardware_select',
+                                               values)
+
+            scanner_hardware = request.env()['scanner.hardware'].sudo()
```
lmignon: `sudo`????  Security? Record Rules?

stock_scanner_web/README.rst
JordiBForgeFlow: @lmignon Our intention is that it should be always compatible. Why do you think that the readme suggests it's not?

lmignon: @jbeficient bcs you ask to disable the login/logout scenarios

JordiBForgeFlow: Agreed. We'll adjust that.

stock_scanner_web/views/web_templates.xml
JordiBForgeFlow: Agreed! We'll fix that.

stock_scanner_web/controllers/main.py
JordiBForgeFlow: Agreed. We should respect the security rules and throw an error message is an operation is not permitted.

gurneyalex:
@jbeficent can you rebase your PR? This will get the updated travis configuration and allow a runbot build.

stock_scanner_web/README.rst
LoisRForgeFlow: Yes, because the login is checked when the user enters the odoo database, so we thought is not needed to ask for credentials again in the scanner web as the access it's done through the inventory menu. The web will show to the user his/her hardware allowed to use directly and check for permissions if he or she try to access another hardware via typing other device's address.
In order to make possible to use web and sentinel at the same time we think the login/logout option should be moved to the hardware configuration. This way you can specify no login/logout for hardware using web and enable it for the sentinel.

cc @sylvain-garancher

Garamotte: Maybe it would be possible to override the `_scenario_list` method, to filter out the login/logout scenarios when using the web interface ?

This would allow to use both web interface and sentinel, from the same user, using the same hardware configuration.

lmignon: IMO the proposal of @sylvain-garancher is the way to go

LoisRForgeFlow:
How can we skip the js and css src files in the travis validation? (Because they are external)

Garamotte:
You have to move external libraries into `static/lib`.

stock_scanner_web/controllers/main.py
```diff
@@ -0,0 +1,86 @@
...
+    ], website=True, auth='user')
+    def scanner_call(self,
+                     terminal_number='',
+                     action='',
+                     message=False,
+                     type='http',
+                     auth='public',
+                     website=True):
+        values = {}
+        scanner_hardware = request.env()['scanner.hardware']
```
lasley: Why calling env as a method? Should just be [`request.env['scanner.hardware']`](https://github.com/odoo/odoo/search?utf8=%E2%9C%93&q=%22request.env%22). Please update all other instances of this as well.

stock_scanner_web/controllers/main.py
```diff
@@ -0,0 +1,86 @@
...
+        try:
+            user = request.env()['res.users'].browse(request.uid)
```
lasley: `request.env.user`

stock_scanner_web/controllers/main.py
```diff
@@ -0,0 +1,86 @@
...
+            if terminal_number:
+                if not allowed_hardware(user, terminal_number):
+                    values = {
+                        'code': 'E',
+                        'result':
+                            'Hardware {} not allowed for user {}.'.format(
```
lasley: Should these results be translated?

stock_scanner_web/static/description/index.html
```diff
@@ -0,0 +1,124 @@
+<section class="oe_container">
```
lasley: Please put this in [README.rst](https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst)

pedrobaeza:
Following Odoo's naming conventions, shouldn't this be stock_scanner_website?

lmignon:
IMO, `stock_scanner_web` is the right name. Indeed, this module provides a dedicated web interface for a backend functionality. This interface is not intended to be used by a public user. It's a specialized web interface designed to be used with some specialized hardware. Moreover this module doesn't depend of the `website` module.

lmignon:
last build on runbot...
image

LoisRForgeFlow:
@sylvain-garancher @lmignon the repo is set up with .gitignore to exclude "lib" folders from git. https://github.com/OCA/stock-logistics-barcode/blob/10.0/.gitignore#L16. That's why the static libraries do not show up. Shall we remove this restriction from .gitignore? @gurneyalex

gurneyalex:
@lreficent yes please.

stock_scanner_web/controllers/main.py
```diff
@@ -0,0 +1,114 @@
...
+                     terminal_number='',
+                     action='',
+                     message=False,
+                     type='http',
+                     auth='public',
+                     website=True):
+        values = {}
+        if message == 'False':
+            message = False
+        try:
```
lasley: This try/except is quite long & I am unable to determine what line(s) it is actually guarding. It would be better to immediately perform the action that could generate the exception & return. It would also reduce the nesting

stock_scanner_web/controllers/main.py
```diff
@@ -0,0 +1,114 @@
...
+            # Determine the correct hardware:
+            scanner_hardware = False
+            user = request.env.user.browse(request.uid)
```
lasley: Why not just `request.env.user`?

stock_scanner_web/controllers/main.py
```diff
@@ -0,0 +1,114 @@
...
+                        'result':
+                            _('Hardware {} not allowed for user {}.').format(
+                                terminal_number, user.name)
+                    }
+                    return http.request.render(
+                        'stock_scanner_web.hardware_select',
+                        values)
+            else:
+                terminal_list = []
+                for hardware in user.scanner_hardware_ids:
```
lasley: `terminal_list = user.scanner_hardware_ids.mapped('code')`

stock_scanner_web/controllers/main.py
```diff
@@ -0,0 +1,114 @@
...
+            if not scanner_hardware:
+                values = {
+                    'code': 'N',
+                    'result': _("No valid terminal.")
+                }
+                return http.request.render(
+                    'stock_scanner_web.hardware_select',
+                    values)
+            if not message and action == 'reset':
+                scanner_hardware.sudo().empty_scanner_values()
```
lasley: Why is the sudo required?

stock_scanner_web/controllers/main.py
```diff
@@ -0,0 +1,114 @@
...
+                                       values)
+        except Exception as e:
+            values = {
+                'code': 'E',
+                'result':
+                    _('Error: %s') % e.message
+            }
+            return http.request.render(
+                'stock_scanner_web.hardware_select',
+                values)
```
lasley: Add newline

stock_scanner_web/models/res_user.py
```diff
@@ -0,0 +1,14 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from openerp import models, fields
+
+
+class ResUser(models.Model):
+    _inherit = 'res.users'
+
+    scanner_hardware_ids = fields.Many2many(comodel_name='scanner.hardware',
+                                            string='Scanners allowed to use')
```
lasley: `Permitted Scanners` sounds a little better IMO

stock_scanner_web/views/res_user_view.xml
```diff
@@ -0,0 +1,27 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<!--
+© 2016 Lois Rilo <lois.rilo@eficent.com>
+Eficent Business and IT Consulting Services, S.L.
+License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+-->
+
+<openerp>
```
lasley: `<odoo>` and can remove data tag

stock_scanner_web/views/res_user_view.xml
```diff
@@ -0,0 +1,27 @@
...
+    <data>
+        <!-- Add scanners allowed for the user. -->
+        <record model="ir.ui.view" id="user_scanners_form_view">
```
lasley: `id` should be the same as the one you are inheriting `view_users_form`

stock_scanner_web/views/warehouse_menu.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<!--
+    Copyright 2016 Eficent Business and IT Consulting Services, S.L.
+    <http://www.eficent.com>
+    License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+-->
+
+<openerp>
```
lasley: `<odoo>`

stock_scanner_web/views/web_templates.xml
```diff
@@ -0,0 +1,243 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--
+    Copyright 2016 Angel Moya <http://angelmoya.es>
+    Copyright 2016 Eficent Business and IT Consulting Services, S.L.
+    <http://www.eficent.com>
+    License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+-->
+
+<openerp>
```
lasley: odoo tag & remove data

stock_scanner_web/views/web_templates.xml
```diff
@@ -0,0 +1,243 @@
...
+    <data>
+
+        <template id="stock_scanner_web.assets_frontend" name="Website assets">
+            <meta name="viewport"
+                  content="width=device-width, initial-scale=1"/>
+            <script type="text/javascript"
+                    src="/stock_scanner_web/static/lib/js/jquery-1.11.3.min.js"/>
```
lasley: Why not just use the Odoo builtin jQuery for non-mobile?

stock_scanner_web/views/web_templates.xml
```diff
@@ -0,0 +1,243 @@
...
+                    </div>
+                </div>
+            </t>
+        </template>
+
+        <template id="scanner_call" name="header" page="True">
+            <t t-call="stock_scanner_web.layout">
+                <div id="wrap" class="oe_structure oe_empty" data-role="page">
+                    <div data-role="main" class="ui-content">
+                        <!--<a t-attf-href="/scanner_call/"-->
```
lasley: Please remove commented code

stock_scanner_web/controllers/main.py
```diff
@@ -0,0 +1,114 @@
...
+    allowed = False
+    for hardware in user.scanner_hardware_ids:
+        if t_num == hardware.code:
+            allowed = True
+    return allowed
+
+
+class ScannerWeb(http.Controller):
+    @http.route([
+        '/scanner_call',
```
lasley: I dunno if we should add into a namespace that isn't owned by this module. It is convention to use `/module_name/*` in order to maintain namespace isolation of modules

stock_scanner_web/views/web_templates.xml
```diff
@@ -0,0 +1,243 @@
...
+<openerp>
+    <data>
+
+        <template id="stock_scanner_web.assets_frontend" name="Website assets">
+            <meta name="viewport"
+                  content="width=device-width, initial-scale=1"/>
+            <script type="text/javascript"
+                    src="/stock_scanner_web/static/lib/js/jquery-1.11.3.min.js"/>
+            <link rel="stylesheet"
+                  href="/stock_scanner_web/static/lib/css/jquery.mobile-1.4.5.min.css"/>
```
lasley: Odoo >=9 minifies for us. We should include libs as non-minified versions

stock_scanner_web/views/web_templates.xml
```diff
@@ -0,0 +1,243 @@
...
+                    src="/stock_scanner_web/static/lib/js/jquery.mobile-1.4.5.min.js"/>
+        </template>
+
+        <template id="layout" name="Main layout">&lt;!DOCTYPE html&gt;
+            <html>
+                <head>
+                    <title>Scanner Web</title>
+                    <t t-call-assets="stock_scanner_web.assets_frontend" t-js="false"/>
+                </head>
+                <body style="background-color:powderblue;">
```
lasley: Aren't these colors what the `_CURSES_COLORS` is for? Maybe expose that to the template via controller? Or maybe I'm misunderstanding

stock_scanner_web/controllers/main.py
LoisRForgeFlow: Ok, we'll change to stock_scanner_web

stock_scanner_web/controllers/main.py
LoisRForgeFlow: We will improve that.

stock_scanner_web/controllers/main.py
LoisRForgeFlow: stock_scanner uses also calls to this method with sudo() in similar instances. By default only scanner hardware manager can modify hardware data, but empty_scanner_values is an exception. cc @sylvain-garancher

stock_scanner_web/controllers/main.py
LoisRForgeFlow: Yes, that's enough. I'll change it

stock_scanner_web/controllers/main.py
LoisRForgeFlow: Agree

stock_scanner_web/models/res_user.py
LoisRForgeFlow: Ok, I will change that.

stock_scanner_web/views/res_user_view.xml
LoisRForgeFlow: I'll fix

stock_scanner_web/views/web_templates.xml
LoisRForgeFlow: Yes, that's a good idea I will change it.

stock_scanner_web/views/web_templates.xml
LoisRForgeFlow: I will try that for version 9.

stock_scanner_web/views/web_templates.xml
LoisRForgeFlow: You're totally right. it is a remaining of the very first design.

stock_scanner_web/views/web_templates.xml
LoisRForgeFlow: I'll fix

lharvengt:
Dear All, 

We have been playing around with the v8 version of this PR as we would like to use the stock_scanner_web functionalities for one of our customers. I am myself very new with Odoo, found a couple of issues and would like to submit them to the community for advice before going forward:

1. Using the browser back button can lead to errors or inconsistencies.
For example, if you press back on the location prompt of the stock scanner inventory scenario, you resubmit the product where we would expect a location.
A blunt solution would perhaps be to add an UUID to the URL (say after the terminal number). It would be generated inside the ScannerWeb.scanner_call routine and saved in a list related to the scanner hardware. We could then check early on in the same routine if the UUID being submitted is already known, in which case we would translate this to a back action. It would thus be able to manage several “back” presses in a row. A benefit would perhaps be that it is completely browser-independent. Does anyone have a better idea? I am told by colleagues that adding some javascript somewhere could perhaps do the trick but I am fairly ignorant in the area.
2. In certain cases, where the next page’s URL is identical to the current one, the browser doesn’t seem to resubmit the page so you end up being stuck.
For example, in the step types tutorial, you are stuck on the “Message step” step. 
The proposed solution to 1. would also deal with this issue as the URLs would always be different. Otherwise I guess we would have to insert a variation in it.
3.  The float input type currently uses the same validation as integers.
For example, try the float step of the tutorial. 
A quick fix would be to differentiate the validation for floats and add a step=”0.000000000001” in the input tag of the relevant template. Again, colleagues are suggesting that a better solution would be to somehow accept as a parameter the precision.

Thank you in advance for your feedback,

Lionel.

LoisRForgeFlow:
Yes, I agree with your issues @lharvengt but I'm also quite new and not sure about which is the correct approach to solve them. I would appreciate your work and proposals in the topic.

LoisRForgeFlow:
With last commit we solve the point 2 I think. 

Referring to the first point, I would say that we should not allow the user to use the back button when into a scenario, due to the way the models of stock_scanner module are defined. So, IMO a easy way that might cover this is to implement a little javascript that pops up a warning message to the user when trying to use the back button of the web browser.

Also, we are considering the possibility of develop a new web that manage everything into the same url through javascript, but this is a kind of a more long term idea.

Again, your feedback and proposals are welcome.

cc @lharvengt

JordiBForgeFlow:
@lmignon What would you think about managing it more on the way Odoo implements web apps, like the POS or the barcode v8? In that way, the user stays in the same URL always, and the interaction is handled by means of javascript. Then the module would be using a series of widgets would be in charge of the whole interaction between user and server. We're not much experienced in this part, but we could explore this option.

We have also been told that it's dangerous to navigate to new URL's using handheld scanners. If the user scans very fast repeatedly, the browser may not be able to handle the input event properly. We have not been able to reproduce this issue so far, though.

lasley:
@jbeficent - I agree that JS is the way to go here. You are also correct in your eval of successive scanning in relation to the URI.

Odoo SA provides an awesome barcodes scanner JS already, I recommend we use that. Take a look at my PoC in #38, which uses it. The JS is just a mirror of the barcodes.py, so it's fairly easy to learn.

lmignon:
@jbeficient I agree that JS is the way to go. IMO We must continue the work started here since the approach is more robust and extensible than the barcodes scanner from Odoo SA.

lmignon:
@jbeficent  I agree that JS is the way to go. IMO We must continue the work started here since the approach is more robust and extensible than the barcodes scanner from Odoo SA.

lasley:
> We have also been told that it's dangerous to navigate to new URL's using handheld scanners. If the user scans very fast repeatedly, the browser may not be able to handle the input event properly. We have not been able to reproduce this issue so far, though.

I was recently able to replicate this using RFID scanners. Definitely a real issue.

angelmoya:
I think that fast scan could be a problem also with JS bassed app... you will allways get this problem if you want to do a process after read any code. So I think we can solve this creating new step in scenario to scan several lots... this new step scenario could be simple text multi line input, and process this input in next step of this scenario. This could be not  pretty... but could be right way.
Years ago I created similar app in android, to solve fast scan I did multi process app, but If barcode that you read is not valid, then you lost the control off this process... so I think is better to first read all, and then process your input.

Also I tested this module on Odoo 10 and works nice

angelmoya:
@lreficent  What do you think about use autocomplete="off" on inputs?

LoisRForgeFlow:
@angelmoya I think is a good idea, because the autocomplete is in fact not useful at all.

Are you working in some improvements on this topic?

angelmoya:
Now I'm only testing, I check it on Odoo 10, and is working right, I also test inventory scenario, my next steep will be create new scenario for special case of receptions... and then I will start to work on it.

I also thought about other visual improvements, like:
- Enable/disable visual keypad by scanner hardware, and show directly without button.
- Aling buttons to bottom
- On text field prevent enter capture to be able to write multi line text.

LoisRForgeFlow:
@angelmoya I like your first point, or maybe even to have three options: permanently enabled, disabled and button. But I will have to check how to do it. Whereas for the other points I will need to refresh my mind but you are probably right.

I will give some time to this in the following weeks.

angelmoya:
Thanks @lreficent we will talking about it

lasley:
@angelmoya - If we use the Odoo Barcodes JS, these are fired as events. We can then trap the events into a history queue. 

Take a look at [my implementation in web_stock](https://github.com/LasLabs/stock-logistics-barcode/blob/3a098684024cd546da5539de41275c9786b8c396/web_stock/static/src/js/barcode.js#L63), which I have now tested with an RFID reader & 50 labels in the same scan. `self.barcodeHistory` is the queue.

Note that the RFID reader works a bit differently, so there was some additional code to split the stream that's not shown here. Totally not relevant though IMO.

angelmoya:
Thanks @lasley I will check it, I need to see it in action to understand it.

The problem that I want to prevent is to do to specific development for one solution. this is why I like scenario player.

You can use an history for all barcode read, but finally you need to process it.

As I see is that you can do process in several ways, but finally in one step, you need to read one barcode or more than one, If you need to read one barcode you can wait to process it. If you need to read more than one, for me it's ok to have a list with all this barcodes antes de procesarlos.

Anyway I need to check your development before to be sure that I understand everything

angelmoya:
I'm working with this module, and creating new scenarios. I also check @lasley repo, it's nice but is not the same. Anyway I think that we can define new step scenario using this JS.

Maybe could be nice to do a conference together to discuss it.

IMO this PR could be merged, and we can create new PRs to improve it

LoisRForgeFlow:
I've just pushed a first idea to do what @angelmoya said about the numpad. With similar approach, it would be possible to make the web a little more configurable from the hardware form view.

Whereas for the conference, it could be a good idea, @jbeficent and me are ok with it.

angelmoya:
OK, I just see your commit, looks good, but could be beter show numpad in original view instead of in a pop up.

LoisRForgeFlow:
Yes, but at the moment the <a href="http://a.kabachnik.info/numeric-keypad-for-jquerymobile.html">numpad</a> I'm using only has options to pop up, none of them are a permanently displayed one.

angelmoya:
@lreficent and @lasley I had been working a lot with this module and also with web_stock.

I think two modules are necesaries, this for work with "PDA" and define process.
And web_stock for work in fixed table with PC connected to barcode scanner and show more information.

The problem that we have now with stock_scanner_web, is that we are trying to do all things over the ssh process focus. So we need to check message as an array and we have not to much options to improve.

I think we need to split stock_scanner in two modules, stock_scanner_base and stock_scanner_ssh, this way we can get base stock_scanner module and then we will be more free to improve stock_scanner_web.

lasley:
@angelmoya Your proposal to split the modules makes a lot of sense 👍

JordiBForgeFlow:
I would first double check with @sylvain-garancher, as stock_scanner is a well consolidated module.

angelmoya:
I'm working a lot with this stock_scanner_web and it's really powerfull, but the problem is that had a lot of hardcode tricks neededs to work with ssh environment...
For example, it use strings arrays, and each array is one line, we can do better on html. And if string start with '|' then it's a title....
Sure, we need to talk with @sylvain-garancher for do this

Garamotte:
I'm not against changing some old weird things in the `stock_scanner` module, or splitting it in two.

I think the main concerns would be about retro-compatibility and *cross-client* compatibity, but you can decide that `stock_scanner_web` will not be able to run "old-style" scenarios, or later to write scenarios that use features only available in one of the clients (like using HTML strings, which could work, but will look ugly in the ncurses client).

There are not much hardcoded things in the module itself. The client-specific parts are mostly located in the scenarios (which are simply data), so the main goals of such a refactoring would be to provide helpers to help keeping scenarios compatible with various clients without hacks.

angelmoya:
Yes @sylvain-garancher , finally I think that if we want to take advantage of html in stock_scanner_web, scenarios will not be compatible with ncurses... :(
It's not easy to fing god solution to this

Garamotte:
The best solution I found for now is to provide some functions in the core module.

For example, the res value could be generated by a "Markdown (or any other formatting language, to be defined) to scenario" function, which by default returns the value as plain text, and, depending on the client, could return an html formatted string, or anything else.
The type of client (text, html, other) could be declared in the `scanner_check` method (keeping plain text as default value, for retro-compatibility), if we don't want to attach that to specific hardware configurations.

angelmoya:
@sylvain-garancher @lreficent @lasley @jbeficent  maybe we can talk about in Barcelona OCA Sprintcode.

I will start to play with new type of scenario, playable only in web, that gets directly html or template, because I need more than one input in one scenario for do process faster.

lasley:
@angelmoya - I'm too far to hit the Barcelona Sprint, so I won't be seeing you guys until Belgium. 

I fully agree that we need a scenario type specific to web, which directly receives the templates. This is a wall I have been hitting for some time, and is why I have the parallel PR in #38 that I keep forgetting to work on.

angelmoya:
@lasley  I will see you on Belguim then...

I'm testing new improvement to load templates on  steeps... and works nice, I will prepare a PR to this branch, also with demo scenario.

I tested also #38

Garamotte:
Sorry, I won't be present for the Barcelona sprint either.

angelmoya:
superseded by #95

MiquelRForgeFlow:
If this PR was superseeded, then shouldn't it be closed?

PR #62, opened 2016-12-27
[9.0][FIX] stock_scanner: Fix lint

lasley:
This fixes our 9.0 build, which is currently failing due to Lint.


lasley:
Assuming we go green, I am just going to merge this. It's only carriage returns & is fixing a currently red build on a production branch

PR #60, opened 2016-12-27
[10.0][ADD] barcodes_ean14: EAN 14 Barcode Nomenclature

lasley:
This adds GTIN (EAN14) barcode nomenclature to both Python and JavaScript for full compliance with the new barcodes interfaces.

This is a forward migration of #36 + a rename and removal of the abstract validator, which no longer seems useful.

PR #63, opened 2016-12-27
[10.0] Port product_barcode_generator

JosDeGraeve:


product_barcode_generator/__init__.py
```diff
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 ###############################################################################
```
pedrobaeza, line 2: Shorten headers like in https://github.com/OCA/maintainer-tools/blob/master/template/module/__init__.py

pedrobaeza:
Check Travis

product_barcode_generator/models/product.py
```diff
@@ -30,67 +30,67 @@ def isodd(x):
 class ProductCategory(models.Model):
     _inherit = 'product.category'
 
-    ean_sequence_id = fields.Many2one('ir.sequence', string='Ean sequence')
+    barcode_sequence_id = fields.Many2one(comodel_name='ir.sequence',
+                                          string='Ean sequence')
```
Garamotte, line 34: Please add the `oldname` attribute for renamed fields.

product_barcode_generator/__manifest__.py
```diff
@@ -23,19 +23,19 @@
...
-        "data/ean_sequence.xml",
+        "data/barcode_sequence.xml",
         "views/res_company_view.xml",
         "views/product_view.xml",
         "views/sequence_view.xml",
     ],
-    'installable': False,
-    "active": False,
+    'installable': True,
+    "active": True,
```
Garamotte, line 40: Remove the `active` key.

product_barcode_generator/models/product.py
```diff
@@ -30,67 +30,67 @@ def isodd(x):
...
 
 
 class ProductProduct(models.Model):
```
Garamotte, line 37: Rename the class to `ProductTemplate`.

product_barcode_generator/models/product.py
```diff
@@ -30,67 +30,67 @@ def isodd(x):
...
-    _inherit = 'product.product'
+    _inherit = 'product.template'
+
+    @api.one
```
Garamotte, line 40: There is no need to loop here.
Use `@api.multi` instead of `@api.one`.

product_barcode_generator/models/product.py
```diff
@@ -30,67 +30,67 @@ def isodd(x):
...
+    def generate_barcode(self):
+        return self.product_variant_ids.generate_barcode()
+
 
-    ean13 = fields.Char(copy=False)
-    ean_sequence_id = fields.Many2one('ir.sequence', string='Ean sequence')
+class ProductProduct(models.Model):
+    _inherit = 'product.product'
 
     @api.model
```
Garamotte, line 48: Change the decorator to `@api.multi` when you read values from fields.

product_barcode_generator/models/product.py
```diff
@@ -30,67 +30,67 @@ def isodd(x):
...
-    _inherit = 'product.product'
+    _inherit = 'product.template'
+
+    @api.one
+    def generate_barcode(self):
+        return self.product_variant_ids.generate_barcode()
+
 
-    ean13 = fields.Char(copy=False)
-    ean_sequence_id = fields.Many2one('ir.sequence', string='Ean sequence')
```
Garamotte, line 40: Why did you remove this field ?

product_barcode_generator/models/product.py
```diff
@@ -30,67 +30,67 @@ def isodd(x):
...
-    def _get_ean_next_code(self, product):
-        sequence_obj = self.env['ir.sequence']
-        if product.ean_sequence_id:
-            ean = sequence_obj.next_by_id(product.ean_sequence_id.id)
-        elif product.categ_id.ean_sequence_id:
-            ean = sequence_obj.next_by_id(product.categ_id.ean_sequence_id.id)
-        elif product.company_id and product.company_id.ean_sequence_id:
-            ean = sequence_obj.next_by_id(
-                product.company_id.ean_sequence_id.id)
-        elif self.env.context.get('sequence_id', False):
```
Garamotte, line 52: Why did you remove the context sequence_id lookup ?

product_barcode_generator/models/product.py
```diff
@@ -30,67 +30,67 @@ def isodd(x):
...
                   "the 13 being used as a control digit"
                   "You will have to redefine the sequence or create a new one")
-                )
+            )
 
-        return ean
+        return barcode
 
-    def _get_ean_control_digit(self, code):
+    def _get_barcode_control_digit(self, barcode):
```
Garamotte, line 71: This method is now available in the standard model `barcode.nomenclature` of the `barcodes` module, maybe you could use it ?
But this would require a new dependency...

product_barcode_generator/views/sequence_view.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<openerp>
+<odoo>
     <data>
```
Garamotte, line 3: Remove the `<data>` tag.

product_barcode_generator/views/res_company_view.xml
```diff
@@ -1,22 +1,23 @@
 <?xml version="1.0" encoding="utf-8"?>
-<openerp>
+<odoo>
     <data>
```
Garamotte, line 3: Remove the `<data>` tag.

product_barcode_generator/views/product_view.xml
```diff
@@ -1,35 +1,52 @@
 <?xml version="1.0" encoding="utf-8"?>
-<openerp>
+<odoo>
     <data>
```
Garamotte, line 3: Remove the `<data>` tag.

Garamotte (changes requested):

product_barcode_generator/data/barcode_sequence.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
```
lasley, line 2: This can be simplified as `<odoo noupdate="1">` & can remove data tag

product_barcode_generator/models/product.py
lasley: Additionally, api.one is deprecated so we can't use it anymore 😦

product_barcode_generator/models/product.py
```diff
@@ -30,67 +30,67 @@ def isodd(x):
...
-            ean = sequence_obj.next_by_id(product.ean_sequence_id.id)
-        elif product.categ_id.ean_sequence_id:
-            ean = sequence_obj.next_by_id(product.categ_id.ean_sequence_id.id)
-        elif product.company_id and product.company_id.ean_sequence_id:
-            ean = sequence_obj.next_by_id(
-                product.company_id.ean_sequence_id.id)
-        elif self.env.context.get('sequence_id', False):
-            ean = sequence_obj.next_by_id(self.env.context.get('sequence_id'))
+    def _get_barcode_next_code(self, ):
+        if self.categ_id.barcode_sequence_id:
```
lasley, line 50: Looks like this expects a singleton, will need a `self.ensure_one()`

product_barcode_generator/models/product.py
lasley: I think the dependency is worth it TBH. Will allow for this to fit with additions to it like #60

product_barcode_generator/models/product.py
```diff
@@ -30,67 +30,67 @@ def isodd(x):
...
+    def _generate_barcode_value(self):
+        barcode = self._get_barcode_next_code()
+        if not barcode:
             return None
-        key = self._get_ean_control_digit(ean)
-        ean13 = ean + key
-        return ean13
+        return barcode + self._get_barcode_control_digit(barcode)
 
     @api.one
```
lasley, line 88: api.one is deprecated

lasley (commented):
Thanks @JosDeGraeve - few comments inline

product_barcode_generator/models/product.py
lasley: We may also want to change the string attribute if we're abstracting the definition of the field for barcode vs just ean.

lasley:
On second thought, I think [`barcodes_generator_product`](https://github.com/OCA/stock-logistics-barcode/tree/9.0/barcodes_generator_product) is the successor to this module

cc @legalsylvain

legalsylvain:
indeed. @josdegraeve : Did you taken a look on this module ? It is based on barcode nomenclature core feature and allow sequenced or manual numerotation.

lasley:
I went ahead and just upgraded them all for good measure, I was about to need 2 of the 3 anyways. #65, #66, #67

lasley:
Closing, #66 supersedes because this module is deprecated & removed

PR #65, opened 2016-12-28
[10.0][MIG] barcodes_generator_abstract

lasley:
Migration of `barcodes_generator_abstract` to v10. Was simple enough:

* Rename manifest
* Change openerp references to odoo
* Bump version
* Alphabetize imports

Regarding the coveralls- while it may not be incredibly good practice, I propose we let the other modules cover the testing gap that is introduced here. 

As you can see in #66, the drop is way less significant, and I would guess that adding the partner module probably covers the rest. Just my 2 cents on allowing me less work really 😛

PR #68, opened 2016-12-28
[10.0][ADD] barcodes_generator_package

lasley:
* Create new module to generate barcodes for product packaging.

Depends:
- [x] #65


barcodes_generator_package/README.rst
```diff
@@ -0,0 +1,100 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+===============================
+Generate Barcodes for Packaging
+===============================
+
+This module expends Odoo functionality, allowing user to generate barcode
```
pedrobaeza: s/expends/expands

barcodes_generator_package/README.rst
```diff
@@ -0,0 +1,100 @@
...
+depending on a given barcode rule for Product Packaging.
+
+For exemple, a typical pattern for partners is  "042........." that means
```
pedrobaeza: s/exemple/example

barcodes_generator_package/README.rst
```diff
@@ -0,0 +1,100 @@
...
+* a 13 digit control
+
+With this module, it is possible to:
+
+* Affect a pattern (barcode.rule) to a product.packaging
+
+* Define a Barcode base: 
+    * manually, if the base of the barcode must be set by a user. (typically an
+      internal code defined in your company)
+    * automaticaly by a sequence, if you want to let Odoo to increment a
```
pedrobaeza: s/automaticaly/automatically

barcodes_generator_package/__manifest__.py
```diff
@@ -0,0 +1,28 @@
...
+    "summary": "Generate Barcodes for Product Packaging",
+    "version": "10.0.1.0.0",
+    "category": "Tools",
+    "author": "LasLabs, "
+              "Odoo Community Association (OCA)",
+    "website": "https://www.odoo-community.org",
+    "license": "AGPL-3",
+    "depends": [
+        "barcodes_generator_abstract",
+        "product",
```
pedrobaeza: You don't need this explicit dependency

pedrobaeza (changes requested):
You should use an icon with a light variation over the use one, as too many modules will have the same icon. See for example https://github.com/OCA/vertical-association/blob/8.0/membership_prorrate/static/description/icon.png, https://github.com/OCA/vertical-association/blob/8.0/membership_prorrate_variable_period/static/description/icon.png and https://github.com/OCA/vertical-association/blob/8.0/membership_variable_period/static/description/icon.png for the variations I made for three modules that are related.

legalsylvain:
@pedrobaeza : seems that the new commits fixes your remarks.

legalsylvain (approved):
:+1:

pedrobaeza (approved):

PR #70, opened 2017-01-12
[10.0][ADD] barcodes_generator_abstract: Isolate encoding types

lasley:
I'm in a predicament right now with EAN-14, which is encoded as simply `ean` in pyBarcodes. This PR adds a new column allowing for customization of the barcode type as pyBarcodes sees it - with pass-thru for the standard encoding way that works with most everything else.

* Add generation encoding type with selection of supported types
* Add compute/inverse to allow pass-thru of normal encoding if custom not selected

cc @legalsylvain 

Tests to come, this was primarily to gather thoughts on implementation or other strategies.

Depends:
- [x] #65

### DIFF
--- a/barcodes_generator_abstract/models/barcode_generate_mixin.py
+++ b/barcodes_generator_abstract/models/barcode_generate_mixin.py
@@ -53,7 +53,7 @@ class BarcodeGenerateMixin(models.AbstractModel):
             if custom_code:
                 custom_code = custom_code.replace('.' * padding, str_base)
                 barcode_class = barcode.get_barcode_class(
-                    item.barcode_rule_id.encoding)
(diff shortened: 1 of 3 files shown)


lasley:
Note I ended up having to add EAN-14 to pyBarcodes anyways. Their GTIN was incorrectly using EAN-13, so I wasn't getting the desired numbers. This is the PR - https://bitbucket.org/whitie/python-barcode/pull-requests/11/feature-ean14/

Regardless of that, I think this change is still worth considering. Even though the names now align for me, it may be an issue going forward.

lasley:
This has become irrelevant now that I'm a maintainer of the barcodes lib we use, closing!

PR #72, opened 2017-01-16
[FIX] Travis file.

legalsylvain:
product_multi_ean has been disabled so no need to manage exclude / include travis configuration.


pedrobaeza (approved):
Fast-tracking

PR #75, opened 2017-01-18
[9.0][FIX] barcodes_generator_product: Propagate barcode rule to all product variants.

sergio-teruel:
If change or when create a product with variants the rule not is propagted to all variants.
cc @Tecnativa

### DIFF
--- a/barcodes_generator_product/demo/product.xml
+++ b/barcodes_generator_product/demo/product.xml
@@ -36,8 +36,4 @@ License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
         <field name="attribute_value_ids" eval="[(6,0,[ref('product.product_attribute_value_4')])]"/>
     </record>
 
-    <record id="product_template_multi_variant" model="product.template">
-        <field name="name">Template with Generated Barcode (Multi Variant)</field>
-    </record>
-
 </odoo>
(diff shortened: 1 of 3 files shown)


pedrobaeza (changes requested):
One thing I don't get: the field is only defined at template level. Why do you need to propagate it then?

sergio-teruel:
@pedrobaeza Mhuuu but product_product inherits from barcode_mix_in so barcode_rule is in product_product too.

pedrobaeza:
OK, maybe the fix should be to remove that inheritance. Anyway, you should propagate always the rule, having one variant (default) or having several, isn't it?

lasley (changes requested):
IMO there should be an option to opt-in for this functionality. I have a few use cases where I am also barcoding the variants, which would conflict with this.

legalsylvain (changes requested):
Agree with @lasley. variants has differents barcodes. This feature you introduces is breaking Odoo design, that allow a distinct barcode per variant.
I think that if you need this feature, you'll have to create a new extra module for that.

legalsylvain:
Well, the community doesn't seem to be agree with that design. I close this PR. Feel free to reopen it if design changed.

kind regards.

PR #77, opened 2017-01-31
[10.0][IMP] barcodes_generator_abstract: Add automated generator

lasley:
* Add an option to allow for automatic generation of barcode when a record of mixin model is created

This is dependent on the migration PR:
- [x] #65 

I'd like the two reviewed independently though to allow for easier back port. Relevant commit 	02d2303

Still need to test more & add actual tests, so consider this a PoC.

cc @legalsylvain

### DIFF
--- a/barcodes_generator_abstract/models/barcode_rule.py
+++ b/barcodes_generator_abstract/models/barcode_rule.py
@@ -69,7 +69,6 @@ class BarcodeRule(models.Model):
 
         It also clears the cache of automated rules if necessary.
         """
-        should_clear = False
         for record in self:
             if not record.generate_automate:
                 continue
@@ -84,9 +83,17 @@ class BarcodeRule(models.Model):
                     'Only one rule per model can be used for automatic '
                     'barcode generation.'
                 ))
-            should_clear = True
-        if should_clear:
-            self.clear_caches()
+
+    # CRUD
+    @api.model
+    def create(self, vals):
+        self._clear_cache(vals)
+        return super(BarcodeRule, self).create(vals)
+
+    @api.multi
+    def write(self, vals):
+        self._clear_cache(vals)
+        return super(BarcodeRule, self).write(vals)
 
     # View Section
     @api.multi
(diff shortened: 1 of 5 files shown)


lasley:
Rebased now that dependency is merged, this is ready for review!

legalsylvain:
:+1: great improvment ! Thanks.
(code review, no test.)

legalsylvain (approved):

barcodes_generator_abstract/models/barcode_rule.py
```diff
@@ -54,6 +61,33 @@ def onchange_generate_type(self):
...
+                ('generate_automate', '=', True),
+            ]
+            if len(self.search(domain)) > 1:
+                raise exceptions.ValidationError(_(
+                    'Only one rule per model can be used for automatic '
+                    'barcode generation.'
+                ))
+            should_clear = True
+        if should_clear:
+            self.clear_caches()
```
pedrobaeza: Why don't you clear this on create/write when vals contains 'generate_automate' value?

lasley: Hmmmm yeah that would work too, although then we'd be splitting the logic into two spots & it would need to be a conjoined if - because this is also clearing cache if `generate_model` is changed too.

I could remove the somewhat ghetto `should_clear` as `if len(self)` instead though, which would remove two lines and be the same level of efficiency.

I do somewhat take reservation to a mutating operation in a constraint though, so it might still be best to bite the bullet and move into the two spots. Any additional thoughts?

pedrobaeza: For keeping the same logic, just create a new method that is called from both write/create methods. I still miss the wanted @onwrite decorator: https://github.com/odoo/odoo/pull/11042

lasley: Oh wow I love this `@onwrite` idea. It also reminded me that I needed to create the API lib from OCA/server-tools#615. I think I'll try to tackle that this weekend.

Also, yeah cool a helper method seems almost excessive, but it is better than code duplication. It's also better than a mutation in a constraint, even if the mutation is just a cache deletion.

pedrobaeza (approved):

PR #80, opened 2017-02-18
[10.0][ADD] barcodes_generator_stock_picking: Create new module

lasley:
This module adds the abstract barcode generator amazingness onto `stock.picking`

Generate Barcodes for Stock Pickings
====================================

This module expands Odoo functionality, allowing user to generate barcode
depending on a given barcode rule for Stock Picking.

For example, a typical pattern for partners is  "042........." that means
that:
* the EAN13 code will begin by '042'
* followed by 0 digits (named Barcode Base in this module)
* a 13 digit control

With this module, it is possible to:

* Assign a pattern (barcode.rule) to a stock.picking

* Define a Barcode base: 
    * manually, if the base of the barcode must be set by a user. (typically an
      internal code defined in your company)
    * automatically by a sequence, if you want to let Odoo to increment a
      sequence. (typical case of a customer number incrementation)

* Generate a barcode, based on the defined pattern and the barcode base

Configuration
=============

To configure this module, see the 'Configuration' Section of the description
of the module 'barcodes_generator_abstract'

Usage
=====

To use this module, you need to:

* Go to a Picking form:

1 for manual generation
    * Set a Barcode Rule
    * Set a Barcode Base
    * click on the button 'Generate Barcode (Using Barcode Rule)'

2 for automatic generation
    * Set a Barcode Rule
    * click on the button 'Generate Base (Using Sequence)'
    * click on the button 'Generate Barcode (Using Barcode Rule)'


legalsylvain:
Hi @lasley.
Thanks for this work. LGTM. Just a remark about the name, I think that barcodes_generator_picking will be more consistent. 
(other modules are xxx_partner, xxx_product, xxx_lot, xxx_location).

What do you think ?

lasley:
@legalsylvain - agreed, and done!

legalsylvain (approved):
Thanks ! :+1:

hparfr (approved):

PR #81, opened 2017-02-18
[10.0][REM] tr_barcode*: Remove deprecated modules

lasley:
There are a lot of `tr_barcode*` modules that seem like they haven't received any attention over the years, but also seem to now be duplicating functionality provided in `barcodes_generator_abstract` & related modules. I propose we remove these modules in the v10 branch, unless someone has some objection to this.

My reasoning behind this proposal is due to confusion that we already had with other modules similar to this that were also deprecated (such as #63). Last thing we want is a contributor wasting a bunch of time upgrading the wrong module IMO.

The code will remain in the v9 branch, and previous branches - where it may even be more relevant. From my high level scans of all this though, it is all not relevant in v10. 

Thoughts?


elicoidal:
:+1:

pedrobaeza (approved):
And they still contains a reference to the developer company...

lasley:
Wow I spent so long trying to figure out what the TR prefix meant from a technical perspective. In hindsight, it's so obvious now!

PR #82, opened 2017-02-19
[10.0][IMP] barcodes_generator_abstract: Add search

lasley:
This PR adds a centralized search method on `barcode.nomenclature` for barcode rules implementing the abstract generator. 

This works by first parsing the barcode against the nomenclature to determine the correct rule, then  using the `generate_model` field to identify which model to search against the `barcode` field on said model. The matched recordset is then returned.

This is very tightly coupled with the generator, so I figured it made more sense to add it as a feature vs. creating a whole new module. No harm in the extra method if someone's not using it IMO.

### DIFF
--- a/barcodes_generator_abstract/README.rst
+++ b/barcodes_generator_abstract/README.rst
@@ -76,7 +76,46 @@ model.
 Alternatively, you can develop a custom module for a custom model. See
 'Inheritance' parts.
 
-Try this module on Runbot
+Search By Barcode
+-----------------
+
+A wizard is available in the Barcode Nomenclature's action menu that allows for
+you to search for records using a barcode.
+
+For developers, there are two handy methods in `barcode.nomenclature` as well:
+
+.. code-block:: python
+
+    @api.multi
+    def find_by_barcode(self, barcode):
+        """Return the record associated with the barcode.
+
+        Args:
+            barcode (str): Barcode string to search for.
+
+        Returns:
+            BaseModel: A record matching the barcode, if existing.
+            None: No match.
+        """
+
+.. code-block:: python
+
+    @api.multi
+    def get_form_action_for_barcode(self, barcode):
+        """Return the form action for the record related to barcode.
+
+        Args:
+            barcode (str): Barcode string to search for.
+
+        Returns:
+            dict: Default form action dictionary for the barcode.
+
+        Raises:
+            UserError: If no match was found for the barcode.
+        """
+
+Try On Runbot
+-------------
 
 .. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
    :alt: Try me on Runbot
--- a/barcodes_generator_abstract/__init__.py
+++ b/barcodes_generator_abstract/__init__.py
@@ -1,2 +1,3 @@
 # -*- coding: utf-8 -*-
 from . import models
+from . import wizards
--- a/barcodes_generator_abstract/__manifest__.py
+++ b/barcodes_generator_abstract/__manifest__.py
@@ -7,7 +7,7 @@
 {
     'name': 'Generate Barcodes (Abstract)',
     'summary': 'Generate Barcodes for Any Models',
-    'version': '10.0.1.0.0',
+    'version': '10.0.1.0.1',
     'category': 'Tools',
     'author':
         'GRAP, '
--- a/barcodes_generator_abstract/__manifest__.py
+++ b/barcodes_generator_abstract/__manifest__.py
@@ -23,6 +23,7 @@
         'security/res_groups.xml',
         'views/view_barcode_rule.xml',
         'views/menu.xml',
+        'wizards/barcode_search_view.xml',
     ],
     'demo': [
         'demo/res_users.xml',
--- a/barcodes_generator_abstract/models/__init__.py
+++ b/barcodes_generator_abstract/models/__init__.py
@@ -1,4 +1,4 @@
 # -*- coding: utf-8 -*-
 from . import barcode_generate_mixin
-
+from . import barcode_nomenclature
 from . import barcode_rule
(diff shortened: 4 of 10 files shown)


legalsylvain:
Hi @lasley, 
Interesting feature ! As I see that find_by_barcode is not called in your code (except in the test), I suggest to add a wizard to use this feature. I mean : 
A new menu item, opening a wizard "Search Item by barcode".
if the item is found, the wizard could open the form view of the item found.

what do you think ?

lasley:
@legalsylvain hahahah dangit I was hoping nobody would call me out on that. It doesn't fit what I specifically need, but I'll admit I tried it. I'll take another stab though, because you're totally right that it's needed.

My two problems:
* Where to put the wizard in the UI
* How to identify the default form action for a record

lasley:
Hmmm I have an intriguing idea on placement actually. What do you think if I added a button with a barcode handler in the main menu, similar to how I did for the Red October connector:

image

Click the button, scan or enter a barcode into the text area in the dropdown, enter will take you to the form? In this sense, we would almost have the Enterprise barcodes interface.

legalsylvain:
About the UI, yes, it will be not very good, because barcode_generator_abstract depend on base barcodes only, that depends on 'base', so we don't have any menu item for basic users. (since sale menu disappeared).
I propose to add a menu item in a settings / technical menu for administrator purpose.

After, in other modules, we could add extra menu item, with the action defined in this menu.
We could for exemple, in xxx_lot, xxx_location and xxx_package, define a menu, with a xml_id = 'barcodes_generator_abstract.menu_stock_search_by_barcodes' so even if the three modules are installed, there will be only one menu item in 'stock' parts.

Maybe we could add a group 'search_by_barcodes', to avoid to display useless feature for some users. 

point 2 : self.env['ir.ui.view'].search([('model', '=', MODEL), ('type', '=', 'form')], limit=1, order='sequence desc')
don't you think ?

legalsylvain:
sorry, i didn't see your answer. Yes, your idea sound's good !

lasley:
Great idea with the security for the displaying of the barcodes menu! Also, I believe your view search will work too 🎉 

I'll scope the changes and slate them on our dev schedule. This will be an awesome addition IMO

lasley:
Note for implementation, there are actually methods for this as long as you have a recordset:
```
>>> dir(recordset)
...
 '_get_default_calendar_view',
 '_get_default_form_view',
 '_get_default_graph_view',
 '_get_default_kanban_view',
 '_get_default_pivot_view',
 '_get_default_search_view',
 '_get_default_tree_view',
...
```

legalsylvain:
Hi @lasley. Is it still WIP ? i just tried on runbot, but didn't found how it works. 

thanks.

lasley:
Hey @legalsylvain - Yeah I gave this a shot the other day in some spare time & am running into issues. I'm going to remove the Needs Review label, my bad on leaving it up.

Regarding how to use, I added the wizard in the `More` menu of the `barcode.nomenclature` model.

rousseldenis:
@lasley Can you review travis errors ?

legalsylvain:
Hi @lasley. 
I'd like to use this search feature, but in a more generic way. (not depending on generator module). In fact, the search by barcode should be used even if generator is not used. (and if it is installed, it should work even if generate_model is not defined). For the time being, the search will fail for barcodes that are not generated via the generator feature)

I'd like to refactor your PR to allow this feature to work just with a search button in the header menu.
Are you agree if I refactor your PR ?
What I see : 

- creating a new module barcode_search (depending on barcodes)
- move the search code in the barcode rule in the barcode search transient model
- change the search algorithm, searching on all the barcode field of all the models present in the database.

+ adding a feature to allow to compute quantity (or price) based on some specific barcodes (in PoS). (for some products, there are a pattern like 22xxxxx{NNNDD} where NNNDD could be a price, or a weight) so the search function should matches and return the product 221234500000 if 221234503678 is entered).

Thanks for your comments.

lasley:
@legalsylvain I agree with your motivation, and would love to see this adapted! 

> change the search algorithm, searching on all the barcode field of all the models present in the database.

Here's where I ran into difficulties when I tried this (my plan was a solo module before I just did it here). We need a way to identify that a column is the barcode column for a model. That's the only thing that binds my logic to this module, because it uses the generator to identify the column to search.

We could maybe make a new model `barcode.model`, which could hold model to field name combos. Find the field name for the model, then perform the proper searches. Modules looking to add a supported barcode would simply need a data file defining the model/field combo.

I think everything else would just fall in place? We may need `barcode.model` to also hold the concept of the quantity and price fields in order to support the last point.

legalsylvain:
Hi @lasley. 
Thanks for your answer.

Well, about identifying barcode field, you're right, but since V9+, the field barcode seems to be named allways 'barcode'. (before, in 8 -, the name was ean13 for product.product, for example).
The proper module barcodes_generator_abstract **is based on this predicate** : ref : 
https://github.com/OCA/stock-logistics-barcode/blob/10.0/barcodes_generator_abstract/models/barcode_generate_mixin.py#L72
If the field name is not 'barcode' we have a problem anyway.

So I was thinking about : 
- get from your PR the model barcode.search.
 add a function @api.model get_models_with_barcode that return a list of (model_name, field_name), by default with all models that has a field 'barcode'. So if stock is installed, something like [('product.product', 'barcode'), ('stock.location', 'barcode')] ...
- make the search base on the given models. (just take care that the current user has readaccess on the model, and the field)
- if a custom module add a custom field, we could just add a key in this list.

barcode.model should be great, but I think it will be not so good to do it in a OCA modules. Should be defined as a mixin in the Odoo Core with all the related features. (having a field 'barcode', make check about syntax, ...)  (V16 ? ;-))

About the last point (barcode with price / weight info), I was thinking about : 
- make a search on all the models, based on the models provided by get_models_with_barcode. 
- If not found, check barcode.rule with type in ('weight' or 'price') (defined in point_of_sale module or 'price_to_weight', defined in the OCA / pos module 'price_to_weight' (ref : https://github.com/OCA/pos/blob/10.0/pos_price_to_weight/models/barcode_rule.py#L13)
- make a search, replacing the text placed on the pattern {NNNDD} by '00000'
For the time being, there are only product model that has this special behaviour. (many barcodes match 1 item, with extra data in the barcode)

legalsylvain:
superseded by https://github.com/OCA/stock-logistics-barcode/pull/121

PR #84, opened 2017-02-19
[9.0][FIX] travis: Switch after_success statement

lasley:
* Switch after_success from deprecated coveralls to `travis_after_tests_success`


pedrobaeza (approved):
Fast-tracking it

PR #85, opened 2017-02-19
[ADD] barcodes_generator_lot: Production lot serialization

lasley:
This is a WIP PR adding barcode generation to stock production lots. In this instance, the `barcode` field is pointed to `name` so that it is being used as the serial number for the lot.

Paired with the new automatic barcoding functionality, this effectively allows for automatic serialization of lots via nomenclature rules.


pedrobaeza:
Remove the partner image

lasley:
Good catch, thanks @pedrobaeza - I updated it to a picking image & fixed the remaining issues. This is now ready for final review!

pedrobaeza:
The same thing about the icon

lasley:
Is the icon not generic enough to fit all barcode generation modules? Seems like it'd be better than the default OCA one IMO & I'm not really sure how to differentiate a picking from a partner in something small like an icon.

pedrobaeza:
The problem is the same icon for all the models (excepting the generic OCA one). The goal of the icon is to visually differentiate the module you want to install. This way you don't get the purpose.

lasley:
Hmmm alright then. These are already aligned in the other barcode modules though, so we would be creating a fractured icon environment of sorts:
* [product](https://github.com/OCA/stock-logistics-barcode/blob/9.0/barcodes_generator_product/static/description/icon.png)
* [partner](https://github.com/OCA/stock-logistics-barcode/blob/9.0/barcodes_generator_partner/static/description/icon.png)
* [abstract](https://github.com/OCA/stock-logistics-barcode/blob/9.0/barcodes_generator_abstract/static/description/icon.png)

lasley:
No worries answering me, I just started working through my other PRs & noticed your other comments. I'll make a variation.

lasley:
Hrmmmm any ideas on what a "Production Lot" looks like visually? Everything I'm coming up with looks like a package & this seems like kind of a cop-out 😆 

image

pedrobaeza:
You can put a label like this one with the lot word inside: https://openclipart.org/detail/3755/roped-label

SVG versions of the icons would be very appreciated.

lasley:
> You can put a label like this one with the lot word inside

Cool that's better than just putting the word, thanks!

> SVG versions of the icons would be very appreciated.

I agree, but the existing assets are all PNG. I went to the [backlink in the ReadMe](http://www.iconarchive.com/show/oxygen-icons-by-oxygen-icons.org.html) & it looks like only PNG assets are provided there too 😢

pedrobaeza:
You can always make an SVG with the source bitmap + the label & text in vector

barcodes_generator_lot/README.rst
```diff
@@ -0,0 +1,103 @@
...
+Generate Barcodes for Stock Production Lots
+===========================================
+
+This module expands Odoo functionality, allowing user to generate barcode
+depending on a given barcode rule for Stock Production Lots.
+
+The barcode will be used as the ``name`` of the lot, which effectively allows
+this module to create automatic serialization rules for production lots.
+
+For example, a typical pattern for partners is  "042........." that means
```
legalsylvain: Are there some nomenclature for lots ? If, yes maybe you could change the description of the module. (partner sample is not very relevant for that module).
(just a question, I don't use lot)

legalsylvain (approved):
One minor remarks in readme file. LGTM otherwise.

Code review / no test.

barcodes_generator_lot/README.rst
lasley: I think I just missed a reference to the module I copied the ReadMe from. `042` isn't specific to partners in some sort of standard is it? I admittedly used this in all of my ReadMe's because I was too lazy to change from the one you had.

legalsylvain: I don't know exactly standards. I know that some code are reserved, but I don't know if there officials rules.
Not a problem. :+1:

legalsylvain:
@pedrobaeza is it now ok for you ?

pedrobaeza:
The icon is still the generic one.

legalsylvain:
Please, @pedrobaeza, it should be great not to block a PR because of svg / png.

It's a 'eventually nice to have feature', and can be added after, if people has time for that.

I think that OCA members should focus on features (code) and quality (coverage / travis). Regarding that points, that PR seems OK for me.

My 2 cents.

pedrobaeza:
Well, I know that @lasley will fix this as soon as he see this. If he say that he can't, then I won't block it of course, but you know that things that are not done now, then are forgotten later.

lasley:
Hahaha yup, I think my Chrome crashed recently or something & I lost all my To-Do tabs. I feel like I've missed a lot of comments. I should probably take this time to come up with a better To-Do strategy... maybe something with the Github connector. 

Fixed!

pedrobaeza (approved):

pedrobaeza:
Thanks, Dave!

PR #95, opened 2017-05-29
[9.0] ADD stock_scanner_web

angelmoya:


LoisRForgeFlow:
@angelmoya let us know then this PR is ready!

stock_scanner_web/controllers/main.py
```diff
@@ -0,0 +1,177 @@
...
+
+        scenario_step = request.env['scanner.hardware'].\
+            search([('code', '=', terminal_number)]).step_id
+        step = scenario_step and int(scenario_step) or 0
+        header = False
+        lines = False
+        if code == 'L':
+            header = ''
+            lines = []
+            for line in result:
```
lmignon, line 131: Where is `result` initialized?

angelmoya: line 98

lmignon:
@angelmoya How do you ensure that the steps flow defined by a scenario is enforced by your implementation? I don't find into your implementation code to check that a step is allowed according to the state of the scenario nor a proper management of the browser history.... I've the feeling that we'll have troubles if the user press back/forward into the web browser or type directly an url... Do I miss something?

angelmoya:
Maybe you are right @lmignon ... If you type directly an url and is not right then the controller will return error stage.

If scenario is right and user use it with interface then it will work, but as you say maybe could be needed to  check restrictions.

lmignon:
@angelmoya Imagine the following scenario Step1 -> Step2 -> Step3 with Step2 configured with `step_back=False` . What's going to happen if the user click 2 times back into the browser when the Step3 is displayed?

angelmoya:
@lmignon I need to check, but I think that step_back is not implemented, only on error step.

I will check step_back implementation

stock_scanner_web/models/scanner_hardware.py
```diff
@@ -0,0 +1,46 @@
...
+from openerp import models, api, fields
+
+
+class ScannerHardware(models.Model):
+    _inherit = 'scanner.hardware'
+
+    show_numpad = fields.Selection(
+        [('yes', 'Yes'),
+         ('no', 'No'),
+        ],
```
MiquelRForgeFlow, line 15: flake8 error here

MiquelRForgeFlow (commented):
@angelmoya, would you greenify this PR?

rousseldenis:
@angelmoya Do you plan to correct travis ?

angelmoya:
We are not working now on it, but I will check flake8 errors

angelmoya:
@rousseldenis , it's done

rousseldenis:
@angelmoya Is this finished ? I think this sould be reviewed and merged.

angelmoya:
I'm not working on this from one year ago, it was working fine, maybe I will migrate to version 11 next monts

PR #96, opened 2017-06-02
[FIX] Fix v10.0 export scenario script

Garamotte:
With the new API migration, some changes were not applied in the script
- The read method now always returns a list (even on a single id)
- The context argument is now at the end, so we need to add the load
    argument when not using named arguments

### DIFF
--- a/stock_scanner/scripts/export_scenario.py
+++ b/stock_scanner/scripts/export_scenario.py
@@ -90,7 +90,8 @@ warehouse_obj = Object(cnx, 'stock.warehouse')
 user_obj = Object(cnx, 'res.users')
 group_obj = Object(cnx, 'res.groups')
 scen_read = scenario_obj.read(
-    int(opts.scenario_id), [], {'active_test': False})
+    int(opts.scenario_id), [], '_classic_read', {'active_test': False})
+scen_read = scen_read and scen_read[0]
 if not scen_read:
     logger.error('Scenario ID %s not found' % opts.scenario_id)
     sys.exit(1)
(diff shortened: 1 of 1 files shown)


Garamotte:
@pedrobaeza @gurneyalex Do you know why the transifex job has been cancelled ?

@lmignon Can review this simple PR ?

pedrobaeza:
I have restarted it

lmignon:
@sylvain-garancher I'll review these fixes ASAP...

lmignon (approved):
:+1: Thank you  @sylvain-garancher ! (Code review and import/export test)

stock_scanner/scripts/export_scenario.py
```diff
@@ -90,7 +90,7 @@ def normalize_name(name):
 user_obj = Object(cnx, 'res.users')
 group_obj = Object(cnx, 'res.groups')
 scen_read = scenario_obj.read(
-    int(opts.scenario_id), [], {'active_test': False})
+    int(opts.scenario_id), [], '_classic_read', {'active_test': False})[0]
```
lmignon: @sylvain-garancher If teh id is not found  `[0]` raise `IndexError: list index out of range` and `if not scen_read:`will never be called
```python
scen_read = scenario_obj.read(
    int(opts.scenario_id), [], '_classic_read', {'active_test': False})
scen_read = scen_read and scen_read[0]
if not scen_read:
    ....
```

lmignon (changes requested):

stock_scanner/scripts/export_scenario.py
Garamotte: Oh, good catch.

Garamotte:
@lmignon Fixed, thanks.

lmignon:
@sylvain-garancher Thank you :+1: 
@rousseldenis Can you confirm that this PR works also for you?

sla-subteno-it:
👍

Garamotte:
@lmignon @rousseldenis Could you review/approve this PR ?

rousseldenis (approved):
:+1: Code review and tested. Works like a charm!

lmignon (approved):

PR #97, opened 2017-06-16
[10.0] product_multi_ean

dion-pci:
Port of product_multi_ean module for Odoo v10 from product_multi_ean module on branch 8.0.


pedrobaeza:
Have you followed migration guide? https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-10.0

product_multi_ean/views/product_view.xml
```diff
@@ -0,0 +1,37 @@
+<?xml version="1.0" encoding="utf-8"?>
+<openerp>
```
elicoidal: odoo everywhere

product_multi_ean/__init__.py
```diff
@@ -1,22 +1,5 @@
 # -*- coding: utf-8 -*-
```
elicoidal, line 1: Can you add a README File following https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst?

elicoidal (changes requested):
Not far from perfect!
Thanks for the contribution

dion-pci:
Thanks for the review. I have updated the file accordingly but still unsure about the unittest and translations.

product_multi_ean/__init__.py
dion-pci: The README file (based from branch 8.0) has been included in the latest commit.

rvalyi:
ping @elicoidal please could you check it it is okay with your review now?

product_multi_ean/models/product_product.py
```diff
@@ -0,0 +1,96 @@
...
+    @api.depends('ean13')
+    def _compute_barcode(self):
+        for product in self:
+            product.barcode = product.ean13
+
+    @api.multi
+    def _inverse_ean13(self):
+        for product in self:
+            if product.ean13_ids:
+                product.ean13_ids[:1].write({'name': product.ean13})
```
legalsylvain, line 75: Why don't you test if the current value doesn't exist in any of the ean13_ids items ?

product_multi_ean/models/product_product.py
```diff
@@ -0,0 +1,96 @@
...
+class ProductProduct(models.Model):
+    _inherit = 'product.product'
+
+    ean13_ids = fields.One2many(
+        comodel_name='product.ean13', inverse_name='product_id',
+        string='EAN13')
+    ean13 = fields.Char(
+        string='Main EAN13', compute='_compute_ean13', store=True,
+        inverse='_inverse_ean13', readonly=True)
+    barcode = fields.Char(
```
legalsylvain, line 54: AFAIK, barcode is now, not surely an ean13.
what about to create a model product.barcode and an one2many barcode_ids ?

legalsylvain (changes requested):
Hi @dion-pci.

thanks for your PR.

Will this module work with PoS, if secondary barcode is scanned.
Some question inlines.

kind regards.

bealdav:
@dion-pci please is there any news here ?
Thanks

product_multi_ean/models/product_product.py
dion-pci: I'm not really sure about changing the model since the original module (or at least, the version for v8) only specify ean13. It may or may not be the original intent of the original developer.

Also, I never worked with PoS before so I'm not sure if this will work with PoS.

product_multi_ean/models/product_product.py
dion-pci: Other than field name & views, I only did minimal changes to the function. Noted for improvement (if you mean an additional check like `if ean13 not in ean13_ids`).

rousseldenis:
As there is more recent PR, closing this.

PR #99, opened 2017-07-13
[8.0][FIX] product_multi_ean: Resolved incompatibilities in search function

kikopeiro:
This patch resolved this error which was gotten searching two products in the same action

image

 ```
 File "/opt/odoo/shared/odoo-server/openerp/api.py", line 268, in wrapper
    return old_api(self, *args, **kwargs)
  File "/opt/odoo/shared/odoo-server/openerp/models.py", line 4497, in _where_calc
    e = expression.expression(cr, user, domain, self, context)
  File "/opt/odoo/shared/odoo-server/openerp/osv/expression.py", line 659, in __init__
    self.expression = distribute_not(normalize_domain(exp))
  File "/opt/odoo/shared/odoo-server/openerp/osv/expression.py", line 223, in normalize_domain
    assert expected == 0, 'This domain is syntactically not correct: %s' % (domain)
AssertionError: This domain is syntactically not correct: [('active', '=', 1), '|', '|', '|', ['default_code', 'ilike', '180.9434.01'], ['name', 'ilike', '180.9434.01'], '|', '|', ['default_code', 'ilike', '180.9434.02'], ['name', 'ilike', '180.9434.02'], ('ean13_ids', 'in', [])]
```


pedrobaeza (changes requested):
Seems correct, but please change version number and squash commits

PR #101, opened 2017-07-29
[10.0][Fix] pip install external dependencies broken

etherealite:
I'm not sure this will fix the (as of this posting) busted build. You can't use pip install on any of these packages right now since the Python Barcode package is busted. Now that you've switched to viivakoodi somehow the packing scripts don't know this change.

PR #104, opened 2017-08-30
[10.0] barcodes_generator_*: add image generation wizard (+ usability improvements)

alexis-via:
Don't use parenthesis in button labels, use help message to give more info.
Hide the button "Generate Barcode" when a barcode is already present.

### DIFF
--- a/barcodes_generator_abstract/README.rst
+++ b/barcodes_generator_abstract/README.rst
@@ -6,34 +6,35 @@
 Generate Barcodes for any Models (Abstract)
 ===========================================
 
-This module expends Odoo functionality, allowing user to generate barcode
+This module extends Odoo functionality, allowing users to generate barcodes
 depending on a given barcode rule for any Model.
 
 For example, a typical pattern for products is  "20.....{NNNDD}" that means
 that:
-* the EAN13 code will begin by '20'
+
+* the EAN13 code will begin by *20*
 * followed by 5 digits (named Barcode Base in this module)
 * and after 5 others digits to define the variable price
-* a 13 digit control
+* a 13th control digit (checksum)
 
 With this module, it is possible to:
 
 * Affect a pattern (barcode.rule) to a model
 
-* Define a Barcode base: 
-    * manually, if the base of the barcode must be set by a user. (typically an
-      internal code defined in your company)
+* Define a Barcode base:
+    * manually, if the base of the barcode must be set by a user (typically an
+      internal code defined in your company).
     * automaticaly by a sequence, if you want to let Odoo to increment a
-      sequence. (typical case of a customer number incrementation)
+      sequence (typical case of a customer number incrementation).
 
 * Generate a barcode, based on the defined pattern and the barcode base
 
 Installation
 ============
 
-This module use an extra python library named 'viivakoodi' you should install
-to make barcode generation works properly. 'viivakoodi' is a more active for of
-'pyBarcode'.
+This module use an extra python library named *viivakoodi* you should install
+to make barcode generation works properly. *viivakoodi* is a more active fork of
+*pyBarcode*.
 
 ``sudo pip install viivakoodi``
 
@@ -69,12 +70,16 @@ Usage
 
 This module is an abstract module. You can configure Barcode Rule, but to
 enable this feature, you need to install an extra module for a given model.
-This repository provide 'barcodes_generator_product' and
-'barcodes_generator_partner' module to generate barcode for product or partner
-model.
+This repository provides several such modules:
+
+* *barcodes_generator_product* for Products,
+* *barcodes_generator_partner* for Partners,
+* *barcodes_generator_location* for Stock Locations,
+* *barcodes_generator_lot* for Production Lots/Serial Numbers,
+* *barcodes_generator_package* for Packages.
 
 Alternatively, you can develop a custom module for a custom model. See
-'Inheritance' parts.
+*Inheritance* parts.
 
 Try this module on Runbot
 
@@ -86,7 +91,7 @@ Inheritance
 ===========
 
 If you want to generate barcode for another model, you can create a custom
-module that inherits on 'barcodes_generator_abstract' and inherit your model
+module that inherits *barcodes_generator_abstract* and inherit your model
 like that:
 
 class MyModel(models.Model):
@@ -98,7 +103,7 @@ class barcode_rule(models.Model):
 
     generate_model = fields.Selection(selection_add=[('my.model', 'My Model')])
 
-Finally, you should inherit your model view adding buttons and fields.
+Finally, you should inherit your model view to add buttons and fields.
 
 Note
 ----
--- a/barcodes_generator_abstract/README.rst
+++ b/barcodes_generator_abstract/README.rst
@@ -29,6 +29,8 @@ With this module, it is possible to:
 
 * Generate a barcode, based on the defined pattern and the barcode base
 
+* Generate an image of the barcode via a wizard
+
 Installation
 ============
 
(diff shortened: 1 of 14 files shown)


barcodes_generator_product/views/view_product_product.xml
```diff
@@ -21,21 +21,23 @@ License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
...
                     groups="barcodes_generator_abstract.generate_barcode" colspan="2"/>
-                <button name="generate_base" type="object" string="Generate Base (Using Sequence)" attrs="{'invisible': ['|',
-                    ('generate_type', '!=', 'sequence'),
-                    ('barcode_base', '!=', 0)]}"
+                <button name="generate_base" type="object"
+                    string="Generate Barcode Base"
+                    help="Generate Barcode Base from the sequence configured on the Barcode Rule"
+                    attrs="{'invisible': ['|',
+                        ('generate_type', '!=', 'sequence'),
+                        ('barcode_base', '!=', 0)]}"
```
legalsylvain, line 29: LGTM.

barcodes_generator_product/views/view_product_product.xml
```diff
@@ -21,21 +21,23 @@ License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
...
-                <button name="generate_barcode" type="object" string="Generate Barcode (Using Barcode Rule)"
-                    attrs="{'invisible': ['|',
+                <button name="generate_barcode" type="object"
+                    string="Generate Barcode"
+                    help="Generate a barcode using the selected Barcode Rule and the value of the Barcode Base"
+                    attrs="{'invisible': ['|', '|',
                             ('barcode_rule_id', '=', False),
-                            ('barcode_base', '=', 0)]}"
+                            ('barcode_base', '=', 0),
+                            ('barcode', '!=', False)]}"
```
legalsylvain, line 43: Not convinced for this point.  (just the ('barcode', '!=', False) condition.

This will prevent users to regenerate barcode, that is usefull  if you change the pattern of your barcode, changing prefix from 021... to 022.... for exemple.

barcodes_generator_product/views/view_product_template.xml
```diff
@@ -21,21 +21,23 @@ License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
...
-                <button name="generate_barcode" type="object" string="Generate Barcode (Using Barcode Rule)"
-                    attrs="{'invisible': ['|',
+                <button name="generate_barcode" type="object"
+                    string="Generate Barcode"
+                    help="Generate a barcode using the selected Barcode Rule and the value of the Barcode Base"
+                    attrs="{'invisible': ['|', '|',
                             ('barcode_rule_id', '=', False),
-                            ('barcode_base', '=', 0)]}"
+                            ('barcode_base', '=', 0),
+                            ('barcode', '!=', False)]}"
```
legalsylvain, line 42: same here.

legalsylvain (changes requested):
one question the rest LGTM.

thanks for improving this module.

barcodes_generator_product/views/view_product_product.xml
alexis-via: Right, I didn't have this scenario in mind (not sure if changing barcodes on existing product is a scenario that is supposed to happen, but...). In this case, they would have to empty the barcode field and then the button will be displayed again.

By the way, I forgot to say in the comments that I also removed the readonly attribute on the barcode field when generated from sequence so that users can remove a barcode from a product (this happens when they re-use an existing barcode for another product)

alexis-via:
@legalsylvain In the README.rst of barcodes_generator_abstract, in the installation section, there is:

```
'viivakoodi' is a more active for of 'pyBarcode'.
```

Did you mean "a more active fork" ? If yes, i'll fix this in this PR.

legalsylvain:
> Did you mean "a more active fork" ? If yes, i'll fix this in this PR.

indeed, thanks !

barcodes_generator_product/views/view_product_product.xml
legalsylvain: OK. Well, the problem I see in this configuration is the possiblity to write a bad barcode associated to a product with barcode_rule_id defined.

A medium proposal could be : 
- hide button generate Barcode if barcode is defined. (what you already did)
- let barcode readonly (to secure data). (by the way, should be readonly if barcode_rule_id != False).
- Add an extra button "Clear barcode" that set barcode to False, displayed if barcode is defined and barcode_rule_id is defined.

what do you think about that proposal ? More secure, and allow to change barcodes.

legalsylvain:
barcode image generation is a good improvment. Thanks.

barcodes_generator_abstract/README.rst
```diff
@@ -6,34 +6,37 @@
...
 
+* Generate an image of the barcode via a wizard
+
 Installation
 ============
 
-This module use an extra python library named 'viivakoodi' you should install
-to make barcode generation works properly. 'viivakoodi' is a more active for of
-'pyBarcode'.
+This module use an extra python library named *viivakoodi* you should install
```
rousseldenis, line 37: @alexis-via You should declare it in external_dependecies manifest part

barcodes_generator_abstract/wizard/barcode_image_generate.py
```diff
@@ -0,0 +1,123 @@
...
+    _description = 'Wizard to generate a barcode image'
+
+    @api.model
+    def _supported_barcode_formats(self):
+        res = []
+        for barcode_format in barcode.PROVIDED_BARCODES:
+            res.append((barcode_format, barcode_format))
+        return res
+
+    state = fields.Selection([
```
rousseldenis, line 28: Maybe a help part to describe briefly each option?

rousseldenis (changes requested):
Code review. Minor comments.
LGTM

barcodes_generator_abstract/README.rst
rousseldenis: Ok, did'nt see https://github.com/OCA/stock-logistics-barcode/pull/101/

rousseldenis:
@legalsylvain Can you review last changes?

legalsylvain:
@rousseldenis,
https://github.com/OCA/stock-logistics-barcode/pull/104#discussion_r136033225 is not answered. Or did I missed something ? 

regards.

barcodes_generator_product/views/view_product_product.xml
alexis-via: I find it a bit heavy... but I can understand that when you have "untrusted" users for which you want to lock down things and block mistakes as much as possible. So I just implemented it your way.

legalsylvain: Thanks a lot. :+1:

legalsylvain (approved):
Thanks for all the improvment.

code review. no test.

barcodes_generator_abstract/wizard/barcode_image_generate.py
```diff
@@ -0,0 +1,123 @@
...
+        default=2.0,
+        string='Space Between Barcode and Text',
+        help="Size of the white space between the bottom of the barcode and "
+             "the text of the barcode")
+    module_height = fields.Float(
+        default=15, string='Barcode Height',
+        help="Vertical size of the barcode")
+    module_width = fields.Float(
+        default=0.2, string='Barcode Width',
+        help="It doesn't seem to have any effect in EAN13...")
```
legalsylvain, line 56: Sorry, I'm rechecking this PR. I don't understand this helper. do you mean this field is useless : 
- allways ? 
- for ean13 ? 

in all cases, i don't think that this helper will help end user. Maybe a comment should be more usefull. (and maybe make this field hidden, if it is useless. don't you think ?

PR #106, opened 2017-09-08
[IMP] barcodes_generator_product: OpenUpgrade migration script

pedrobaeza:
Script for migrations coming from v8 module product_barcode_generator

cc @Tecnativa

### DIFF
--- a/barcodes_generator_product/migrations/9.0.1.0.0/post-migration.py
+++ b/barcodes_generator_product/migrations/9.0.1.0.0/post-migration.py
@@ -1,0 +1,39 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Tecnativa - Pedro M. Baeza
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from openupgradelib import openupgrade
+
+
+def convert_barcode_sequences(env):
+    """This converts the sequences found with a barcode to barcode rules for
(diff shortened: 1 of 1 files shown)


barcodes_generator_product/migrations/9.0.1.0.0/post-migration.py
```diff
@@ -0,0 +1,39 @@
...
+    """This converts the sequences found with a barcode to barcode rules for
+    being selectable at product level."""
+    if not openupgrade.column_exists(
+            env.cr, 'ir_sequence', 'barcode_sequence'
+    ):
+        # Return if product_barcode_generator was not previously installed
+        return
+    rule_obj = env['barcode.rule']
+    env.cr.execute("""
+        SELECT id, name, barcode_sequence
```
chienandalu, line 18: Shouldn't `number_next_actual` be fetched?

pedrobaeza: No, this relies the same on the sequence, so no need of that.

chienandalu: Yes, I see.

chienandalu (approved):

luismontalba (approved):

pedrobaeza:
OK, merging then.

PR #108, opened 2017-10-03
[IMP] stock_scanner: Add unit tests

Garamotte:
While writing the missing unit tests before migrating the module to v11, I found some minor bugs, so I also fixed them.
I could have refactored some parts, but this was not my goal, so I didn't to keep the diff as small and readable as possible.

Notes about the modifications:
- Bad indent fix (the loading script only loaded a single user per scenario): https://github.com/OCA/stock-logistics-barcode/compare/10.0...syleam:10.0-stock_scanner-add-unit-tests?expand=1#diff-9bbbfdfa76e6771500e6a08b417ba58bL86
- Don't crash if the supplied `xml_id` for a parent scenario doesn't exist: https://github.com/OCA/stock-logistics-barcode/compare/10.0...syleam:10.0-stock_scanner-add-unit-tests?expand=1#diff-9bbbfdfa76e6771500e6a08b417ba58bL121
- Impossible value (if the `message` argument is an integer, the previous search will crash, so we can't access this line with an integer): https://github.com/OCA/stock-logistics-barcode/compare/10.0...syleam:10.0-stock_scanner-add-unit-tests?expand=1#diff-9bbbfdfa76e6771500e6a08b417ba58bL121
- Removed the `current_object` argument everywhere: Looking carefully at the code, this argument could only contain the value of the `reference_document` field, and is only used to... write its value in `reference_document`, which is totally useless. This seems to be a remaining of some old code which was refactored/deleted.
- There is always at least the first step in the history, so it's never empty: https://github.com/OCA/stock-logistics-barcode/compare/10.0...syleam:10.0-stock_scanner-add-unit-tests?expand=1#diff-653282914028649fb94eeaf68edf9f64L434

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,9 +21,8 @@ env:
   matrix:
   - LINT_CHECK="1"
   - TRANSIFEX="1"
-  # 3 error log lines are generated during tests of the stock_scanner module, this is expected
-  - TESTS="1" SERVER_EXPECTED_ERRORS="3" ODOO_REPO="odoo/odoo"
-  - TESTS="1" SERVER_EXPECTED_ERRORS="3" ODOO_REPO="OCA/OCB"
+  - TESTS="1" ODOO_REPO="odoo/odoo"
+  - TESTS="1" ODOO_REPO="OCA/OCB"
 
 virtualenv:
   system_site_packages: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,8 +21,9 @@ env:
   matrix:
   - LINT_CHECK="1"
   - TRANSIFEX="1"
-  - TESTS="1" ODOO_REPO="odoo/odoo"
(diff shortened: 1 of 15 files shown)


Garamotte:
Ready to review, I based my WIP v11.0 branch for the `stock_scanner` module on this one to benefit from the new tests during migration.

cc @lmignon @damdam-s @pedrobaeza

pedrobaeza:
You can add `mute_logger` decorator instead for avoiding to add expected errors. Check OCA/social for some examples.

Garamotte:
`mute_logger` added.
I also moved import of the wrong scenario files from manifest to python test files.

lmignon (approved):
Thank you @sylvain-garancher What a great job!

Code review only

sla-subteno-it:
👍

pedrobaeza:
Merging this for having it for v11. Excellent work, Sylvain!

PR #115, opened 2017-11-21
[FIX] Handle KEY_RESIZE event in a selection action

benwillig:
Using ```sentinel```, there is an error when resizing the terminal. During resizing, KEY_RESIZE is sent but it's not handled in ```sentinel.py```. This errors occurs when using a scanner which has no physical keyboard (KEY_RESIZE is sent when the virtual keyboard is shown) but it can be reproduced with a standard terminal by connecting with the sentinel user.

This PR only fixes the error on the menu list screen. For other screens, the process will try to display chars outside the screen because the KEY_RESIZE event forced the terminal size. Then, following error is raised :  ```_curses.error: addstr() returned ERR``` and after that, current screen and related actions are blocked and user must kill & reopen the SSH connection in order to work again.

At the moment, I didn't find any solution to fix all screens without adding other errors. @sylvain-garancher Do you have any ideas?

### DIFF
--- a/stock_scanner/hardware/sentinel.py
+++ b/stock_scanner/hardware/sentinel.py
@@ -122,6 +122,11 @@ class Sentinel(object):
 
         # Initialize window
         self.screen = stdscr
+        self.auto_resize = False
+        self.window_height = 18
+        self.window_width = 6
+        # Store the initial screen size before resizing it
+        initial_screen_size = self.screen.getmaxyx()
         self._set_screen_size()
 
         self._init_colors()
@@ -139,10 +144,8 @@ class Sentinel(object):
                 _('Autoconfiguration failed !\nPlease enter terminal code'))
             self.scanner_check()
 
-        # Resize window to terminal screen size
-        self._resize()
-
         # Reinit colors with values configured in OpenERP
+        self._resize(initial_screen_size)
         self._reinit_colors()
 
         # Initialize mouse events capture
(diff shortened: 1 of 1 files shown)


Garamotte:
Thanks for the bug report and for starting the fix.

To manage text/integer/float inputs and confirm steps, I just added the `if key == 'KEY_RESIZE'` and it works.
For the simple message display steps, I had to create a new method, which only checks the `KEY_RESIZE` key to avoid duplicating code, because the `_display` method initializes many things before going to `getkey`.
Here is my patch: https://gist.github.com/sylvain-garancher/75a57329cce966f40a4d301b3cee8c60
If I'm not mistaken, all step types should now work.

lasley:
@sylvain-garancher - were you going to submit a PR to @benwillig branch with this patch?

Garamotte:
@lasley Good idea.
I never think about submitting PRs to PRs branches... :)

Done: https://github.com/acsone/stock-logistics-barcode/pull/1

Garamotte:
With this change, I think we can remove the screen width/height fields from the `scanner.hardware` model, and let the sentinel client automatically detect the screen's size during initialization.

Thoughts ?
[Edit: Maybe remove only for v11 migration to keep the v10 behaviour unchanged]

benwillig:
Thanks for the work! I tried so hard to fix this issue, I never thought the solution would be that simple. I will test it ASAP. 

Indeed I think we would remove width/height options. But then, we should trigger the KEY_RESIZE event at the beginning of the sentinel script.

Garamotte:
@benwillig I don't know if this PR is still useful, since I extracted the `sentinel` client in a python package, including this fix.

Maybe we can update the README instead ?
If you continue this PR, please see the latest version of the fix here: https://github.com/OCA/odoo-sentinel/pull/3

benwillig:
I tested the fix for several days and it works like expected. I already works with a sentinel implementation with the sentinel script inside the stock scanner module so this PR is still usefull for me. But indeed, it's a better idea to extract the package.

Garamotte:
It seems you forgot my latest fixes about terminals with a fixed-size virtual screen (at least Motorola MC3xxx running Windows). See https://github.com/OCA/odoo-sentinel/pull/3

Also, please rebase my fixup commit into yours :)

benwillig:
Done @sylvain-garancher. A huge thank you for your work.

Garamotte (approved):

PR #117, opened 2017-12-04
[IMP] stock_scanner: Add automatic json serialization for temporary values

Garamotte:
Based on an idea from @gurneyalex :)

This allows to access the temporary values by using `json_tmp_valX`, which will automatically (de)serialize the data. This is useful to store complex data structures in the temporary values, without having to manually do the serialization in the scenario's code.

cc @lmignon

### DIFF
--- a/stock_scanner/models/scanner_hardware.py
+++ b/stock_scanner/models/scanner_hardware.py
@@ -2,6 +2,7 @@
 # © 2011 Sylvain Garancher <sylvain.garancher@syleam.fr>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
+import json
 import logging
 import random
 import time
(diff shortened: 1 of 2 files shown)


sla-subteno-it (approved):

sa3m (approved):

lasley (approved):
I like; thanks @sylvain-garancher

benwillig:
With this PR, there is a problem when ```stock_scanner``` module is installed among ```queue_job``` module. [This line](https://github.com/OCA/queue/blob/10.0/queue_job/models/base.py#L20) inspects the model and tries to get the value of each property of the class. Because of the new methods, a singleton error is raised when the server is starting.

I don't know if ```queue_job``` module should be fixed rather than ```stock_scanner```. But if we fix ```stock_scanner```, we could change theses properties to compute & inverse methods. And maybe we could change the field type of the temporary values to [```Serialized```](https://github.com/odoo/odoo/blob/10.0/odoo/fields.py#L2445).

What do you think @sylvain-garancher @lasley ?

Garamotte:
When I added these properties, I first tried to create computed fields, but I didn't achieve to get it to work (I don't remember exactly what were the errors).

benwillig:
And what about if we add new Serialized fields instead of creating json computed fields based on temp values ?

PR #118, opened 2017-12-06
[8.0][PORT] mobile app inventory

legalsylvain:
Work make by @hparfr and me. Still work in progress for the time being.
supersed https://github.com/OCA/stock-logistics-barcode/pull/114


legalsylvain:
Hi all. Is this PR still WIP ?
regards & happy new years !

legalsylvain (commented):
Hi @hparfr. 
I just tested your module. I observe 2 regressions.
- if an error occurs, when adding a product, there is no message displayed, and it let the app in an inconsistent state.
- the focus on the scan field is regurlarly lost when the user add a new product, forcing user to click again on the field.

did you faced the same problem ? 
tested 
- on Mobile env (Firefox 57)
- on destop (firefox 57)

CC : @alexis-via 

regards.

hparfr:
>if an error occurs, when adding a product, there is no message displayed, and it let the app in an inconsistent state.

What kind of error ? Is it different from v7 ?

> the focus on the scan field is regurlarly lost when the user add a new product

Can you be more specific ?

legalsylvain:
>> if an error occurs, when adding a product, there is no message displayed, and it let the app in an inconsistent state.

> What kind of error ?

Some server error. A typical exemple you can reproduce. 
- create an inventory 1 (stock location), and add the demo product. Do not confirm.
- create an inventory 2 (same stock location) and add the same product. -> this will fail silently.

Investigating it seems that the error is not handled in the JS script.
image


>  Is it different from v7 ?

yes, in V7, there was a message if a call to the server was failing. "Something wrong occured, please refresh your app".

>> the focus on the scan field is regurlarly lost when the user add a new product

> Can you be more specific ?

Create an inventory / select a location / scan "5400313040109" / type 12 (for quantity) / click on validate. the focus is lost and doesn't allow to scan again a new barcode.

More generaly, as said previously, I'm glad to see all the good improvments added to the PR during the port of the module from V7. (JS improvment, API securization, etc...) and thank you for your insights.
But in the other hand, even if I understand your point of view, I dislike the new design of the UI with the "all in one" page when user can in a single input : 1) scan a stock location, 2) scan a product or 3) set a quantity. this new design loose the "automate" design of the current OCA module. (= in a state, we can only have an action allowed). We talked a lot about that and I'm convinced that we won't be agree on that point, that is not a big deal finally. I so propose you that I make a work to reintroduce the "automate" system, keeping your "all in one" design and to have the possibility to choose (by a setting) wich behaviour we want.
I know that it will duplicate some part of the code, but it will preserve the current behaviour of the module, and introduce the behaviour you want. 
Could you give me a feedback about that proposal ? 

Kind regards.

legalsylvain:
FYI, I talked today with @hparfr. We are agree to have this module having two bevahiours. (all-in-one and automate). 
I'll work on it on February to finish this PR.
kind regards.

legalsylvain:
Refactoring finished. Ready for review.

gurneyalex (approved):

hparfr (approved):

hparfr:
ping @legalsylvain

legalsylvain:
Hi @hparfr. i have my first inventory in V8 this end of week so, i'll test this module in production. I propose to test the last changes you introduced today and give you a feed back monday. If all is OK, I propose to merge this PR.
OK with this planning ?
Thanks for all.

hparfr:
:+1:

legalsylvain:
Finaly merged !

Regards.

PR #121, opened 2018-01-04
[ADD][10.0] new module 'barcodes_search'

legalsylvain:
This PR add a new module barcodes_search that allow user to search any item in any model, by it barcodes.

supersed : https://github.com/OCA/stock-logistics-barcode/pull/82

after the remark here : https://github.com/OCA/stock-logistics-barcode/pull/82#issuecomment-354831785

This module add a new button in the main menu

barcode_button_menu

The pop up allow user to search any item by it barcode.

partner_search

If found the form view of the item is displayed.

In two cases (two items with the same barcode or extra data present in the barcode (like price or weight)) a list is displayed.

partner_user_search

price_product_search

The module provides a developpers functions : 

`result = self.env['barcode.search'].search_by_barcode('12345567890123')`

See manifest file for a full description.

#GRAPOCA


barcodes_search/README.rst
```diff
@@ -0,0 +1,143 @@
...
+=====
+
+* In the main menu click on the barcode new button
+
+.. image:: /barcodes_search/static/description/barcode_button_menu.png
+
+* In the pop up form, enter the barcode and click on the search button
+
+.. image:: /barcodes_search/static/description/partner_search.png
+   :width: 800 px
```
lasley: Add an alt please

barcodes_search/README.rst
```diff
@@ -0,0 +1,143 @@
...
+
+This module expends Odoo functionality, allowing user to search any item by
+it barcode.
+
+Usage
+=====
+
+* In the main menu click on the barcode new button
+
+.. image:: /barcodes_search/static/description/barcode_button_menu.png
```
lasley: Add an alt please

rgarnau (approved):
LGTM :+1:

barcodes_search/README.rst
```diff
@@ -0,0 +1,143 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+=================
+Search by Barcode
+=================
+
+This module expends Odoo functionality, allowing user to search any item by
+it barcode.
```
rousseldenis: its

rousseldenis (approved):
Minor comment.

Tested on runbot. Working as expected.

legalsylvain:
@rousseldenis @rgarnau : thanks for your review. (i fixed the typo)

@lasley, could you take a look on that PR. I'd like to have your :+1:, as this PR supersed fully your PR.

thanks !

barcodes_search/static/src/js/barcodes_search.js
```diff
@@ -0,0 +1,26 @@
...
+    // * Menu item appended in the systray part of the navbar
+    // On click : Display a barcode search form view
+    // */
+    var BarcodeSearchItem = Widget.extend({
+        template:'barcodes_search.BarcodeSearchItem',
+        events: {
+            "click": "on_click",
+        },
+
+        on_click: function (event) {
```
lasley: Not a blocker or anything - but generally you want to use lowerCamelCase in JS for things you would normally snake_case in Python. Odoo SA really failed hard at that, but they seem to be working towards it lately.

barcodes_search/tests/test_barcode_search.py
```diff
@@ -0,0 +1,69 @@
...
+            'barcode_nomenclature_id': self.ref(
+                'barcodes.default_barcode_nomenclature'),
+            'sequence': 1,
+            'type': 'price',
+            'encoding': 'ean13',
+            'pattern': '23.....{NNNDD}',
+        })
+
+    # Test Section
+    def test_01_search_single_result(self):
```
lasley: Any reason you're numbering the tests? These are fine to be executed out of order IMO.

On a general note - if the tests are not fine to be executed out of order, it's worth taking a step back to see if you're properly setting the tests up. Typically unit tests should be focused on one goal per method and entirely atomic from all other methods. I haven't really found good use case for breaking those rules, but I'm sure there are some.

barcodes_search/views/templates.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <data>
```
lasley: No need for data tag

barcodes_search/wizards/barcode_search.py
```diff
@@ -0,0 +1,121 @@
+# -*- coding: utf-8 -*-
+# Copyright (C) 2017-Today GRAP (http://www.grap.coop)
+# @author: Sylvain LE GAL (https://twitter.com/legalsylvain)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import api, fields, models
+
+
+class BarcodeSearch(models.TransientModel):
+    _name = 'barcode.search'
```
lasley: Please add a _description  tag

barcodes_search/wizards/barcode_search.py
```diff
@@ -0,0 +1,121 @@
...
+
+    barcode = fields.Char(
+        string='Barcode', required=True, help='Barcode to search for.')
```
lasley: Not a big deal, but field declarations are much easier to read and maintain if they're defined one arg per line.

barcodes_search/wizards/barcode_search.py
```diff
@@ -0,0 +1,121 @@
...
+
+    state = fields.Selection(
+        selection=[('new', 'New'), ('done', 'Done')], default='new')
+
+    line_ids = fields.One2many(
+        comodel_name='barcode.search.line', inverse_name='barcode_search_id',
```
lasley: Please define one arg per line as indicated in [idioms section for multiline lists](https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#idioms)

barcodes_search/wizards/barcode_search.py
```diff
@@ -0,0 +1,121 @@
...
+        readonly=True)
+
+    @api.multi
+    def button_search(self):
```
lasley: `action_search` aligns more with the naming that Odoo SA uses, and that this returns an action

barcodes_search/wizards/barcode_search.py
```diff
@@ -0,0 +1,121 @@
...
+            self.line_ids = line_vals
+
+            # Return Action
+            action = self.get_formview_action()
+            action['target'] = 'new'
+            return action
+
+    @api.model
+    def search_by_barcode(self, barcode):
+        """Return a list of tuple (field, record, {extra_data})
```
lasley: Hmmmm the tuple structure is really hard to follow IMO. Any reason you didn't use a frozendict or namedtuple?

As evidence of this, I had to jump down here while reading the L26 if statement in order to ascertain what that statement does.

barcodes_search/wizards/barcode_search.py
```diff
@@ -0,0 +1,121 @@
...
+        self.ensure_one()
+        res = self.search_by_barcode(self.barcode)
+        if len(res) == 1 and not res[0][2]:
+            # Display the result in the according form view
+            record = res[0][1]
+            return record.get_formview_action()
+        else:
+            # Update Wizard Values
+            line_vals = []
+            self.line_ids.unlink()
```
lasley: IMO it would be better to put a `(5, 0)` in `line_vals` declaration and drop this line to save a one-off db call.

From what I understand, the records on O2M are cleaned when the foreign key is a required field. In this instance regardless, the vacuum will hit them because Transient.

barcodes_search/wizards/barcode_search.py
```diff
@@ -0,0 +1,121 @@
...
+            for item in res:
+                line_vals.append((0, 0, {
+                    'field_id': item[0].id,
+                    'item_id': item[1].id,
+                    'item_name': item[1].name,
+                    'extra_data': item[2],
+                }))
+            self.state = 'done'
```
lasley: `self.write` would be better here. Pretty sure the Odoo Active Record pattern hits the db each assignment, so this and the below line are two db calls. That's obviously not the case when in cache like an onchange, but I don't think we're in cache here.

barcodes_search/wizards/barcode_search.py
```diff
@@ -0,0 +1,121 @@
...
+            return action
+
+    @api.model
+    def search_by_barcode(self, barcode):
+        """Return a list of tuple (field, record, {extra_data})
+        that matches with the given barcode
+        """
+        barcode_fields = self.get_barcode_fields()
+        res = self._search_by_barcode_barcode_fields(
+            barcode, barcode_fields)
```
lasley: Please add trailing comma and the ) on its own line.

barcodes_search/wizards/barcode_search.py
```diff
@@ -0,0 +1,121 @@
...
+        nomenclatures = nomenclature_obj.search([])
+        rule_types = self.get_model_by_rule_type()
+        for nomenclature in nomenclatures:
+            parsed_result = nomenclature.parse_barcode(barcode)
+            rule_type = parsed_result['type']
+            if rule_type in rule_types:
+                barcode_fields = self.get_barcode_fields(rule_types[rule_type])
+                base_code = parsed_result['base_code']
+                extra_results = self._search_by_barcode_barcode_fields(
+                    base_code, barcode_fields)
```
lasley: Please add trailing comma and the ) on its own line.

barcodes_search/wizards/barcode_search.py
```diff
@@ -0,0 +1,121 @@
...
+                for extra_result in extra_results:
+                    extra_result[2]['type'] = rule_type
+                    extra_result[2]['value'] = parsed_result['value']
+                res += extra_results
+        return res
+
+    @api.model
+    def get_barcode_fields(self, model_name=False):
+        """Return a recordset of fields that represent a barcode, in any model.
+        By default, it will return all fields named 'barcode'.
```
lasley: Please add a blank line between your docblock summary and the additional text. Take a look at [PEP-257](https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings) for more info

barcodes_search/wizards/barcode_search.py
```diff
@@ -0,0 +1,121 @@
...
+        Note : Overload that function in a custom module, if you define
+        a barcode field with a name different than 'barcode'"""
+        field_obj = self.env['ir.model.fields']
+        if model_name:
+            domain = [
+                ('name', '=', 'barcode'), ('model', '=', model_name)]
+        else:
+            domain = [
+                ('name', '=', 'barcode'), ('model', '!=', 'barcode.search')]
```
lasley: I don't think there is ever an instance in which we would actually want `barcode.search` model right? It's transient, plus there is no use case for a barcode on it. In this light, I feel the below would be a bit better:

```python
domain = [
    ('name', '=', 'barcode'), ('model', '!=', 'barcode.search'),
]
if model_name:
    domain.append(('model', '=', model_name))
```

lasley: Note it might even make sense to add the below at the top of this method, which would help developers if they're being stupid but compile away with a Python running under `PYTHONOPTIMIZE=TRUE`:

```python
assert model_name != 'barcode.search'
```

barcodes_search/wizards/barcode_search.py
```diff
@@ -0,0 +1,121 @@
...
+        return field_obj.search(domain)
+
+    @api.model
+    def get_model_by_rule_type(self):
+        """Return a dictionary of rule type: model
```
lasley: Please add a summary line for this method. Multiline docblocks should contain a single line summary, followed by the extended description if necessary. 

I'd make suggestions, but I'm honestly having a bit of trouble understanding what this method is for. Is this to allow you to know what model to search when using a nomenclature of type X (weight, price, ..)?

barcodes_search/wizards/barcode_search.py
```diff
@@ -0,0 +1,121 @@
...
+        'price' is define in Odoo / 'point_of_sale' module.
+        'price_to_weight' is defined in OCA / 'pos_price_to weight' module
+        """
+        return {
+            'weight': 'product.product',
+            'price': 'product.product',
+            'price_to_weight': 'product.product',
+        }
+
+    # Private Section
```
lasley: This comment doesn't add any value and should be removed IMO

barcodes_search/wizards/barcode_search.py
```diff
@@ -0,0 +1,121 @@
...
+    @api.model
+    def _search_by_barcode_barcode_fields(self, barcode, barcode_fields):
+        res = []
+        for barcode_field in barcode_fields:
+            model_obj = self.env.get(barcode_field.model, False)
+            if not type(model_obj) is bool:
```
lasley: This is not a good way to test typing in Python. Typically you want to use `isinstance(` in order to allow for inheritance. Was there a reason to not use that here?

lasley: Ahh looking at this more, you're using this to determine if there was a `barcode_field.model` in the env? If so, typically EAFP is considered the more Pythonic approach:

```python
try:
    model_obj = self.env[barcode_field.model]
except KeyError:
    _logger.debug(
        'The %s model was not found in the env.', barcode_field.model,
    )
    continue
items = model_obj.search([(barcode_field.name, '=', barcode)])
res += [
    (barcode_field, item, {}) for item in items
]
```
  
Note the logger, which is to follow "Errors should never pass silently, unless explicitly silenced" of [PEP-20](https://www.python.org/dev/peps/pep-0020/). IMO the model not existing is an error, but I could be misunderstanding that part.

barcodes_search/wizards/barcode_search_line.py
```diff
@@ -0,0 +1,47 @@
...
+# Copyright (C) 2017-Today GRAP (http://www.grap.coop)
+# @author: Sylvain LE GAL (https://twitter.com/legalsylvain)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+import ast
+from odoo import _, api, fields, models
+
+
+class BarcodeSearchLine(models.TransientModel):
+    _name = 'barcode.search.line'
```
lasley: `_description`

barcodes_search/wizards/barcode_search_line.py
```diff
@@ -0,0 +1,47 @@
...
+
+    barcode_search_id = fields.Many2one(
+        comodel_name='barcode.search')
+
```
lasley, line 15: Why the extra spacing between the field declarations?

barcodes_search/wizards/barcode_search_line.py
```diff
@@ -0,0 +1,47 @@
...
+    extra_data = fields.Char()
+
+    extra_data_display = fields.Text(
+        string='Extra Data', compute='_compute_extra_data_display')
+
+    @api.depends('extra_data')
+    def _compute_extra_data_display(self):
+        for line in self:
+            res = []
+            for k, v in ast.literal_eval(line.extra_data).iteritems():
```
lasley: You should probably just use `items` unless there's more than a few hundred elements. This makes the code directly Python3 compatible (assuming a dict is returned from `literal_eval` anyways)

barcodes_search/wizards/barcode_search_line.py
```diff
@@ -0,0 +1,47 @@
...
+                if k == 'type':
+                    res.append('- %s : %s' % (_('Type'), v))
```
lasley: Is there a reason for the space between the name and the semicolon? Typically you want to put the semicolon directly against the name, as you would in a dict declaration:

```
'- %s: %s' % (_('Type'), v)
```

Which would result in:

```
- Type: Something
```

barcodes_search/wizards/barcode_search_line.py
```diff
@@ -0,0 +1,47 @@
...
+    barcode_search_id = fields.Many2one(
+        comodel_name='barcode.search')
+
+    model_id = fields.Many2one(
+        comodel_name='ir.model', readonly=True, related='field_id.model_id')
+
+    field_id = fields.Many2one(
+        comodel_name='ir.model.fields', readonly=True, string='Field')
+
+    item_id = fields.Integer(string='Item ID', readonly=True)
```
lasley: Can we name this differently? I had to jump back up because I thought it was an actual ID column. Maybe `item_identifier`? or `item_ref`?

barcodes_search/wizards/barcode_search_line.py
```diff
@@ -0,0 +1,47 @@
...
+                    res.append('- %s : %s' % (_('Type'), v))
+                elif k == 'value':
+                    res.append('- %s : %s' % (_('Value'), v))
+            line.extra_data_display = '\n'.join(res)
+
+    @api.multi
+    def button_view(self):
+        self.ensure_one()
+        model_obj = self.env[self.model_id.model]
+        record = model_obj.browse(self.item_id)
```
lasley: Is there a reason to not use `fields.Reference(` instead of the `model_id` and `item_id` combo? IMO it's much easier to wield than manually converting all the time, and comes with a nifty record selector in the front end.

barcodes_search/wizards/barcode_search_view.xml
```diff
@@ -0,0 +1,46 @@
...
+<odoo>
+
+    <record id="view_barcode_search_form" model="ir.ui.view">
+        <field name="model">barcode.search</field>
+        <field name="arch" type="xml">
+            <form>
+                <group>
+                    <field name="state" invisible="1"/>
+                    <field name="barcode" />
+
```
lasley: No need for the extra space IMO

lasley (changes requested):
Thanks @legalsylvain - comments inline 🚀

barcodes_search/wizards/barcode_search.py
legalsylvain: >  Is this to allow you to know what model to search when using a nomenclature of type X (weight, price, ..)?

Yes. 
- I want to search only on certain nomenclature.
- I want to make this list overloadable.

barcodes_search/wizards/barcode_search.py
legalsylvain: It's not an error. It can occure "normaly" especially in tests cases, if you have other module installed because when the test is executed, only a part of model are loaded in env, but all are in database. So I don't think a KeyError is necessary.

lasley: Thanks for the explanation @legalsylvain. Just the `_logger` was the part indicating the error though, so that would just be omitted if it's a normal case. The `KeyError` is just a Python idiom for [EAFP](https://docs.python.org/2/glossary.html#term-eafp). Assuming that a key exists and reacting if it doesn't is way cheaper in computational complexity than checking for it to exist every time.

barcodes_search/wizards/barcode_search.py
lasley: Thanks for the elaboration. Does the following hit the intent?

```
Provides a mapping for nomenclature types to Odoo models.

The dictionary returned by this method maps barcode nomenclature rule types (``barcode.rule.type``) to the models that they represent. A barcode nomenclature rule type allows for numerical content to be represented in a barcode by utilizing the `......{NNDDD}` format to indicate the fixed and variable parts of the barcode respectively. 

Note: Modules adding new barcode nomenclature rule types will need to override this method to update the mapping in order to be supported.

For example:
* `weight` is defined in the Odoo `stock` module
* `price` is defined in the Odoo `point_of_sale ` module
* `price_to_weight ` is defined in the OCA `pos_price_to ` module

Returns:
    dict: A mapping of models keyed by their respective barcode rule type (``barcode.rule.type``).
```

That said, aren't we missing a bunch of rule types here?

barcodes_search/wizards/barcode_search.py
mourad-ehm: I think that  "string='Barcode'" is not necessary in this case. It is generated by odoo. this allows to reduce the code

barcodes_search/wizards/barcode_search_line.py
gurneyalex: that's a French idiosyncrasy (French typography mandates a non breaking space before colon and semicolon, and hitting the space bar before these characters it is kind of hard wired in French ppl...)

barcodes_search/wizards/barcode_search_line.py
legalsylvain: I don't get it. Feel free to make a PR to update the code, if you want.

legalsylvain:
Hi @lasley : I took into account your remarks. Could you update your review ? 

@mourad-ehm : Is it now OK for you ? 

Thanks !

legalsylvain:
Hi. @oca/logistics-maintainers, could you merge this approved PR ?

rousseldenis:
@legalsylvain  Ok, is this ready ? 

FYI, @lasley is no more active on OCA.

legalsylvain:
@rousseldenis : yes it is !

rafaelbn (approved):
:+1:  LGTM

Issue #126, opened 2018-01-23
Odoo server can't start when stock_scanner is installed with queue_job.


benwillig:
There is a problem when ```stock_scanner``` module is installed among ```queue_job``` module. [This line](https://github.com/OCA/queue/blob/10.0/queue_job/models/base.py#L20) inspects the model and tries to get the value of each property of the class. Because of the new methods, a singleton error is raised when the server is starting.

I don't know if ```queue_job``` module should be fixed rather than ```stock_scanner```.(https://github.com/odoo/odoo/blob/10.0/odoo/fields.py#L2445).

As @sylvain-garancher said in https://github.com/OCA/stock-logistics-barcode/pull/117, there is a problem with compute and inverse methods.

In my opinion, we should add new ```Serialized``` fields instead of trying to work with ```Char``` fields.

rousseldenis:
Closing this as fixed

PR #127, opened 2018-01-23
[CHG] Temporary values as JSON

benwillig:
Fixes https://github.com/OCA/stock-logistics-barcode/issues/126.

@sylvain-garancher With this way, we can add as much temporary values as we want.

### DIFF
--- a/stock_scanner/__manifest__.py
+++ b/stock_scanner/__manifest__.py
@@ -5,7 +5,7 @@
 {
     'name': 'Stock Scanner',
     'summary': 'Allows managing barcode readers with simple scenarios',
-    'version': '10.0.1.0.2',
+    'version': '10.0.1.0.3',
     'category': 'Generic Modules/Inventory Control',
     'website': 'https://odoo-community.org/',
     'author': 'SYLEAM,'
--- a/stock_scanner/models/scanner_hardware.py
+++ b/stock_scanner/models/scanner_hardware.py
@@ -172,6 +172,9 @@ class ScannerHardware(models.Model):
         required=True,
         default='red',
         help='Color for the error background.')
+    tmp_values = fields.Serialized(
+        readonly=True
+    )
 
     @property
     @api.multi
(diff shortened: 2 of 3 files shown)


rousseldenis (approved):
LGTM.

@benwillig Just a test for empty_scanner_values() ?

benwillig:
@rousseldenis Done

rousseldenis:
:+1:

Garamotte:
@benwillig This means that any module that defines properties cannot be installed with `queue_job` ?
In this case, it's an issue of the `queue_job` module, IMO.

Also, removing the properties will make our scenarios fail when we'll update the module... :(

Anyway, you should split the tests, at least one per method, instead of a big single method, to be able to see which test fails.

benwillig:
Having properties on a model is not the problem. The error which is raised on startup (or during test) is caused by the ```self.ensure_one()``` defined at the beginning of the method. This is why I'm not sure which module we should fix, as I already said.

I can write a migration script which stores the value of tmp_val<nr> (if it's a valid json) in ```tmp_values['json_tmp_val<nr>']``` or ```tmp_values['tmp_val<nr>']```. But I can't create the property ```json_tmp_val<nr>``` because of the ```ensure_one```.

Garamotte:
I understand the issue, but removing the properties will need to adapt scenarios with a simple update of the module.

I made some tests, and I can confirm that any module with properties will crash if they contain `self.ensure_one()`. And ensuring we have a single record is better when we call a property. This is due to Odoo's models desin (because self can be nothing, a single record, or a collection of records).

Even if it's a bit ugly, what di you think about adding a warning in the properties, and remove them for the next version ?
```python
    @property
    @api.multi
    def json_tmp_val1(self):
        if not self:
            logger.warning('DEPRECATED: This attribute will be removed in the next version.')
            return

        self.ensure_one()
        return json.loads(self.tmp_val1 or 'null')
```

IMO, the best solution would be to find a way to list methods without calling the properties in `queue_job`, to avoid later conflicts with any module that declares properties, but I don't know the `inspect` module enough to change that.

benwillig:
I thought about this solution but as you said, it's a bit ugly. But I think it's the only workaround at the moment. I will change this PR to readd thoses properties with the warning.

Garamotte:
@benwillig I finally found how to fix `queue_job`:
```diff
diff --git queue_job/models/base.py queue_job/models/base.py
index 5ded5c66ebf6..4ab252281fc5 100644
--- queue_job/models/base.py
+++ queue_job/models/base.py
@@ -23,7 +23,7 @@ def _register_hook(self):
             if attr_name == '_cache':
                 continue
             try:
-                attr = getattr(self, attr_name)
+                attr = getattr(self.__class__, attr_name)
             except AttributeError:
                 continue
             if inspect.ismethod(attr) and getattr(attr, 'delayable', None):
```
The issue was caused by using a recordset (eg. an instance of the class) instead of the class itself.
By doing this, properties are evaluated, instead of being returned directly.

Uing a Serialized field is a great idea, and I'll hapilly approve the PR, but this is not related to the bug anymore :)

benwillig:
Sorry I didn't have time to readd properties.  The code you gave comes from v11, so we need to backport thoses commits to 10.0, then I will be able to add your properties without the warning.

Garamotte:
I'm currently creating PRs on `queue_job` for both versions ;)

Edit: https://github.com/OCA/queue/pull/49 (10.0) and https://github.com/OCA/queue/pull/50 (11.0)

benwillig:
Updated my PR. Thanks again @sylvain-garancher

Garamotte:
Thanks.
If I'm not wrong, there is no test for the `default` argument of `get_tmp_value`.

Also, could you split the test in several smaller tests ?
This allows to have a better overview of tests when there are fails.

benwillig:
Spllited and added a test.

Garamotte (approved):

rousseldenis (approved):
LGTM

LoisRForgeFlow (approved):
LGTM

PR #130, opened 2018-03-27
[ADD] Picking Lot barcode module

SimoRubi:
This module allows the user to scan a lot name when in the picking form; this behaviour was present in v8 but has been removed in v10.

Please check

### DIFF
--- a/picking_lot_barcode/README.rst
+++ b/picking_lot_barcode/README.rst
@@ -1,0 +1,47 @@
+.. image:: https://img.shields.io/badge/license-AGPL--3-blue.svg
+   :target: https://www.gnu.org/licenses/agpl
+   :alt: License: AGPL-3
+
+===================
+Picking Lot barcode
+===================
+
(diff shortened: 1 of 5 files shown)


SimoRubi:
Closing this because depends on an enterprise module

pedrobaeza:
I think this can be done without depending on the enteprise module.

SimoRubi:
@pedrobaeza do you mean depending on `stock` and `barcodes`? Then it would be some kind of a substitute for `stock_barcode` module.. Is this allowed?

pedrobaeza:
Yes, why not? The only rule is to not copy enterprise code, but you can put similar features in OCA that you have in Odoo enterprise. The motivation for that is to have an open source alternative and distribute costs another way: more cost at the beginning (due to development) vs a constant lower cost across time (due to enterprise fee).

There are already several examples of features in OCA that overlaps somehow enteprise modules: responsive interface, website form builder, bom versioning, quality control...

PR #132, opened 2018-04-17
[10.0][IMP] Identical name management in scenario export

ctr-subteno-it:
Added mangement for steps and transitions having identical names. Also fixed XML ID file naming.

Previously, if 2 steps or transitions had the same name, they would "share" the same XML ID so one would overwrite the other. This adds a "_2" at the end (and "_3", "_4", ...) if necessary.

Also, the filename generated contains the module name - if you generated the workflow from scratch, the first export generated `dir/step_1.py` but subsequent exports generate `dir/module.step_1.py`. Now, the module is ignored if it is the same as the scenario's module.

closes #128

PR #133, opened 2018-04-18
[11.0][IMP] stock_scanner: Ensure that transition names are unique within same scenario

ghost:
### Summary
When the scenario is exported, transition XMLID-s are being constructed using their `name` and `scenario_id.name` fields, so in case we're having multiple transitions w/ same name within the same scenario, only one of those is going to survive the export.

### DIFF
--- a/stock_scanner/README.rst
+++ b/stock_scanner/README.rst
@@ -220,6 +220,7 @@ Images
 Contributors
 ------------
 * Alexandre Fayolle <afayolle.ml@free.fr>
+* Artem Kostyuk <a.kostyuk@mobilunity.com>
 * Christophe CHAUVET <christophe.chauvet@syleam.fr>
 * Damien Crier <damien@crier.me>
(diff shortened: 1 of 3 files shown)


ctr-subteno-it:
Hi - If you look at PR #132, I manage this already. I've prepared a V11 version that allows multiple transitions and steps with the same name, but I haven't yet tested it. It's based on the v10 version so it works the same way - do you want me to do a PR despite not being able to test ?

ghost:
@ctribbeck yeah, just noticed that there is a PR solving this issue already :smile:
sure, go ahead, and I'll check that v11 runs properly

ghost:
I should admit that your approach is better than mine :wink:

ctr-subteno-it:
I've pushed it - thanks for testing! We have a v11 project but we're not yet ready to export the scenarios (and I'm not in that project). It's PR #133 (if you haven't seen it).

pedrobaeza:
Closing as not needed with #134

PR #135, opened 2018-05-26
[FIX]Improved record id.

MeetKD:


pedrobaeza:
Why this change?

MeetKD:
**view_production_lot_form** record id is used in stock module. To uniquely identify the record id I have changed the id and used **view_barcode_production_lot_form** to differentiate record id.

pedrobaeza:
On contrary, the style guide indicates to put the same ID, as the fully qualified XML-ID is module+ID, so no possible choke. Please check these guidelines: https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md

MeetKD:
Okay. Thank you for clarification.

PR #136, opened 2018-06-08
[FIX] do not copy barcode_base during duplication

legalsylvain:
Trivial PR. 
Just set copy=False on barcode_base field. This field is a kind of ID of the product to generate the barcode for product, partner, etc... 
By default it should not be copied during duplication process.

thanks for the review.


bealdav (approved):
Except small lints

pedrobaeza:
Please change module version number

legalsylvain:
@pedrobaeza. Done.

pedrobaeza (approved):
Fast-tracking it

PR #140, opened 2018-06-11
[11.0][MIG] product_multi_ean

xavierjimenez:
Migration of the module **product_multi_ean** to version 11.0.


product_multi_ean/models/product_product.py
```diff
@@ -1,33 +1,46 @@
...
     @api.onchange('name')
     def _check_name(self):
+        barcode_obj = self.env['barcode.nomenclature']
+
         for record in self:
-            if not addons_product.check_ean(record.name):
-                raise UserError(
-                    _('You provided an invalid "EAN13 Barcode" reference. You '
-                      'may use the "Internal Reference" field instead.'))
+            if record.name:
```
pedrobaeza: Add `.filtered('name')` instead

pedrobaeza (changes requested):
Don't change file permissions

xavierjimenez:
Done!

product_multi_ean/models/product_product.py
```diff
@@ -1,33 +1,46 @@
...
 # © 2015 Roberto Lizana (Trey)
 # © 2016 Pedro M. Baeza
+# © 2018 Xavier Jimenez (QubiQ)
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
-from openerp import models, fields, api, _
-from openerp.exceptions import Warning as UserError
-from openerp.addons.product import product as addons_product
+from odoo import models, fields, api, _
+from odoo.exceptions import Warning as UserError
```
pedrobaeza: Import directly `UserError`, as now the other is deprecated

product_multi_ean/models/product_product.py
```diff
@@ -1,33 +1,46 @@
...
     @api.onchange('name')
     def _check_name(self):
+        barcode_obj = self.env['barcode.nomenclature']
+
         for record in self:
-            if not addons_product.check_ean(record.name):
-                raise UserError(
-                    _('You provided an invalid "EAN13 Barcode" reference. You '
-                      'may use the "Internal Reference" field instead.'))
+            if record.filtered('name'):
```
pedrobaeza: This should be put in the previous line: `for record in self.filtered('name'):` The idea is to save one line and one cyclomatic complexity level

product_multi_ean/__manifest__.py
```diff
@@ -1,19 +1,23 @@
...
     'license': 'AGPL-3',
     'author': "Camptocamp, "
               "Trey, "
               "Tecnativa, "
+              "QubiQ, "
               "Odoo Community Association (OCA)",
     'category': 'Product Management',
-    'depends': ['product'],
+    'depends': [
+        'sale',
```
pedrobaeza: Instead of adding this dependency, put `compute_sudo=True` and `inverse_sudo=True` and everything will work correctly without adding the permission

pedrobaeza (changes requested):

xavierjimenez:
Changes done!

pedrobaeza:
Travis and runbot is now red

xavierjimenez:
@pedrobaeza it fails because it can't find the sales_team.group_sale_manager. 

On 8.0, that group was created on 'base' module so I think that the 'sale' module must be added as a dependency, is that all right?

pedrobaeza:
But we already talked yesterday of removing that ACL, isn't it?

HaraldPanten (approved):
It's working fine for me :+1:

pedrobaeza (approved):

PR #141, opened 2018-06-11
11.0 [MIG] barcodes_generator_package

misern2:
Standard migration


barcodes_generator_package/__manifest__.py
```diff
@@ -14,6 +13,8 @@
     "depends": [
         "barcodes_generator_abstract",
         "stock",
+        "delivery"
```
legalsylvain: Hi @misern2.
Thanks for porting this module.

I don't think adding a dependency to ```delivery``` module is a good idea. Generating barcode for package should be possible without delivery module installed.

@lasley : any PoV on that point ?

kind regards.

legalsylvain (changes requested):

barcodes_generator_package/__manifest__.py
misern2: Yes you are right.
There is something that I dont undertand about the stock module.
The packaging_id (package type) field used for filling the barcodes is only available in the list view.

The delivery module adds the field on the form view and that was the reason for adding the depedency.

Any clue how to fix it?

legalsylvain: Hi @misern2. thanks for your answer.

I don't understand why you need package_id. It is not required in the xml view I think.
If you use the original file barcodes_generator_package/views/product_packaging.xml, it should work. Don't you think ? Or did I miss something ?

regards.

barcodes_generator_package/models/product_packaging.py
```diff
@@ -0,0 +1,12 @@
...
+
+from odoo import fields, models
+
+
+class ProductPackaging(models.Model):
+    _name = 'product.packaging'
+    _description = 'Product Packaging'
+    _inherit = ['product.packaging', 'barcode.generate.mixin']
+
+    barcode = fields.Char()
```
rousseldenis, line 12: @misern2 Did you see my PR that reimplement former barcode field (product_packaging_barcode) ? #153

rousseldenis:
@misern2

PR #146, opened 2018-06-12
10.0 [MIG] stock_scanner_receipt

misern2:
Standard Migration.

PR #148, opened 2018-06-13
[FIX] Added parent menu Stock  for stock_scanner_XXX modules

misern2:
As commented in pr #146

### DIFF
--- a/stock_scanner/__manifest__.py
+++ b/stock_scanner/__manifest__.py
@@ -25,6 +25,7 @@
         'data/ir_cron.xml',
         'data/scenarios/Login/Login.scenario',
         'data/scenarios/Logout/Logout.scenario',
+        'data/scenarios/Stock/Stock.scenario',
         'wizard/stock_scanner_config_wizard_view.xml',
         'views/menu.xml',
         'views/scanner_scenario.xml',
--- a/stock_scanner/data/scenarios/Stock/Stock.scenario
+++ b/stock_scanner/data/scenarios/Stock/Stock.scenario
@@ -1,0 +1,9 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<scenario>
+  <id>stock_scanner.scanner_scenario_stock</id>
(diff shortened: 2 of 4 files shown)


misern2:
@pedrobaeza As soon as it is merged I'll update the rest of PR.

Thanks!

pedrobaeza (approved):
Waiting for CIs for fast-tracking this

Garamotte:
This will create a menu which will be always displayed. Not anyone will want this menu.

Moreover, the active flag is set in the data, and it is not set as noupdate.

pedrobaeza:
Yeah, but it's the only way for avoiding an error in runbot in the rest of the modules, and that menu is not a problem at all. You can't put it non active, or you will have the same problem reactivating it on children menus.

Garamotte:
Maybe we can mimic Odoo's behaviour: If a menu has no child, do not display it on the scanner.

The issue is that always displaying it will be boring for most users when there is no child in this menu. And I think it will crash if someone selects it.

pedrobaeza:
OK, I think that can be feasible. @misern2 do you think you can do it?

misern2:
@pedrobaeza @sylvain-garancher  Finally I think the best approach is to hide the menu option only in SSH client and leave it in  the backend (Odoo) as it is

pedrobaeza:
Do you know why Travis is red?

pedrobaeza:
```
stock_scanner/models/scanner_hardware.py:765:26: E231 missing whitespace after ','
stock_scanner/models/scanner_hardware.py:765:35: E231 missing whitespace after ','
stock_scanner/models/scanner_hardware.py:766:25: E231 missing whitespace after ','
stock_scanner/models/scanner_hardware.py:766:34: E231 missing whitespace after ','
stock_scanner/models/scanner_scenario.py:79:21: E251 unexpected spaces around keyword / parameter equals
```

pedrobaeza:
Merging this as well

PR #167, opened 2018-11-26
[10.0][FIX] stock_scanner: Improve login method

rousseldenis:


stock_scanner/models/scanner_hardware.py
```diff
@@ -451,6 +451,8 @@ def scanner_end(self, numterm=None):
 
     @api.one
     def check_credentials(self, login, password):
+        if password == '':
```
sbidoul:
```suggestion
        if not password:
```

jgrandguillaume (approved):
Hi,

Thx for the contrib. LGTM if you changed what's Stéphane pointed out.

Regards,

Joël

stock_scanner/models/scanner_hardware.py
rousseldenis: Ok! It was to highlight the use case but I change it

rousseldenis: Done!

benwillig (approved):

lmignon (approved):

dreispt (approved):

lmignon (approved):
Thank you @rousseldenis

lmignon:
@rousseldenis Nice your last test.. travis is :red_circle:  :smirk:

rousseldenis:
needs fixing. Want to add a test

rousseldenis:
Done !

pedrobaeza:
Should this be forward-ported to v11?

rousseldenis:
@pedrobaeza No. The problem is fixed in Odoo (I made the OCA PR as I didn't know how long that could take) since https://github.com/odoo/odoo/commit/6639630de11e5e3b4046968f8981415d6ccca6cd

PR #169, opened 2018-12-10
[FIX] stock_scanner: Pass tests with auth_crypt

yajo:
The value of the user's `password` field is encrypted if `auth_crypt` is installed, so that would make this test fail.

The password is always going to be "demo" anyways, so I just hardcode it here to make the test pass in integrated environments.

Notice that this is the common usage, and in fact this same file uses it in other places, both [to test it fails][1] and [to test it works][2].

[1]: https://github.com/OCA/stock-logistics-barcode/blob/d432b52538c9b9ddb35c05ade7bb9dec5ccc1c51/stock_scanner/tests/test_stock_scanner.py#L217
[2]: https://github.com/OCA/stock-logistics-barcode/blob/d432b52538c9b9ddb35c05ade7bb9dec5ccc1c51/stock_scanner/tests/test_stock_scanner.py#L272

@Tecnativa

### DIFF
--- a/stock_scanner/tests/test_stock_scanner.py
+++ b/stock_scanner/tests/test_stock_scanner.py
@@ -226,7 +226,7 @@ class TestStockScanner(common.TransactionCase):
             code, action='action', message=user_demo.login)
         # and the right pwd
         ret = scanner_hardware.scanner_call(
-            code, action='action', message=user_demo.password)
+            code, action='action', message="demo")
         # now we are logged in
         self.assertEqual(('F', [
             'You are now authenticated as demo !',


pedrobaeza (approved):

chienandalu (approved):

cubells (approved):
:+1: 

LGTM

pedrobaeza:
@rousseldenis can you confirm this isn't a problem?

rousseldenis:
@Yajo Was fixed in https://github.com/OCA/stock-logistics-barcode/pull/167 in addition of odoo fix https://github.com/odoo/odoo/commit/6639630de11e5e3b4046968f8981415d6ccca6cd

pedrobaeza:
@rousseldenis this is different. It's for making the test to work when `auth_crypt` is installed.

rousseldenis:
That's why I did the fix :smile: It's the same

rousseldenis:
@pedrobaeza FYI auth_crypt is installed by default. So, the test is ok after my PR was merged

pedrobaeza:
Uhm, so do you mean to make the same PR than in v10 for v11?

rousseldenis:
Oups, didn't see branch version. @Yajo could you please take my commit as it involve security problems ?

yajo:
You said in v11+ there's no such problem: https://github.com/OCA/stock-logistics-barcode/pull/167#issuecomment-443651113

Are you sure I have to cherry-pick it?

rousseldenis:
@Yajo Mmmh. I see. This is ok !

rousseldenis (approved):

PR #173, opened 2019-03-20
[MIG] stock-scanner: Migration to 12.0

fmdl:
[FIX] duplicate scenario


fmdl:
@sla-subteno-it

rousseldenis (approved):
Code review. Seems good. @fmdl Thanks for improvements!

tbaden:
thanks for the work 👍 
I did a quick check and the xml files should be refactored (see contribution guidelines)
https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst

fmdl:
@tbaden was is the exact issue with xml files ?

tbaden:
@fmdl 
id pattern

> Use the following patterns, where <model_name> is the name of the model that the menu, view, etc. belongs to (e.g. for a res.users form view, the name would be res_users_view_form):
> 
> For a menu: <model_name>_menu
> For a view: <model_name>_view_<view_type>, where view_type is kanban, form, tree, search, ...
> For an action: the main action respects <model_name>_action. Others are suffixed with _<detail>, where detail is an underscore lowercase string explaining the action (should not be long). This is used only if multiple actions are declared for the model.
> For a group: <model_name>_group_<group_name> where group_name is the name of the group, generally 'user', 'manager', ...
> For a rule: <model_name>_rule_<concerned_group> where concerned_group is the short name of the concerned group ('user' for the 'model_name_group_user', 'public' for public user, 'company' for multi-company rules, ...).


 and id should be the first attribute
> When declaring a record in XML:
> 
> Place id attribute before model

fmdl:
@pedrobaeza I need to update the xml file ?

pedrobaeza:
@fmdl @tbaden well, it's something desirable for being homogeneous, but not mandatory for blocking the PR, so if you want to do it, it would be welcome. What I see more advisable is to squash all your migration commits in 1, as they don't add any value to be split.

tbaden:
@pedrobaeza 
how should I handle these kind of changes in the future?
just suggesting these changes and approve anyway?
and which kind of changes described in the contribution guideline are really "pr block worthy"?

@fmdl will approve when squashed

tbaden (approved):
lgtm now

stock_scanner/__manifest__.py
```diff
@@ -0,0 +1,48 @@
+# © 2011-2015 Sylvain Garancher <sylvain.garancher@syleam.fr>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    'name': 'Stock Scanner',
+    'summary': 'Allows managing barcode readers with simple scenarios',
+    'version': '12.0.1.0.0',
+    'category': 'Generic Modules/Inventory Control',
+    'website': 'https://odoo-community.org/',
```
tbaden: website should be: https://github.com/OCA/stock-logistics-barcode

tbaden (changes requested):
noticed this just now

stock_scanner/__manifest__.py
fmdl: done

tbaden (approved):

PR #120, opened 2018-01-03
[10.0] MIG mobile_app_inventory

alexis-via:

PR #174, opened 2019-04-02
[10.0][MIG] mobile app inventory

abondis:
Based on #120 
Fixes:
- update inventory location when possible
- add a route to change location
- get locations returns all available locations when changing inventory location or creating a new inventory
￼- do not confirm inventory until a product is added
- set inventory in scan_state data when creating a new inventory


abondis:
I can't seem to see what's breaking the build, the error seems to come from another module that this PR did not touch (not that I can see at least)

rvalyi:
Hello @abondis
thanks for the work. However you should preserve the project commits history. So basically you should create a new branch from the #120 branch and add your extra commits after.

abondis:
@rvalyi thanks for the information. I re-imported the history from #120 from where we started.

PR #179, opened 2019-04-30
[12.0][ADD] stock_picking_barcode

AdriaGForgeFlow:
This module allows to use a barcode scanner to assign serial numbers to a product in a picking.

cc ~ @Eficent


stock_picking_barcode/readme/USAGE.rst
```diff
@@ -0,0 +1,25 @@
+To use this module you need a barcode scanner.
+
+In Odoo, click on the scan button in the picking form:
+
+.. image:: /stock_picking_barcode/static/description/stock_picking_scan.png
```
pedrobaeza: You have to use relative paths for proper conversion to absolute one: `./static/description/...

legalsylvain: In fact ../static/description

pedrobaeza: Really? Check here last image:

https://raw.githubusercontent.com/OCA/web/12.0/web_responsive/readme/DESCRIPTION.rst

and the result in resulting README and I see it correct:

https://raw.githubusercontent.com/OCA/web/12.0/web_responsive/README.rst

legalsylvain: Well, 

- if you set ../static/description it will work in the file "USAGE.rst" AND the main file "README.rst"
- if you set ./static/description it will work in the main file "README.rst" but the file usage.rst will not be well displayed if you try to see it via github.

Note : in your sample (https://github.com/OCA/web/blob/12.0/web_responsive/readme/DESCRIPTION.rst), all the image url are absolute, not relative, except the last one and you can see it doesn't work.

It's a minor problem because the most important is the main readme file, but writing ../ make the review easier during pull request because you can see all the description (text and images) in one time.

Regards.

pedrobaeza: Wow, that's great! Thanks for the insights. I didn't know about that. Definitively all my next relative image paths will be with `..`!

JordiBForgeFlow:
@ageficent I think that we can close this one, right? The `stock_barcodes` module should provide all we need.

PR #180, opened 2019-05-13
[11.0][FIX] product_multi_aen: error on search multiple product references

HviorForgeFlow:
Steps to reproduce:

Go to product variants and search for 2 product references: "reference 1" + ENTER + "reference 2"
Search will crash because domain is not correctly constructed.

Current behaviour 
```AssertionError: This domain is syntactically not correct: ['|', '|', '|', ['default_code', 'ilike', 'reference 1'], ['name', 'ilike', 'reference 1'], '|', '|', ['default_code', 'ilike', 'reference 2'], ['name', 'ilike', 'reference 2'], ('ean13_ids', 'in', [])]``

Desired behaviour
<class 'list'>: ['|', '|', '|', ['default_code', 'ilike', 'reference 1'], ['name', 'ilike', 'reference 1'], ('ean13_ids', 'in', []), '|', '|', ['default_code', 'ilike', 'reference 2'], ['name', 'ilike', 'reference 2'], ('ean13_ids', 'in', [])]

CC @Eficent

### DIFF
--- a/product_multi_ean/models/product_product.py
+++ b/product_multi_ean/models/product_product.py
@@ -93,16 +93,16 @@ class ProductProduct(models.Model):
 
     @api.model
     def search(self, domain, *args, **kwargs):
-        if list(filter(lambda x: x[0] == 'barcode', domain)):
-            ean_operator = list(
-                filter(lambda x: x[0] == 'barcode', domain)
-            )[0][1]
-            ean_value = list(
-                filter(lambda x: x[0] == 'barcode', domain)
-            )[0][2]
-
-            eans = self.env['product.ean13'].search(
-                [('name', ean_operator, ean_value)])
-            domain = list(filter(lambda x: x[0] != 'barcode', domain))
-            domain += [('ean13_ids', 'in', eans.ids)]
+        for sub_domain in list(filter(lambda x: x[0] == 'barcode', domain)):
+            domain = self._get_ean13_domain(sub_domain, domain)
         return super(ProductProduct, self).search(domain, *args, **kwargs)
+
+    def _get_ean13_domain(self, sub_domain, domain):
+        ean_operator = sub_domain[1]
+        ean_value = sub_domain[2]
+        eans = self.env['product.ean13'].search(
+            [('name', ean_operator, ean_value)])
+        domain = [('ean13_ids', 'in', eans.ids)
+                  if x[0] == 'barcode' and x[2] == ean_value
+                  else x for x in domain]
+        return domain
--- a/product_multi_ean/__manifest__.py
+++ b/product_multi_ean/__manifest__.py
@@ -6,7 +6,7 @@
 
 {
     'name': 'Multiple EAN13 on products',
-    'version': '11.0.1.0.0',
+    'version': '11.0.1.0.1',
     'license': 'AGPL-3',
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by rousseldenis** [comment]
@hveficent Makes sense.



> Go to product variants and search for 2 product references: "reference 1" + ENTER + "reference 2"

 Could you highlight this behaviour by adding a test line with both references ?

https://github.com/OCA/stock-logistics-barcode/blob/11.0/product_multi_ean/tests/test_product_multi_ean.py#L43

**Comment by HviorForgeFlow** [comment]
@rousseldenis, comment attended

**Review by rousseldenis — APPROVED** [review]
Code review

**Comment by HviorForgeFlow** [comment]
@pedrobaeza @xavierjimenez
FF

File: product_multi_ean/models/product_product.py [hunk 1]
```diff
@@ -93,16 +93,16 @@ def _prepare_ean13_vals(self):
...
+            domain = self._get_ean13_domain(sub_domain, domain)
         return super(ProductProduct, self).search(domain, *args, **kwargs)
+
+    def _get_ean13_domain(self, sub_domain, domain):
+        ean_operator = sub_domain[1]
+        ean_value = sub_domain[2]
+        eans = self.env['product.ean13'].search(
+            [('name', ean_operator, ean_value)])
+        domain = [('ean13_ids', 'in', eans.ids)
+                  if x[0] == 'barcode' and x[2] == ean_value
```
pedrobaeza, line 106: This is redundant

HviorForgeFlow: Maybe but it keeps proper domain format across more complex filters.

Before this patch if you look for multiple products you only get ean_ids for one of them, with this change I'm ensuring that even if redundancy or not, search domains are full of it, search has consistency.

pedrobaeza: I mean you are already checking this before calling this method.

rousseldenis: @pedrobaeza the loop is on domain, but the filter is on subdomain. Right @hveficent ?

HviorForgeFlow: Yes, here i go across the domain looking for specific subdomain that needs to be changed. Without this if there is a multiple search, it will overwrite both 'barcode' domains. I also need to check that domain contains barcode because in most common search case by product it look for name, internal reference and barcode

pedrobaeza: Uf, the code is a bit confusing. Why not using `osv.expression.OR` method for doing this in one shot?

HviorForgeFlow: It could be an option, I never used before, so I think we can move this forward to fix current issue, and maybe I will create another PR to improve readability using OR expression.

pedrobaeza: OK, I'm not blocking this PR, but I don't want also to approve without understanding full code, and now I'm afraid that I don't have time for deeper analysis, so go ahead if you think it's correct.

HviorForgeFlow: :+1:

xavierjimenez (approved):
Have checked it and everything seems to be working fine, for me it's good to merge to fix the issue

PR #188, opened 2019-07-08
[FIX] stock_scanner: singleton error

### LINKED ISSUE #187: [11.0] stock_scanner: singleton error
Expected singleton: scanner.hardware(1, 2)" while evaluating 'model.timeout_session()
Reproduce:
  Create more than one scanner.hardware records
  Use odoo-sentinel for login scenario or set manual for scanner.hardware records Last call
  Run cron "Remove the current user_id and scenario values on session timeout"

kos94ok-3D:
Closes: #187

### DIFF
--- a/stock_scanner/demo/stock_scanner_demo.xml
+++ b/stock_scanner/demo/stock_scanner_demo.xml
@@ -20,4 +20,11 @@
         <field name="screen_height">20</field>
         <field name="warehouse_id" ref="stock.warehouse0"/>
     </record>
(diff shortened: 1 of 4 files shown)


pedrobaeza:
This patch seems to be not correct, as it's changing the previous behavior. Please check also Travis status, which is red.

rousseldenis:
LGTM. @kos94ok-3D Could you add a test case that reproduces what you described in #187 to avoid regressions ?

rousseldenis (approved):
Code review

PR #189, opened 2019-07-08
[IMP] stock_scanner: pass logout if timeout_delay is set 0

kos94ok-3D:


pedrobaeza:
As you are changing the behavior, you should document this in the README for saying that putting 0 means no timeout. Any way, is this is a good idea? Leaving this without timeout will provoke possible problems. I would even say that you can already put a very great timeout for simulating no timeout and no need to change this.

kos94ok-3D:
Thanks for comment.
I agree, it is a bad idea.

Issue #193, opened 2019-07-29
[REF] Switching from viivakoodi lib to python-barcode (back)


legalsylvain:
Ref of this issue : [this comment](https://github.com/OCA/stock-logistics-barcode/pull/119#issuecomment-516057377).

> @yucer said : 
> Is there another project python-barcode that seems to be a fork from viivakoodi. It has additional releases and cleaner packaging.
> Also it has less TODOs. One of them Improve Python 3 support seems to be done. Do you thing it can be used instead of viivakoodi ?

Hi @yucer, 
Python-barcode is not a fork ! viivakoodi is a fork of python-barcode. (ref : https://github.com/OCA/stock-logistics-barcode/pull/76) And we switched because python-barcode was not active anymore. (In fact Bitbucket old links are now broken)
But well, now, python-barcode seems to be more active than viivakoodi.

It seems that @lasley is now maintainer of viivakoodi. [ref](https://github.com/OCA/stock-logistics-barcode/pull/76#issuecomment-275727514). @lasley : what do you think about that proposal ?

I'm personaly fully open to switch from a lib to another.
- I have no tecnical point of view
- regarding community activity, I'm in favor of the switch

**Viivakoodi activity :**

- 7 contributors, 0 commits since 2 years.

![image](https://user-images.githubusercontent.com/3407482/62065868-317d9980-b230-11e9-8cdd-23e43586aa7b.png)

**Python-barcode activity :** 

- 13 contributors, several commits in 2019.

![image](https://user-images.githubusercontent.com/3407482/62065836-23c81400-b230-11e9-8638-f85381bca632.png)

@yucer, do you want to make the PR to switch the lib, to see if all is OK ? 

regards.

yucer:
> @yucer, do you want to make the PR to switch the lib, to see if all is OK ?

Done let us see if there are no regressions. I suppose that the programming interface is the same.

yucer:
However, there is something new: `python-barcode` uses `Pillow` instead of `PIL` when generating raster images.

Does this project generates only vector images (.SVG) ?  

How many things in `Odoo` or `OCA` projects use `PIL` or `Pillow` features?

PR #194, opened 2019-07-30
Update dependency in setup.py from viivakoodi to python-barcode

yucer:
The `python-barcode` project is active again and it seems to enhace
the support for python3. It also uses `Pillow` instead of `PIL`.

Ref: #193

### DIFF
--- a/setup/barcodes_generator_abstract/setup.py
+++ b/setup/barcodes_generator_abstract/setup.py
@@ -5,7 +5,7 @@ setuptools.setup(
     odoo_addon={
         'external_dependencies_override': {
             'python': {
-                'barcode': 'viivakoodi',
+                'barcode': 'python-barcode',
             },
         },
     }


pedrobaeza:
Changing this dependency in a stable version is not a good idea. You can provide code for using the other one if detected, but not change the manifest external dependencies.

legalsylvain (changes requested):
Hi @yucer,

You should change also here. 
https://github.com/OCA/stock-logistics-barcode/blob/11.0/requirements.txt#L2
Could you rgrep your repo to be sure all reference to viivakoodi has been changed ? 

@lasley, please comment this change. regards.

> Changing this dependency in a stable version is not a good idea. You can provide code for using the other one if detected, but not change the manifest external dependencies.

Thanks @pedrobaeza , :+1: Better to do it for V13 serie. (coming soon !)

for that branch, you can add a text in the ROADMAP.rst file of the module(s) to mention that the change should be done when migrating.


king regards.

yucer:
Hi @legalsylvain, @lasley , @pedrobaeza 
> 
> +1 Better to do it for V13 serie. (coming soon !)
> 
Actually, I don't need it for v10.

I did just saw the dependency and try to figure it out what it was. Later I did discover that there was another project more active and we did talk about it in Ref #193.

Feel free to close the PR and submit new ones for the latest version, if the decision is to change.

Kind regards.

legalsylvain:
All right. We'll see when migrating this module. thanks.

PR #195, opened 2019-07-30
[ADD][WIP] stock_barcodes: New module to read barcodes on stocks opertions

sergio-teruel:
cc @Tecnativa


rafaelbn:
Thanks @sergio-teruel amazing work

legalsylvain:
Hi @sergio-teruel, looks promising ! Please ping me when it is not WIP anymore.
General question : i think that the module should be named "___ inventory" as the purpose of this module is to handle inventory.

regards.

pedrobaeza:
@legalsylvain its scope is going to be bigger, covering also pickings. For now, it's only for inventory, but there's already a generic wizard to be plugged to any stock model, so the name is suitable because there's no sense on splitting this in several modules depending on the same base module (`stock`), and using the same generic wizard.

legalsylvain:
> @legalsylvain its scope is going to be bigger, covering also pickings. For now, it's only for inventory, but there's already a generic wizard to be plugged to any stock model, so the name is suitable because there's no sense on splitting this in several modules depending on the same base module (stock), and using the same generic wizard.

My proposal was to split feature in several modules. ``stock_barcodes_abstract``, ``stock_barcodes_inventory`` (and then ``stock_barcodes_picking``, when it will be available).
I think that being modular is better, as I think that other features could be implemented later, like ``stock_barcodes_purchase`` to have the possibility to create purchase order, scanning products, etc...

anyway, thanks for this WIP contribution ! 

kind regards.

pedrobaeza:
That's exactly what I'm saying that we are not going to gain anything. `stock_barcodes_purchase` doesn't make sense technically speaking. When you look at the code, you'll understand it.

legalsylvain:
Hi @pedrobaeza. I looked the code, thanks. ;-)

And yes, I continue to think that ``wiz.stock.barcodes.read`` model that is an mixin model should be in an _abstract_ module, and ``wiz.stock.barcodes.read.inventory`` should be in an _inventory_ module.
(``barcodes_wizard_abstract``, ``barcodes_wizard_inventory``, or something like that).
But well, on that point, let the rest of the community talk about that point.

regards.

pedrobaeza:
The question is that the modularity in this case is not needed. If `stock_inventory` was a separate module in Odoo core, then we would do that split, but all things are in `stock` module. If you invent an extra module `stock_whatever` that needs this wizard, then depends on this one and all OK. No need of extra burden separating things.

legalsylvain:
IMHO, The modularity is needed because users can want a feature, and not the other. Dependency does'nt have nothing to do with that. The same occures with ``barcodes_generator_xxx`` modules.

* If you want to generate barcodes for lots (``stock`` modules) you install this module : 
https://github.com/OCA/stock-logistics-barcode/tree/11.0/barcodes_generator_lot
* If you want to generate barcodes for pickings (``stock`` module too) you install this module : 
https://github.com/OCA/stock-logistics-barcode/tree/11.0/barcodes_generator_picking

regards.

pedrobaeza:
This has nothing to do with `barcodes_generator_*` family modules. If you don't want barcodes in inventory, don't use it. It's totally harmless. The other modules, when installed, actually do things.

pedrobaeza:
And we are talking about 90 lines of code more for maintaining inventory specific feature, so it doesn't worth to split it.

legalsylvain:
Still not convinced on that point. :-1: but well, as said, let wait other point of view.
regards.

sergio-teruel:
Hi @legalsylvain thanks for your interest.
@pedrobaeza and I were discussing this topic, and I decided not split *_inventory from stock because this features are in the same Odoo core module.
Anyway I will wait to know more opinions about it... ok??

stock_barcodes/wizard/stock_barcodes_read.py
```diff
@@ -0,0 +1,184 @@
...
+from odoo import _, api, fields, models
+from odoo.addons import decimal_precision as dp
+
+
+class WizStockBarcodesRead(models.TransientModel):
+    _name = 'wiz.stock.barcodes.read'
+    _inherit = 'barcodes.barcode_events_mixin'
+    _description = 'Wizard to read barcode'
+    # To prevent remove the record wizard until 2 days old
+    _transient_max_hours = 48
```
legalsylvain, line 12: I don't see it very predictive. If inventory is done on more than 2 days... 
I have no magic solution, except move the model into models.Model, and purge the database regularly, but only when the res_model_id is in a done state. What do you think ?
After all, a data keeped during 48 hours is not very transient, doesn't it ?

legalsylvain (commented):
>  Anyway I will wait to know more opinions about it... ok??

yes of course. What I proposed ! If it's a matter of design, let's see what OCA members are thinking. And if it's a matter of time, I can do this refactor, as i'm sure that this module, if it works correctly, could deprecate the previous module I developped with @hpar, ``mobile_app_inventory``.

**Other review point**

On runbot, 
- create an inventory, 
- set "manual entry data"
- remove location (optional field)
- you have an error

image

I see the message not very understandable. (with False).
image

About that two fields, what about to set it mandatory ?

**Usability :**
- I think that all that part should be hidden, if the list is empty. Specially the button that doesn't do nothing (I guess, if the list is empty)

image

- I dont understand what is the objective of the interface, when "Manual entry data" is unchecked. Seems that is not possible to do anything. Right ? 

- what about to update scan_log_ids, when on change of product_id (or location_id). Otherwise, we don't know what we are unlinking, when we click on "Remove last scan".

Thanks for the precision and thanks for all that great job !

stock_barcodes/wizard/stock_barcodes_read.py
```diff
@@ -0,0 +1,184 @@
...
+            'res_model_id': self.res_model_id.id,
+            'res_id': self.res_id,
+        }
+
+    def _add_read_log(self):
+        if self.product_qty:
+            vals = self._prepare_scan_log_values()
+            self.env['stock.barcodes.read.log'].create(vals)
+
+    @api.depends('_barcode_scanned')
```
legalsylvain: where this value is defined ? propose to replace by 'product_id', 'location_id'. (and to be coherent : 'res_model_id', 'res_id' also)

stock_barcodes/wizard/stock_barcodes_read.py
sergio-teruel: The inventory process can last 2, 3 or n days, but the wizard opened from inventory is new each time that user open the scan wizard interface. I put this time to avoid that in the same day with the  same wizard opened the autovacumm deletes this wizard, I am sure that the next day the user continues to do the inventory but he crate a new wizard...
As important data, the scan logs records are linked to inventory (res_model, res_id) in this case.

stock_barcodes/wizard/stock_barcodes_read.py
sergio-teruel: Hi, the Odoo core module barcodes add this field.
This allow listen barcode scanned from hardware scanner.
https://github.com/OCA/OCB/blob/11.0/addons/barcodes/models/barcode_events_mixin.py#L14
Anyway I change this for product_id only, when a location is changed the product_id field is set to False

sergio-teruel:
@legalsylvain about your ux proposals, I have already fixed the error in manual entry mode and improve the messages information.
> I dont understand what is the objective of the interface, when "Manual entry data" is unchecked. Seems that is not possible to do anything. Right ?

Yes!! and this is the good thing!! :smile: 
The barcode scanner interface has two operational modes
1. Automatic mode (As default)
The user only has to scan barcodes in warehouse locations with a scanner hardward, the wizard read the barcode and do operations in this order:

- Try search a product, if found, is assigned to product_id field and creates or update inventory line with 1 unit. (If product has tracking by lots the interface wait for a lot to be scanned) 
- Try search a product packaging, if found, the product_id related is set, product quantities are updated and create or update inventory line.
- Try search a lot (The product is mandatory in this case so you first scan a product and then scann a lot), this lot field is not erased until that product change, so for each product scann the interface add or update a inventory line with this lot.
- Try to search a location, if found the field location is set.
If a barcode has not found, when the message is displayed you can create this lot scanning the product.

2. Manual mode
The user can use the hardware scanner too, this model allow to set quantities manually.
And the user can add a record if a product has not a barcode configured.

¿Why display scan logs?
This allow to user see other records by others users in the same inventory.

I don't know if I have been clear in my explanation... :smiley:

sergio-teruel:
Sorry @legalsylvain Can yo try with hardware scanner?

stock_barcodes/README.rst
```diff
@@ -0,0 +1,100 @@
...
+    :alt: Translate me on Weblate
+.. |badge5| image:: https://img.shields.io/badge/runbot-Try%20me-875A7B.png
+    :target: https://runbot.odoo-community.org/runbot/150/11.0
+    :alt: Try me on Runbot
+
+|badge1| |badge2| |badge3| |badge4| |badge5| 
+
+This module makes a barcode reader interface for stock module.
+
+This module contains a base wizard read barcode that it can be extended by
```
pedrobaeza: remove `it`

stock_barcodes/README.rst
```diff
@@ -0,0 +1,100 @@
...
+.. |badge4| image:: https://img.shields.io/badge/weblate-Translate%20me-F47D42.png
+    :target: https://translation.odoo-community.org/projects/stock-logistics-barcode-11-0/stock-logistics-barcode-11-0-stock_barcodes
+    :alt: Translate me on Weblate
+.. |badge5| image:: https://img.shields.io/badge/runbot-Try%20me-875A7B.png
+    :target: https://runbot.odoo-community.org/runbot/150/11.0
+    :alt: Try me on Runbot
+
+|badge1| |badge2| |badge3| |badge4| |badge5| 
+
+This module makes a barcode reader interface for stock module.
```
pedrobaeza: s/makes/provides

stock_barcodes/README.rst
```diff
@@ -0,0 +1,100 @@
...
+
+This module contains a base wizard read barcode that it can be extended by
+other modules. Inside this module the inventory wizard extends it.
```
pedrobaeza: Instead of the sentence `Inside this module the inventory wizard extends it`, put a new paragraph

`This module also makes use of this wizard for providing barcode support for doing inventories.`

stock_barcodes/README.rst
```diff
@@ -0,0 +1,100 @@
...
+
+**Table of contents**
+
+.. contents::
+   :local:
+
+Usage
+=====
+
+To use barcode interface on inventory:
```
pedrobaeza: s/use/use the

stock_barcodes/README.rst
```diff
@@ -0,0 +1,100 @@
...
+
+#. Go to a *Inventory > operations > Inventory Adjustments*.
```
pedrobaeza: remove `a`

stock_barcodes/README.rst
```diff
@@ -0,0 +1,100 @@
...
+#. Create new inventory with "Select products manually" option.
```
pedrobaeza, line 47: s/new/a new

stock_barcodes/README.rst
```diff
@@ -0,0 +1,100 @@
...
+#. Start inventory.
+#. Click to "Scan barcodes" smart button.
```
pedrobaeza, line 49: s/to/on

stock_barcodes/README.rst
```diff
@@ -0,0 +1,100 @@
...
+#. Start read barcodes.
```
pedrobaeza, line 50: s/read/reading

stock_barcodes/README.rst
```diff
@@ -0,0 +1,100 @@
...
+
+You can change to "manual entry data" to allow select data without scanner
+hardware.
+
+Known issues / Roadmap
+======================
+
+* Barcode interface for reading barcodes on pickings.
+* Excute action_done() method outside onchange environment.
```
pedrobaeza, line 107: s/Excute/Execute.

Explain a bit this point: `There can be a side effect doing create/write operations during an onchange`

stock_barcodes/README.rst
```diff
@@ -0,0 +1,100 @@
...
+* Allow create product when a barcode has not been found.
```
pedrobaeza, line 108: s/Allow/Allow to
s/product/a product

stock_barcodes/README.rst
```diff
@@ -0,0 +1,100 @@
...
+
+* Tecnativa
+
+Contributors
+~~~~~~~~~~~~
+
+
+* `Tecnativa <https://www.tecnativa.com>`_:
+
+  * Sergio Teruel
```
pedrobaeza, line 134: I think I deserve to be here :wink:

stock_barcodes/__manifest__.py
```diff
@@ -0,0 +1,25 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+{
+    "name": "Stock Barcodes",
+    "summary": "It provides read barcode on stock operations.",
+    "version": "11.0.1.0.0",
+    "author": "Tecnativa, "
+              "Odoo Community Association (OCA)",
+    "website": "https://odoo-community.org/",
+    "license": "AGPL-3",
+    "category": "Extra Tools",
```
pedrobaeza, line 11: Put the same category as `stock` module

stock_barcodes/models/stock_barcodes_read_log.py
```diff
@@ -0,0 +1,45 @@
+# Copyright 2019 Sergio Teruel <sergio.teruel@tecnativa.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from odoo import fields, models
+from odoo.addons import decimal_precision as dp
+
+
+class StockBarcodesReadLog(models.Model):
+    _name = 'stock.barcodes.read.log'
+    _description = 'Log barcode scanner'
```
pedrobaeza, line 9: Barcode scanner log

stock_barcodes/models/stock_barcodes_read_log.py
```diff
@@ -0,0 +1,45 @@
...
+    _order = 'id DESC'
```
pedrobaeza, line 10: I would add a new column `date_scan` and order by it (desc). This way, you can feed this table from other sources with historic data for example.

stock_barcodes/models/stock_barcodes_read_log.py
```diff
@@ -0,0 +1,45 @@
...
+
+    name = fields.Char(string='barcode')
+    res_model_id = fields.Many2one(
+        comodel_name='ir.model',
+        index=True,
+    )
+    res_id = fields.Integer(index=True)
+    product_id = fields.Many2one(
```
pedrobaeza, line 18: Not sure about adding these fields... This is a barcode scanning log, not something about the later handle of that barcode.

stock_barcodes/security/ir.model.access.csv
```diff
@@ -0,0 +1,2 @@
+"id","name","model_id:id","group_id:id","perm_read","perm_write","perm_create","perm_unlink"
+access_stock_barcodes_read_log,access_stock_barcodes_read_log,model_stock_barcodes_read_log,base.group_user,1,1,1,1
```
pedrobaeza, line 2: I don't think we should grant full access to all employees, even not for warehouse user/manager. This is something automatic, so people mustn't touch it, only read it.

Thus, we should grant read access for warehouse users, and do create operations with sudo.

stock_barcodes/static/src/js/basic_field.js
```diff
@@ -0,0 +1,25 @@
...
+ * License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). */
+
+odoo.define('stock_barcodes.FieldFloatNumericMode', function(require) {
+    'use strict';
+
+    var basic_fields = require('web.basic_fields');
+    var field_registry = require('web.field_registry');
+
+    var FieldFloatNumericMode = basic_fields.FieldFloat.extend({
+        events: _.extend({}, basic_fields.FieldFloat.prototype.events, {
```
pedrobaeza, line 11: What is this doing? Put some docstring

stock_barcodes/tests/test_stock_barcodes.py
```diff
@@ -0,0 +1,145 @@
+# Copyright 2108-2019 Sergio Teruel <sergio.teruel@tecnativa.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo.tests.common import TransactionCase
+
+
+class TestStockBarcodes(TransactionCase):
```
pedrobaeza, line 7: SavepointCase?

stock_barcodes/tests/test_stock_barcodes.py
```diff
@@ -0,0 +1,145 @@
...
+            'name': '8411822222568',
+            'product_id': self.product_tracking.id,
+        })
+        self.wiz_scan = self.WizScanRead.new()
+
+    def action_barcode_scanned(self, wizard, barcode):
+        wizard._barcode_scanned = barcode
+        wizard._on_barcode_scanned()
+
+    def test_wizard_scan_location(self):
```
pedrobaeza, line 76: Don't do these test in this class, as they are going to be executed also on children classes.

stock_barcodes/views/stock_inventory_views.xml
```diff
@@ -0,0 +1,23 @@
...
+    <record id="view_inventory_form" model="ir.ui.view">
+        <field name="model">stock.inventory</field>
+        <field name="inherit_id" ref="stock.view_inventory_form"/>
+        <field name="arch" type="xml">
+            <div name="button_box" position="inside">
+                <button name="action_barcode_scan"
+                    class="oe_stat_button"
+                    icon="fa-barcode"
+                    type="object"
+                    help="List view of lines"
```
pedrobaeza, line 13: This help is not correct

stock_barcodes/views/stock_inventory_views.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+
+    <record id="view_inventory_form" model="ir.ui.view">
+        <field name="model">stock.inventory</field>
+        <field name="inherit_id" ref="stock.view_inventory_form"/>
+        <field name="arch" type="xml">
+            <div name="button_box" position="inside">
+                <button name="action_barcode_scan"
```
pedrobaeza, line 9: Shouldn't this be only showed when the inventory mode of manually entries has been selected?

stock_barcodes/wizard/stock_barcodes_read.py
```diff
@@ -0,0 +1,191 @@
...
+        digits=dp.get_precision('Product Unit of Measure'),
+    )
+    product_qty = fields.Float(
+        string='Product Qty',
+        digits=dp.get_precision('Product Unit of Measure'),
+    )
+    manual_entry = fields.Boolean(
+        string='Manual entry data',
+    )
+    scan_log_ids = fields.Many2many(
```
pedrobaeza, line 49: ~~Why not a direct one2many with domain over `res_model` and inverse_name = `res_id`? We have done this before with good results (example: `base_custom_info`)~~

Already seen that this provides "realtime" update. Put a comment on the field for specifying it.

stock_barcodes/wizard/stock_barcodes_read.py
```diff
@@ -0,0 +1,191 @@
...
+
+    @api.onchange('packaging_qty')
+    def onchange_packaging_qty(self):
+        if self.packaging_id:
+            self.product_qty = self.packaging_qty * self.packaging_id.qty
+
+    def _set_messagge_info(self, type, messagge):
+        """
+        Set message type and message description.
+        For manual entry mode barcode is no set so is not displayed
```
pedrobaeza: s/no/not

stock_barcodes/wizard/stock_barcodes_read.py
```diff
@@ -0,0 +1,191 @@
...
+            if len(packaging) > 1:
+                self._set_messagge_info(
+                    'more_match', _('More than one package found'))
+                return
+            self.action_packaging_scaned_post(packaging)
+            self.action_done()
+            return
+        lot = self.env['stock.production.lot'].search([
+            ('name', '=', barcode),
+            ('product_id', '=', self.product_id.id),
```
pedrobaeza: What happens if no product set?

stock_barcodes/wizard/stock_barcodes_read_inventory.py
```diff
@@ -0,0 +1,87 @@
...
+            'product_id': self.product_id.id,
+            'location_id': self.location_id.id,
+            'product_uom_id': self.product_id.uom_id.id,
+            'product_qty': self.product_qty,
+            'prod_lot_id': self.lot_id.id,
+        }
+
+    def _add_inventory_line(self):
+        StockInventoryLine = self.env['stock.inventory.line']
+        line = first(StockInventoryLine.search([
```
pedrobaeza: Add a `_prepare_domain` method to this for better extensibility.

stock_barcodes/wizard/stock_barcodes_read_inventory.py
```diff
@@ -0,0 +1,87 @@
...
+        result = super().action_manual_entry()
+        if result:
+            self.action_done()
+        return result
+
+    def reset_qty(self):
+        super().reset_qty()
+        self.inventory_product_qty = 0.0
+
+    def action_remove_last_scan(self):
```
pedrobaeza: shouldn't be this renamed to `undo` instead of `remove`?

pedrobaeza (changes requested):
Technical review done. I like to meet you again for online demo.

kos94ok-3D:
I found bug.
Step to reproduce:
1. Go to a Inventory > operations > Inventory Adjustments.
2. Create new inventory with "Select products manually" option.
3. Start inventory.
4. Click to "Scan barcodes" smart button.
5. Start read barcodes and get 1 qty
6. Back to Inventory and remove created inventory line
7. Click to "Scan barcodes" smart button
8. Start read barcodes and get 1
9. Click to "Remove last scan" smart button two times
Bug: qty == -1.0
изображение
изображение

stock_barcodes/wizard/stock_barcodes_read.py
kos94ok-3D: I think it need move to wiz.stock.barcodes.new.lot, because it code work only from this model as I see.

sergio-teruel: As explain in readme (last update) you need first scan a product, if product has tracking by lot the wizard informs user that is waiting to scan a lot...

sergio-teruel:
@pedrobaeza changes done!!

sergio-teruel:
@kos94ok-3D The field inventory quantities always displays the real inventory lines quantities even though scanned quantities log are more.. 
Already fixed negative units in this case...

stock_barcodes/wizard/stock_barcodes_read_inventory.py
```diff
@@ -31,14 +31,24 @@ def _prepare_inventory_line(self):
...
+
     def _add_inventory_line(self):
         StockInventoryLine = self.env['stock.inventory.line']
-        line = first(StockInventoryLine.search([
-            ('inventory_id', '=', self.inventory_id.id),
-            ('product_id', '=', self.product_id.id),
-            ('location_id', '=', self.location_id.id),
-            ('prod_lot_id', '=', self.lot_id.id),
-        ]))
+        line = first(StockInventoryLine.search(
```
kos94ok-3D: Why use first function instead of parameter limit=1?

stock_barcodes/wizard/stock_barcodes_read.py
```diff
@@ -0,0 +1,193 @@
...
+        ])
+        if lot:
+            self.lot_id = lot
+            return
```
kos94ok-3D: I think is good idea set qty after scan lot.
Example:
I have product traking by serial number.
I do inventory adjastment (steps start after open wiz.stock.barcodes.read from view):
1. scan product
2. scan lot
3. scan product
4. scan lot
5. scan product
count_steps = (count_lots * 2) + 1
If add function like:
> def action_lot_scaned_post(self):
>> self.product_qty = 0.0 if self.manual_entry else 1.0

and call it with action_done()

In the same case we have steps:
1. scan product
2. scan lot
3. scan lot
4. scan lot
count_steps = count_lots + 1
@pedrobaeza What do you think about it?

sergio-teruel:
@legalsylvain autofocus to product qty or packaging qty done in this commit https://github.com/OCA/stock-logistics-barcode/pull/195/commits/95166ce192150d4adc0217871173bf0ac742cef3

stock_barcodes/wizard/stock_barcodes_read.py
sergio-teruel: Is other point of view... I will study the case for product with tracking 'serial'.

stock_barcodes/wizard/stock_barcodes_read_inventory.py
sergio-teruel: Yes!!!! Fixed here https://github.com/OCA/stock-logistics-barcode/pull/195/commits/251cafa67c5fb70e9c92a38397106be31c4fe2e5

stock_barcodes/wizard/stock_barcodes_read.py
sergio-teruel: Ok.. 
I have implemented here https://github.com/OCA/stock-logistics-barcode/pull/195/commits/8a852ca3da3e5a277aaeb2b2d6f126270ddfd1f6
Thanks!!

rafaelbn (approved):
This is not WIP and is in production, tested and working :+1:

pedrobaeza:
@legalsylvain @kos94ok-3D any final comment before merging?

pedrobaeza (approved):

pedrobaeza:
This must be squashed on merge.

kos94ok-3D (approved):
LGTM

legalsylvain:
> @legalsylvain autofocus to product qty or packaging qty done in this commit 95166ce

Thanks !

> @legalsylvain @kos94ok-3D any final comment before merging?

Well, too late. ;-)

pedrobaeza:
Well, if you detect any problem, tell us and we can do later pull requests.

PR #192, opened 2019-07-22
[MIG][11.0] base_gs1_barcode: Migration

sergio-teruel:
cc @Tecnativa

PR #197, opened 2019-07-30
GS1 Barcode API

mediaengagers:


pedrobaeza:
Please preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-12.0 (using the origin branch that corresponds).

See also the migration in progress in #192

PR #200, opened 2019-08-13
12.0 mig barcodes generator abstract

cvinh:
works out of the box in v12


phatnguyenuit (changes requested):
Missing file `requirements.txt` at the top level of this repository. Refer the old one to get it.

barcodes_generator_abstract/models/barcode_generate_mixin.py
```diff
@@ -0,0 +1,101 @@
...
+try:
+    import barcode
+except ImportError:
+    _logger.debug("Cannot import 'viivakoodi' python library.")
+    barcode = None
+
+
+class BarcodeGenerateMixin(models.AbstractModel):
+
+    _name = 'barcode.generate.mixin'
```
pedrobaeza, line 23: Add `_description` for avoiding runbot warning

pedrobaeza (commented):
And you should check #193 for an alternative library.

barcodes_generator_abstract/models/barcode_generate_mixin.py
cvinh: About the library, tell me if I should change it
About _description, I will do a force push tomorrow

pedrobaeza: @legalsylvain @yucer what do you think?

legalsylvain: Hi. As said in #193, I have no point of view, regarding PIL / Pillow neither regarding the quality of the code of both librairies.
But the othe lib has more contributors. So I'm in favour to switch, if it works correctly. 

@cvinh, could you test and check if all is OK ? 

regards.

cvinh: @pedrobaeza could you help with the runbot error ? I don't see where it fails... thanks in advance

cvinh: @gurneyalex could you help with the runbot error ? I don't see where it fails... thanks in advance

pedrobaeza:
The warning is:
```
2019-08-13 23:03:45,710 161 WARNING openerp_test odoo.addons.base.models.ir_model: Two fields (sequence_id, sequence) of barcode.rule() have the same label: Sequence.
```

phatnguyenuit (approved):

legalsylvain (approved):
/ocabot merge

LGTM. Thanks for porting this module.

barcodes_generator_abstract/models/barcode_generate_mixin.py
yucer: @pedrobaeza sorry for the late response. I am not too much familiarized with these modules. Just noted that there were two alternate python libraries for the same feature, and the one not used had more activity.

Thanks for the feed back.

pedrobaeza: Module was already merge, so a bit late I'm afraid. Any way, we have switched the library.

PR #205, opened 2019-09-02
[IMP] product_multi_ean: Make generalize multi barcodes. Remove check EAN restrictions

sergio-teruel:
cc @Tecnativa


pedrobaeza:
Travis is red and maybe this should be forked on a module `product_multi_barcode` instead of doing all of these changes, only limiting the change in this module to not check EAN13 of main field.

pedrobaeza:
@sergio-teruel please take into account my comments.

pedrobaeza:
You can limit the changes in this PR to the essential for not touching `barcode` field (for example, recreating previous `ean13` field as main target, and filling `barcode` with that main `ean13`, but without checking EAN format.

PR #207, opened 2019-09-03
[12.0][MIG] stock_barcodes

astirpe:


stock_barcodes/__manifest__.py
```diff
@@ -3,23 +3,23 @@
 {
     "name": "Stock Barcodes",
     "summary": "It provides read barcode on stock operations.",
-    "version": "11.0.1.0.0",
+    "version": "12.0.1.0.0",
     "author": "Tecnativa, "
               "Odoo Community Association (OCA)",
     "website": "https://odoo-community.org/",
```
AdriaGForgeFlow: Use repository website: https://github.com/OCA/stock-logistics-barcode

AdriaGForgeFlow (commented):
Some comments below:

Code looks good to me but I have found some functional aspects that could be improved.

* Regarding the scanned logs, a column for the lot could be added to help the user see which lots have already been introduced in the inventory adjustment. Currently, the barcode column shows the last barcode scanned which in this case is the Lot # but it's not very clear.
image
* Whenever the user finishes scanning products, to go back to the inventory adjustment main screen a Warning appears since changes have not been saved which is OK because is not needed. IMO warning dialog should not appear (I have not looked into it and I don't know whether is possible or not)
image

astirpe:
@ageficent 

For displaying the lots in the scanned logs, I could just add `lot_id` in the view of the scanned logs. Would it be fine for you?
Anyway I agree that the scanned logs view, as it is at this moment, can be confusing. For a customer of us I modified the scanned logs view completely, but I think it would be out of scope for this migration PR.

For disabling the warning, I don't see a simple way to silent it (if you find it please tell me).
Setting the `action_stock_barcodes_read` action target to `new`, instead of `current`, avoids the warning but creates other issues: https://github.com/OCA/stock-logistics-barcode/blob/d6b072d9f7f803aba7ced3d1d7da35528e1116b5/stock_barcodes/wizard/stock_barcodes_read_views.xml#L117-L118
I tried to set the `action_stock_barcodes_read` action target to `new` and automatically re-launch the action every time the view is closed after a button is pressed. Works fine but I think the changes on the code are kind of invasive. Do we really need those changes? Otherwise I would keep the view as it is.

AdriaGForgeFlow:
@astirpe 

I think adding the `lot_id` would be helpful and maybe remove the `name` field since it does not have any special meaning, it's just the last barcode scanned (maybe edit the string to Last Barcode Scanned).

Regarding the warning if you edit the javascript file can be done, even if it's not the cleanest way, here is the code you would need:
```
canBeDiscarded: function (recordID) {
            if (!this.modelName == "wiz.stock.barcodes.read.inventory") {
                return this._super.apply(this, arguments);
            }
            return $.when(false);
        }, 
```

Let me know what you think.

pedrobaeza:
Don't remove the name. When you have lots like GS1, it's important to have it as they can be different according lot expiry and other information that can be added as well.

astirpe:
@ageficent Done. Thanks!

AdriaGForgeFlow (approved):

pedrobaeza:
We are completing the picking part for v11, so I think you should wait for that for doing this migration.

stock_barcodes/models/stock_barcodes_read_log.py
```diff
@@ -0,0 +1,42 @@
+# Copyright 2019 Sergio Teruel <sergio.teruel@tecnativa.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from odoo import fields, models
+from odoo.addons import decimal_precision as dp
+
+
+class StockBarcodesReadLog(models.Model):
+    _name = 'stock.barcodes.read.log'
```
phatminh, line 8: I suggest like this for use "..." instead '....' when using `_name=`, Ex: `"stock.barcodes.read.log"`

stock_barcodes/models/stock_inventory.py
```diff
@@ -0,0 +1,24 @@
+# Copyright 2019 Sergio Teruel <sergio.teruel@tecnativa.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from odoo import models
+
+
+class StockInventory(models.Model):
+    _inherit = ['barcodes.barcode_events_mixin', 'stock.inventory']
+    _name = 'stock.inventory'
```
phatminh, line 8: Same my suggestions above.

stock_barcodes/wizard/stock_barcodes_read.py
```diff
@@ -0,0 +1,196 @@
+# Copyright 2019 Sergio Teruel <sergio.teruel@tecnativa.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from odoo import _, api, fields, models
+from odoo.addons import decimal_precision as dp
+
+
+class WizStockBarcodesRead(models.AbstractModel):
+    _name = 'wiz.stock.barcodes.read'
```
phatminh, line 8: Same above.

stock_barcodes/wizard/stock_barcodes_read_inventory.py
```diff
@@ -0,0 +1,99 @@
+# Copyright 2019 Sergio Teruel <sergio.teruel@tecnativa.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from odoo import _, fields, models
+from odoo.fields import first
+from odoo.addons import decimal_precision as dp
+
+
+class WizStockBarcodesReadInventory(models.TransientModel):
+    _name = 'wiz.stock.barcodes.read.inventory'
```
phatminh, line 9: Same above.

stock_barcodes/wizard/stock_production_lot.py
```diff
@@ -0,0 +1,37 @@
+# Copyright 2019 Sergio Teruel <sergio.teruel@tecnativa.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from odoo import fields, models
+
+
+class WizStockBarcodesNewLot(models.TransientModel):
+    _inherit = 'barcodes.barcode_events_mixin'
```
phatminh, line 7: Same above.

phatminh (commented):
Please check all my suggest. Thanks!

stock_barcodes/models/stock_barcodes_read_log.py
pedrobaeza: Why? Both quote style is perfectly fine with Python and with our guidelines.

rousseldenis: Agree with @pedrobaeza. Until we introduce black :smile:

astirpe:
@pedrobaeza since https://github.com/OCA/stock-logistics-barcode/pull/214 is still a work in progress, I would prefer to have this PR merged first and then port the new developments in a second time. What do you think?

pedrobaeza:
Well, it's well tested and working, and this is still not reviewed. Why not using this PR if you need it as we do in our customers through git-aggregator?

gurneyalex (approved):

pedrobaeza:
I think this is missing last PR we have made for pickings and merged last week, isn't it?

astirpe:
@pedrobaeza yes, that part is missing. If someone of your team wants to supersede this PR to add that part, he's very welcome :)

pedrobaeza:
You can cherry pick it without more effort in principle. We are not working with this in v12 for now.

astirpe:
But I don't have time to test it. That's why I was proposing to merge this as it is.

pedrobaeza:
The problem merging as is will be that other may take this one for migrating to 13, and then the feature set will be out of sync between versions.

@sergio-teruel are you able to supersed this with missing feature?

pedrobaeza:
#223 contains all the changes, so if you don't mind, can we close it superseded by that one?

astirpe:
Sure!

PR #208, opened 2019-09-04
[IMP] product_multi_ean: Remove check EAN restrictions

sergio-teruel:
cc @Tecnativa
Close other PR: https://github.com/OCA/stock-logistics-barcode/pull/205
Only remove Ean check restriction

### DIFF
--- a/product_multi_ean/models/product_product.py
+++ b/product_multi_ean/models/product_product.py
@@ -30,18 +30,6 @@ class ProductEan13(models.Model):
 
     @api.multi
     @api.constrains('name')
-    @api.onchange('name')
-    def _check_name(self):
-        barcode_obj = self.env['barcode.nomenclature']
-        for record in self.filtered('name'):
-                if not barcode_obj.check_ean(record.name):
-                    raise UserError(
-                        _('You provided an invalid "EAN13 Barcode" reference. '
-                          'You may use the "Internal Reference" '
-                          'field instead.'))
-
-    @api.multi
-    @api.constrains('name')
     def _check_duplicates(self):
         for record in self:
             eans = self.search(
--- a/product_multi_ean/i18n/product_multi_ean.pot
+++ b/product_multi_ean/i18n/product_multi_ean.pot
@@ -77,10 +77,3 @@ msgstr ""
 #, python-format
 msgid "The EAN13 Barcode \"%s\" already exists for product \"%s\""
 msgstr ""
-
-#. module: product_multi_ean
-#: code:addons/product_multi_ean/models/product_product.py:39
-#, python-format
-msgid "You provided an invalid \"EAN13 Barcode\" reference. You may use the \"Internal Reference\" field instead."
-msgstr ""
-
--- a/product_multi_ean/tests/test_product_multi_ean.py
+++ b/product_multi_ean/tests/test_product_multi_ean.py
@@ -29,10 +29,6 @@ class TestProductMultiEan(TransactionCase):
         self.assertEqual(self.product_1.ean13_ids.name, self.product_1.barcode)
 
     def test_set_incorrect_ean(self):
-        with self.assertRaises(Exception):
-            self.product_1.barcode = '1234567890123'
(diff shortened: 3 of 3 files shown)


product_multi_ean/models/product_product.py
```diff
@@ -28,18 +28,6 @@ class ProductEan13(models.Model):
         required=True,
     )
 
-    @api.multi
-    @api.constrains('name')
-    @api.onchange('name')
-    def _check_name(self):
-        barcode_obj = self.env['barcode.nomenclature']
-        for record in self.filtered('name'):
-                if not barcode_obj.check_ean(record.name):
```
rousseldenis, line 37: In fact, the function called in this check was misused.

You cannot just remove this but change the logic.

You have to call this :

https://github.com/odoo/odoo/blob/11.0/addons/barcodes/models/barcodes.py#L125

And then, call :

https://github.com/odoo/odoo/blob/11.0/addons/barcodes/models/barcodes.py#L57

with the first provided encoding.

If you want to implement a new one (GS1), simply override these functions

rousseldenis (dismissed):
IMHO, the need is to change the implementation as it was not correct before.

The removal of this check is not needed

product_multi_ean/models/product_product.py
sergio-teruel: By default, Odoo allow add any barcode to product barcode field, ¿Why this module add this restriction?

sergio-teruel: I Know that this restrictions are used when a barcode is readed... but not when a product is created

rousseldenis: You have to ask the author :smile: 

See refactoring commit :

https://github.com/OCA/stock-logistics-barcode/commit/3fa99ac3a558d1913f72fbe5549346fd0fd3d8cb#diff-71824b7fc2c784066fee1144d32b283a

@pedrobaeza

pedrobaeza: That made sense when done on previous version, but not now. It's something that should be taken into account when migrating

pedrobaeza (approved):

rousseldenis:
Maybe is it too much change on this stable version as users used to this module will wait for constrain triggering (and maybe after update will not pay attention to errors that do not appear anymore...). IMHO ok for v12

pedrobaeza:
@rousseldenis this module was merged not too much time ago, and the problem is not the behavior. It's that this breaks any other module/behavior adding other type of barcodes, so this is simply broken and must be fixed, no matter change of behavior.

rousseldenis:
That's why I proposed to change constraint implementation to make it work for any type of barcode.

I would agree for a constraint deletion only if you implement another module for that purpose (that will allow multi ean with/without constraint).

Or with an option.

pedrobaeza:
Odoo had a constraint on previous versions, and this module mimmicks the same for multi-EAN. Now there's no constraint on Odoo, so this module mustn't have as well. Please don't complicate this break-fixing PR.

rousseldenis:
What I said is that you are not the only users of this module and I try to not change the behaviour for existing users.

But as I'm not part of these users, do what you think.

pedrobaeza:
Thanks for your trust, @rousseldenis I can assure you that the feature is misused here for the main barcode. @sergio-teruel any way, what about providing an alternative removing `barcode` as computed and providing other field for that?

sergio-teruel:
We can create a field "ean13" (like odoo had in older versions) but I think that user can be confuse if see two fields for barcodes in product view...

pedrobaeza:
What do you think, @rousseldenis ?

Joaco1980:
I do not find the problem of removing a restriction that the barcode field does not have without installing the multi ean module. And if you put a selector in inventory settings if you want to validate bar codes like ean13.
There are many types of barcode coding, why put limitations on only one type of coding?

pedrobaeza:
@Joaco1980 please approve the PR then using GitHub mechanism.

Joaco1980 (approved):

PR #219, opened 2019-10-21
[10.0][FIX] travis.yml

AdriaGForgeFlow:
Travis postgres fix

cc ~ @Eficent

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ sudo: false
 cache: pip
 
 addons:
+  postgresql: "9.6"
   apt:
     packages:
       - expect-dev  # provides unbuffer utility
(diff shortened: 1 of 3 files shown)


LoisRForgeFlow (approved):
travis now works, error is outside of the scope of this PR

pedrobaeza:
Please take the occasion to remove TRANSIFEX stuff and isolated build

LoisRForgeFlow:
ups, cancelling the merge

pedrobaeza (approved):

LoisRForgeFlow:
@pedrobaeza travis error is now in stock_scanner itself, what do we do? a manual merge?

pedrobaeza:
We should fix it. In this case, it's not very complicated and it's due to a change in the logger: just use `mute_logger` for silencing the ERROR in the log, but still raising the exception of the test. Do yo know how to use it?

LoisRForgeFlow:
yes, a decorator in tests, but what should we mute? I don't see the relation, it is a recursive data error.

pedrobaeza:
The problem is that the Travis "crawler" see the line with `ERROR` and thinks that it's not OK, but it's just the error that is provoked and detected by the test, so it's the expected behavior. We have to mute `openerp_test odoo.model`, as the line is:
```
2019-10-21 15:44:12,218 6013 ERROR openerp_test odoo.models: Exception while validating constraint
```

LoisRForgeFlow:
isn't it too much to mute odoo.models??

pedrobaeza:
You can do it for just that specific test or operation. See for example https://github.com/OCA/margin-analysis/blob/cc73f9d63ca8c62837a85df1e6e4edea78f39b5f/product_price_history/wizard/historic_prices.py#L41

LoisRForgeFlow:
Ok, thanks for the insights!

pedrobaeza:
Uhm, it seems it doesn't like it at `with level`... Try mutting the test method

LoisRForgeFlow:
let's see now.

/ocabot merge patch

pedrobaeza:
Agh, I don't understand... I think the problem is the logger route, that is not that simple one... @etobella you had to do this in the past. Do you remember how?

LoisRForgeFlow:
Somotime today we'll get it merged I'm sure :rofl: :rofl: 

/ocabot merge patch

pedrobaeza:
Bravo!

Issue #224, opened 2019-11-06
Picking generation using barcode scanning


AdriaGForgeFlow:
With the current stock barcodes module we are able to manage a certain part of the picking workflow. When the picking is already created we can go to the scanning wizard and start scanning the products that are present in the stock moves increasing the quantity done of the move itself. We can also validate pickings from the scanning wizard.

We have recently encountered with the need of the other part (picking generation using barcode scanning). It would be nice to be able to create the stock moves using the barcode scanner. In this way, a user would create an immediate or a planned transfer and start scanning right away. If the stock move with the scanned product is not already created, we generate a new stock move, otherwise we just add a quantity to the specific picking line.

What do you think would be the best approach to do this? @sergio-teruel @pedrobaeza 

cc ~ @jbeficent @joansis @nuriamartinxifre

pedrobaeza:
For me, there shouldn't be an automatic picking generation, but automatic lines generation. AFAIK, this is already happening for `stock.move.line`, but not sure about non existing `stock.move`.

AdriaGForgeFlow:
> For me, there shouldn't be an automatic picking generation, but automatic lines generation. AFAIK, this is already happening for `stock.move.line`, but not sure about non existing `stock.move`.

Yes I didn't mean automatic generation for the pickings. The user manually creates the picking and then using the barcode scanner it introduces the products that are supposed to be in the picking generating the corresponding stock moves automatically.

pedrobaeza:
An extra switch for allowing that should be added. Something like "Free insert", that performs that operation independently from the existing stock.move (but only if not found).

eLBati:
Adding this functionality to 12.0 with https://github.com/OCA/stock-logistics-barcode/pull/288

pedroslopez:
> For me, there shouldn't be an automatic picking generation

@pedrobaeza Why is this not desired? It seem to me that Odoo's stock_barcode app available in enterprise allows this. You can scan a picking type, then scan all products, scan a location if wanted and validate all by using barcodes.

A client is wanting this flow, I'm curious as to why there shouldn't be auto picking generation and if there are maybe consequences I'm not seeing with this.

pedrobaeza:
Pickings are generated from their sources (sales, purchases, etc), so there's no point on creating dangling pickings and will create a quick way for doing that instead of looking the proper one.

pedroslopez:
[...]

PR #226, opened 2019-11-11
[11.0][stock_barcodes] Allow to add new stock moves to a picking

JordiBForgeFlow:
This PR introduces a new toggle 'Free Insert' in the barcodes wizard. When a user uses this toggle and scans a product or a lot, the system will create a new stock move and move line and add it to the picking.

This helps in processes like internal transfers, where you do not have a predefined picking like in incoming shipments or delivery orders. What the operator wants to do in this case is just do an unplanned move for a product from location A to location B.

### DIFF
--- a/stock_barcodes/__manifest__.py
+++ b/stock_barcodes/__manifest__.py
@@ -12,6 +12,7 @@
     "depends": [
         'barcodes',
         'stock',
+        'stock_barcodes_automatic_entry'
     ],
     "data": [
         'security/ir.model.access.csv',
--- a/stock_barcodes/models/stock_picking.py
+++ b/stock_barcodes/models/stock_picking.py
@@ -13,7 +13,6 @@ class StockPicking(models.Model):
         manual_entry = False
         if self.state in ['draft', 'waiting', 'confirmed']:
             free_insert = True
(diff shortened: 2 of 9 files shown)


JordiBForgeFlow:
@sergio-teruel @pedrobaeza @Tecnativa

rousseldenis:
@jbeficent Could you fix travis ?

sergio-teruel:
@jbeficent Why not create a new wizard like as inventory and you work directly with stock.move.lines?
Maybe it's the simplest way...

JordiBForgeFlow:
Superseeded by https://github.com/OCA/stock-logistics-barcode/pull/227

pedrobaeza:
But they are different features IMO: one thing is to move from one location to other, and another dispatch delivery orders or receipts.

LoisRForgeFlow:
Related to https://github.com/OCA/stock-logistics-barcode/issues/224

rafaelbn:
@JordiBForgeFlow this is an old one, would you like to update it or close it?

eLBati:
Porting to 12.0 with https://github.com/OCA/stock-logistics-barcode/pull/288

PR #228, opened 2019-11-14
[11.0][base_gs1_barcode,stock_barcodes_gs1] add a gs1 barcode mapping

JordiBForgeFlow:
This PR defines a new model in `base_gs1_barcode` to map GS1 application identifiers to any model and field.

The module 'stock_barcodes_gs1' then implements a default mapping to the model `wiz.stock.barcodes.read`.

That way, a company can decide to change the default mapping, and use a different application identifier for the fields that are read from GS1. Currently the fields mapped are package, product, lot, and quantity.

cc @sergio-teruel @pedrobaeza @Tecnativa @lreficent


JordiBForgeFlow:
Not a good idea to implement it this way. Closing for now...

PR #230, opened 2019-11-14
[11.0][fix][base_gs1_barcode] add extra logic to determine the group separator

JordiBForgeFlow:
Search if the barcode contains the \x1D group separator already, and then use it. Some scanners are able to pass this group separator, and others are not able. If a user works with both devices, this is becomes the most effective mechanism to ensure co-existence.

cc @lreficent

### DIFF
--- a/base_gs1_barcode/models/gs1_barcode.py
+++ b/base_gs1_barcode/models/gs1_barcode.py
@@ -83,9 +83,15 @@ class GS1Barcode(models.Model):
             return fields.Date.to_string(date)
 
         # Prefix and Group Separator
+        # Search if the barcode contains the \x1D group separator already,
+        # and then use it. Some scanners are able to pass this group separator,
+        # and others are not able. If a user works with both devices,
+        # this is becomes the most effective mechanism to ensure co-existence.
+        if '\x1D' in barcode_string:
+            separator = '\x1D'
+        else:
+            separator = self.env.user.gs1_barcode_separator or '\x1D'
         prefix = self.env.user.gs1_barcode_prefix or ''
-        separator = self.env.user.gs1_barcode_separator or '\x1D'
-
         if not barcode_string.startswith(prefix):
             raise exceptions.ValidationError(
                 _('Could not decode barcode : wrong prefix - the code should '


LoisRForgeFlow (approved):
LGTM :+1:

LoisRForgeFlow:
@sergio-teruel @pedrobaeza is this ok for you? it should not affect any existing installation.

sergio-teruel (approved):

pedrobaeza:
I don't get why do you need this. If there's other separator, configure it in the user.

LoisRForgeFlow:
@pedrobaeza Our use case is that we want users to be able to work with 2 different devices (of different brands). Why? because one is mobile and the other is static and attached to a work station. What happens is that one of the devices is internally parsing the separator '\x1D' and you never get it into Odoo, while the other device does not parse anything and you get into Odoo the plain string.

This is the simplest workaround we found to make this happen, we assume that if you have '\x1D' in the barcode read then the hardware hasn't parsed it at all.

base_gs1_barcode/models/gs1_barcode.py
```diff
@@ -83,9 +83,17 @@ def normalize_date(datestring):
             return fields.Date.to_string(date)
 
         # Prefix and Group Separator
+        # Search if the barcode contains the \x1D group separator already,
+        # and then use it. Some scanners are able to pass this group separator,
+        # and others are not able. If a user works with both devices,
+        # this is becomes the most effective mechanism to ensure co-existence.
+        gs1_group_separators = \
+            [m.start() for m in re.finditer('\x1D', barcode_string)]
+        if gs1_group_separators:
```
pedrobaeza: Why not simply:
```python
if '\x1D' in barcode_string:
    separator = '\x1D'
```

pedrobaeza:
Thanks for the explanation. I get it now.

base_gs1_barcode/models/gs1_barcode.py
JordiBForgeFlow: Done!

pedrobaeza (approved):

PR #232, opened 2019-12-03
fix(barcodes_generator_partner): remove dependency to point of sale a…

agent-z28:
…nd add dep to stock

place the fields in res_partner misc view and Add a Report to generate the Partner barcode label


rousseldenis:
@agenterp Could you fix travis ?

agent-z28:
The problem is not from our side.

Its from related Modules!

Mit freundlichen Grüßen

We look forward to see you. Best Regards

Dipl. Ing. (Fh) Georg Notter

Agent ERP GmbH

www.agenterp.com

rousseldenis:
@agenterp That seems to be related to the modified module.

See: 

image

Maybe you could rebase and try to push (did you succeed to run tests locally ?) ?

agent-z28:
@rousseldenis i can rebase.. We just changed views..

agent-z28:
Shit somehow on my mobile i always close the Pr :((:/( i dont Know how it happens.

Second time already

legalsylvain (changes requested):
I think that this refactor introduce bad design. You are redefining a field barcode on res partner model and view that is already defined in point of sale module.
This will generate errors on views if point of sale is installed. 
Also you add new feature on report that has nothing to do with the objective of this module. 
I'm agree that odoo core is bad designed (barcode declaration in wrong module) but this PR is not the good way to fix it.

Regards.

agent-z28:
Actually we just changed the view reference and Added a Report to print that barcode.

Everything Else Was already there.

Mit freundlichen Grüßen

We look forward to see you. Best Regards

Dipl. Ing. (Fh) Georg Notter

Agent ERP GmbH

www.agenterp.com

barcodes_generator_partner/views/view_res_partner.xml
```diff
@@ -7,12 +7,12 @@ License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
...
 <odoo>
-
     <record id="view_res_partner_form" model="ir.ui.view">
         <field name="model">res.partner</field>
-        <field name="inherit_id" ref="point_of_sale.view_partner_property_form"/>
+        <field name="inherit_id" ref="base.view_partner_form"/>
         <field name="arch" type="xml">
-            <field name="barcode" position="before">
+            <xpath expr="//field[@name='industry_id']" position="after">
+                <field name="barcode"/>
```
legalsylvain, line 15: This line is new.

barcodes_generator_partner/models/res_partner.py
```diff
@@ -4,8 +4,11 @@
...
+from odoo import fields
 
 
 class ResPartner(models.Model):
-    _name = 'res.partner'
-    _inherit = ['res.partner', 'barcode.generate.mixin']
+    _name = "res.partner"
+    _inherit = ["res.partner", "barcode.generate.mixin"]
+
+    barcode = fields.Char(string="Barcode", oldname="ean13")
```
legalsylvain, line 14: Also this one.

legalsylvain (changes requested):
No take a look on the two lines. 
Also changing the report has nothing to do with this module.

barcodes_generator_partner/views/view_res_partner.xml
agent-z28: So please explain howto Add this functionality to Partner without installing point of sale👌

rousseldenis: @legalsylvain IMHO @agenterp remark is correct. The dependency on point of sale should be avoided

legalsylvain: > The dependency on point of sale should be avoided

I'm fully agree ! But it is not a matter of OCA, but a matter of Odoo, and should be fixed in the core. For the time being, barcode field is defined in the module ``point_of_sale``.
https://github.com/odoo/odoo/blob/master/addons/point_of_sale/models/res_partner.py#L16 (*)

This PR is breaking Odoo partner form. Step to reproduce : 
- install this module barcodes_generator_abstract + point_of_sale : 
- Go to partner form 
- The partner form has two field barcodes.

image

**Note :** 
For V12, the barcode field is in base module ([in the model](https://github.com/odoo/odoo/blob/12.0/odoo/addons/base/models/res_partner.py#L172)) but was in the point_of_sale module (for the view). This point was fixed by https://github.com/odoo/odoo/pull/25070/files 
- For V13, and master, this move has been removed, and the field is again in point_of_sale module. 

@mart-e : Do you know why ``barcode`` partner field has moved into base in V12, and has moved again in ``point_of_sale`` for V13 and master ? 

Thanks !

mart-e:
>  Do you know why barcode partner field has moved into base in V12, and has moved again in point_of_sale for V13 and master ?

No sorry. I just discovered that, it was done at odoo/odoo#28316

rousseldenis:
> > Do you know why barcode partner field has moved into base in V12, and has moved again in point_of_sale for V13 and master ?
> 
> No sorry. I just discovered that, it was done at [odoo/odoo#28316](https://github.com/odoo/odoo/pull/28316)

Due to @switch87 remark : https://github.com/odoo/odoo/pull/28316#discussion_r246750020 ...

That's not because something is not used anywhere than in pos, it should be moved there...

barcodes_generator_partner/views/view_res_partner.xml
agent-z28: @legalsylvain i agree that it's not nice. but we cannot rely on them to fix it.
so how can we provide a barcode to partner without installing pos?

rousseldenis: @agenterp The solution I've already done in another place is to get this with possible duplicate (don't add dependency here).

And in pos repository add a module 'auto-install' with this module + pos in dependencies to remove the duplicate in view. That's not ideal, but we need to make it change from master in odoo (I commented the PR)

legalsylvain: Well, this is not acceptable in a stable revision for all instances that are using this module in production, this patch will break form views and the patch will not be installed. (if pos repository is not present, is not updated, if installation is done by pip, etc...)

- I propose to close this PR, and work with @mart-e for master revision. (maybe it will not be necessary)
- If interested, please use this branch as a patch in production.
- for the report part, I think that it should be introduced in a module ``barcodes_partner_report`` and not in this one, because it has nothing to do with the generation of the barcode.

regards.

legalsylvain:
I found the PR I did to move barcode from PoS to base module.

https://github.com/odoo/odoo/pull/25070

@mart-e : how can we move forward ? Do you want a PR again ? if yes, on which branch(es) ?

mart-e:
@legalsylvain I see you have found odoo/odoo#41658

PR #240, opened 2020-01-08
[12.0][IMP] product_multi_ean

JoanSForgeFlow:
This is a forward port of #208.
Also improves tests to avoid conflicts when testing a database with stock installed.

cc ~ @ForgeFlow @LoisRForgeFlow


HviorForgeFlow (approved):

LoisRForgeFlow (approved):
:+1:

PR #243, opened 2020-01-28
[FIX] stock_barcodes: out picking location workflow

chienandalu:
When we've got an out picking, we're the logic workflow is that the scanned locations
is the location we're getting the stock from

cc @Tecnativa TT21753

PR #245, opened 2020-01-30
[FIX] stock_barcodes: out picking location workflow

chienandalu:
When we've got an out picking, the logical workflow is that the scanned location
is the location we're getting the stock from

cc @Tecnativa TT21753

FW Port de https://github.com/OCA/stock-logistics-barcode/pull/243


sergio-teruel (approved):

pedrobaeza (approved):
Fast-tracking as merged in previous version. Linter has been fixed in main branch.

PR #247, opened 2020-01-31
[FIX] stock_barcodes: location_dest_id

chienandalu:
cc @Tecnativa TT21018

PR #248, opened 2020-01-31
[12.0][FIX] stock_barcodes: location_dest_id

chienandalu:
cc @Tecnativa TT21018

FW Port of https://github.com/OCA/stock-logistics-barcode/pull/247


pedrobaeza:
Fast-tracking as well:

/ocabot merge patch

PR #255, opened 2020-02-18
[FIX] stock_barcodes: out pickings default location

chienandalu:
In outgoing pickings setting the default location as the picking
destination one is not correct, as the resulting move line would be from
destination to destination. This patch fix it considering that when we
receive stock we want to set to wich location it will go and when send
stock we want to set from wich location we get it from.

cc @Tecnativa TT21753


pedrobaeza (approved):

sergio-teruel (approved):

sergio-teruel:
Applyy to 12 and 13

PR #258, opened 2020-03-09
[12.0] FIX barcodes_generator_abstract: when installed using pip, wrong 'barcode' library is installed

eLBati:
Steps:

 - Install odoo using pip
 - install `python-barcode` as required by `barcodes_generator_abstract`
 - install `odoo12-addon-barcodes-generator-abstract`

Result:
`python-barcode` and `barcode` libraries are both installed, causing `barcodes_generator_abstract` to fail with `AttributeError: module 'barcode' has no attribute 'get_barcode_class'`

### DIFF
--- a/setup/barcodes_generator_abstract/setup.py
+++ b/setup/barcodes_generator_abstract/setup.py
@@ -2,5 +2,11 @@ import setuptools
 
 setuptools.setup(
     setup_requires=['setuptools-odoo'],
-    odoo_addon=True,
+    odoo_addon={
+        'external_dependencies_override': {
+            'python': {
+                'barcode': 'python-barcode',
+            },
+        },
+    },
 )


eLBati:
@sbidoul do you know better solutions instead of the one proposed here?
Thanks

sbidoul:
@eLBati `external_dependencies_override` is probably what you are looking for? https://github.com/acsone/setuptools-odoo#controlling-setuptools-odoo-behaviour

Note in Odoo 13 one can finally put the real distribution name in external-dependencies.

legalsylvain (changes requested):
Hi @eLBati. I don't understand why this patch can fix anything. this line doesn't force odoo to install a lib, it only checks if a lib is importable.
I think that we should : 
- keep this line, because otherwise, the module can be installed without this library.
- fixing the installation, made by pip.

eLBati:
@sbidoul yes, thanks, I forgot that one.

@legalsylvain updated PR

legalsylvain (approved):
Thanks !

PR #275, opened 2020-05-15
[12.0][MIG] stock_scanner_web

eantones:
Migration to 12.0

Starting from https://github.com/OCA/stock-logistics-barcode/pull/95

@angelmoya @LoisRForgeFlow @realsaiko @lmignon  @MiquelRForgeFlow  @rousseldenis   can you review it please? 

Thanks!
:-)

PR #276, opened 2020-05-15
[12.0][MIG] stock_scanner_web_location_info

eantones:
Migration to 12.0

Starting from https://github.com/OCA/stock-logistics-barcode/pull/95

Related to https://github.com/OCA/stock-logistics-barcode/pull/275

@angelmoya @LoisRForgeFlow @realsaiko @rousseldenis can you review it please? 

Thanks!
:-)


eantones:
How can I avoid the Runbot dependency error if the parent module is not yet merged? I know that I might have both on the same PR but I'd prefer a cleaner solution.
Thanks in advance

rousseldenis:
> How can I avoid the Runbot dependency error if the parent module is not yet merged? I know that I might have both on the same PR but I'd prefer a cleaner solution.
> Thanks in advance

Hi! Thanks for this. You can add a line with the repo name plus the github link to your branch in the oca_dependencies.txt (to be removed before merge).

e.g.: stock-logistics-barcode https://github.com/nuobit/stock-logistics-barcode 12.0-mig-stock_scanner_web

I think there is a more cleaner option is to use test-requirements.txt (but it has to be enabled on the repo)

pedrobaeza:
I'm afraid that doesn't serve when the repository of the dependency is the same.

eantones:
Ok, so in this case, is it better to put both modules on the same PR?

rousseldenis:
> Ok, so in this case, is it better to put both modules on the same PR?

Or basing your branch on the other one.

pedrobaeza:
You can put this one as WIP and rebase when the other is merged. Several modules in one PR is not advisable and can make the review more difficult.

eantones:
I think I did not understand, if I rebase this branch on the parent one then this PR will have also all the commits from the parent and it will contain two modules which is what we are trying to avoid.
What am I missing?

eantones:
@rousseldenis I rebased this as well, in case you need it :-)

PR #284, opened 2020-06-30
[13.0][ADD] stock_barcodes_zpl2

MiquelRForgeFlow:
- [x] Depends on https://github.com/OCA/report-print-send/pull/210.


AaronHForgeFlow:
Fail to install the module:

```
odoo.tools.convert.ParseError: "<class 'ValueError'>: "External ID not found in the system: product.model_stock_inventory" while evaluating
"ref('product.model_stock_inventory')"" while parsing /home/odoo/build/OCA/stock-logistics-barcode/stock_barcodes_zpl2/data/actions.xml:6, near
<odoo noupdate="1">
    <function model="printing.label.zpl2" name="add_action">
        <value eval="ref('stock.model_stock_picking')"/>
    </function>
    <function model="printing.label.zpl2" name="add_action">
        <value eval="ref('product.model_stock_inventory')"/>
    </function>
    <function model="printing.label.zpl2" name="add_action">
        <value eval="ref('product.model_product_product')"/>
    </function>
    <function model="printing.label.zpl2" name="add_action">
        <value eval="ref('product.model_product_template')"/>
    </function>
</odoo>
```

AaronHForgeFlow:
I am having errors in printer_zpl2 module when printing a QR with autofill:

```
  File "/home/odoo/custom/src/report-print-send/printer_zpl2/models/printing_label_zpl2.py", line 154, in _get_to_data_to_print
    data = self._get_component_data(record, component, eval_args)
  File "/home/odoo/custom/src/report-print-send/printer_zpl2/models/printing_label_zpl2.py", line 131, in _get_component_data
    return safe_eval(data, eval_args) or ""
  File "/home/odoo/custom/src/odoo/odoo/tools/safe_eval.py", line 349, in safe_eval
    c = test_expr(expr, _SAFE_OPCODES, mode=mode)
  File "/home/odoo/custom/src/odoo/odoo/tools/safe_eval.py", line 213, in test_expr
    pycompat.reraise(ValueError, ValueError('"%s" while compiling\n%r' % (ustr(e), expr)), exc_info[2])
  File "/home/odoo/custom/src/odoo/odoo/tools/pycompat.py", line 13, in reraise
    raise value.with_traceback(tb)
  File "/home/odoo/custom/src/odoo/odoo/tools/safe_eval.py", line 207, in test_expr
    expr = expr.strip()
ValueError: "'dict' object has no attribute 'strip'" while compiling
{'product_barcode': 'RAS-1', 'lot_barcode': False, 'uom_id': 1, 'package_barcode': False, 'product_qty': 0.0}
```

AaronHForgeFlow:
If I do `return safe_eval(str(data), eval_args) or ""` then it works the auto-fill option but not the normal use case.

MiquelRForgeFlow:
All things solved. The normal case, is used in the models with lines, like stock.picking, remember that the object now are the wizard lines instead of the model, thus they won't have object.name

AaronHForgeFlow:
I don't think it is solved. It fails with object.product_id.display_name or just putting some quoted text "xxxx",  probably it was failing before the supersede of printer_zpl2 but I did not realized of that.

AaronHForgeFlow (approved):
I was wrong @MiquelRForgeFlow . I tested in a blank newly craeted database and this worked as expected. So :+1:  for me.

rousseldenis:
@MiquelRForgeFlow

MiquelRForgeFlow:
@rousseldenis ready

stock_barcodes_zpl2/wizard/print_record_label.py
```diff
@@ -0,0 +1,131 @@
+# Copyright 2020 ForgeFlow S.L. (<http://www.forgeflow.com>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+_supported_models = [
```
rousseldenis, line 6: @MiquelRForgeFlow Shouldn't those kind of things go in the base module ???

MiquelRForgeFlow: Which base module? It is put in the readme description, where it matters.

rousseldenis: Ok, I understand. This model is in printer_zpl2.

stock_barcodes_zpl2/wizard/print_record_label.py
```diff
@@ -0,0 +1,131 @@
...
+    "stock.inventory",
+    "product.product",
+    "product.template",
+]
+
+
+class PrintRecordLabel(models.TransientModel):
+    _inherit = "wizard.print.record.label"
+
+    model = fields.Char(related="active_model_id.model", string="Model Name")
```
rousseldenis: Mmmh, IMHO, those fields are too abstract to be in a stock module, no ?

MiquelRForgeFlow: Relatively. Is it a problem? :)

rousseldenis: That depends. You supposed nobody want to extend printer_zpl2 for another usage :smiley:

MiquelRForgeFlow: - [x] https://github.com/OCA/report-print-send/pull/286.

rousseldenis: Great. Maybe you can add a Selection field for supported documents there too (for the show_lines)?

MiquelRForgeFlow: Where? How? I don't get you now :S

PR #287, opened 2020-07-14
[12.0] FIX stock_barcodes_supplierinfo: wrong result when supplier barcode is not found, return only when product is found

eLBati:
Also Adding tests


rousseldenis:
@eLBati Could you rebase ?

PR #293, opened 2020-08-25
[FIX] stock_barcodes: tests

chienandalu:
- The system user name is `System` by default and `OdooBot` when
`mail_bot` is installed so the resulting string could change
- `python-barcode` requires python 3.6 from version 0.13.0 so until Odoo
oficially supports it we'll have to stick to this version.

cc @Tecnativa TT25264

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,1 +1,1 @@
-python-barcode
+python-barcode == 0.12.0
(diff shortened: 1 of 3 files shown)


pedrobaeza:
Please cherry-pick this from v13 instead:

dab52c9edbd712c7c1079dbef4680c2bf2bfa81a
14f5980fd8d4aaf9b65f9e4a344fb391849f8d9e

chienandalu:
Ok, done :+1:

pedrobaeza:
Please check Travis

chienandalu:
@pedrobaeza fixed

PR #307, opened 2020-11-10
[ADD] [10.0] barcodes_regex_group (use regex groups in nomenclatures)

thomaspaulb:
We can use regex groups to specify which part of the barcode we want to match from a full string.


pedrobaeza:
Remember to not use plurals in module names.

thomaspaulb:
@pedrobaeza I think for `barcodes` that's a passed station, but I corrected the `groups`.

pedrobaeza:
Yes, `barcodes` should be kept as the name is core, but not `group`. See https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#modules for more info.

PR #313, opened 2020-12-14
[MIG] barcodes_generator_abstract: Migration to 14.0

chafique-delli:


barcodes_generator_abstract/models/barcode_generate_mixin.py
```diff
@@ -32,8 +30,6 @@ class BarcodeGenerateMixin(models.AbstractModel):
 
     generate_type = fields.Selection(
         string="Generate Type",
-        selection=_GENERATE_TYPE,
-        readonly=True,
```
legalsylvain: I don't understand that move. could you explain why you remove that two lines ?

alexis-via: On recent Odoo versions, releated fields are readonly by default, so no need to write it explicitely. And selection= is not needed on a related field, it's automatic.

hparfr (changes requested):
Tests written in v13 are missing.

alexis-via:
I won't contribute on this PR any more because I now use my module product_print_zpl_barcode to generate product barcodes (cf https://github.com/akretion/odoo-usability/blob/14.0/product_print_zpl_barcode/models/product.py#L49). Maybe @chafique-delli will continue to work on this PR.

Zar21:
I test your module and the barcode dependency is wrong since this module is using the get_barcode_class function and this function depends on the python library python-barcode, so you should change this.
If you couldn't change this, I can do it.

Zar21:
Superseeded by https://github.com/OCA/stock-logistics-barcode/pull/340

hugho-ad:
@pedrobaeza 

could you close this pls 

superbased by https://github.com/OCA/stock-logistics-barcode/pull/340

PR #315, opened 2021-01-12
[13.0] [FIX] stock_picking_product_barcode_report: Avoid empty data on wizard

CarlosRoca13:
cc @Tecnativa TT27703

With next change we avoid the problem shown in the picture.

image

### DIFF
--- a/stock_picking_product_barcode_report/wizard/stock_barcode_selection_printing.py
+++ b/stock_picking_product_barcode_report/wizard/stock_barcode_selection_printing.py
@@ -60,7 +60,7 @@ class WizStockBarcodeSelectionPrinting(models.TransientModel):
         ].default_get(line_fields)
         for picking_id in self.picking_ids.ids:
             picking = self.env["stock.picking"].browse(picking_id)
-            for move_line in picking.move_line_ids:
+            for move_line in picking.move_line_ids.filtered("product_id.barcode"):
                 product_print_moves_data = dict(product_print_moves_data_tmpl)
                 product_print_moves_data.update(
                     self._prepare_data_from_move_line(move_line)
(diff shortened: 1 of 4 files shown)


pedrobaeza:
I would add a `<p>` in the view saying that here only appears products with barcode defined.

CarlosRoca13:
Done @pedrobaeza

stock_picking_product_barcode_report/wizard/stock_barcode_selection_printing_view.xml
```diff
@@ -11,6 +11,7 @@
                     <strong>Barcode Format</strong>
                 </span>
                 <field name="barcode_format" />
+                <p>Here only appears products with barcode defined</p>
```
pedrobaeza:
```suggestion
                <p>It only appears here products with barcode defined</p>
```

stock_picking_product_barcode_report/wizard/stock_barcode_selection_printing.py
```diff
@@ -60,7 +60,7 @@ def _onchange_picking_ids(self):
         ].default_get(line_fields)
         for picking_id in self.picking_ids.ids:
             picking = self.env["stock.picking"].browse(picking_id)
-            for move_line in picking.move_line_ids:
+            for move_line in picking.move_line_ids.filtered("product_id.barcode"):
```
sergio-teruel, line 63: @pedrobaeza What is better for performance, filter records or a continue inside the bucle?

pedrobaeza: `continue`, but sometimes it doesn't worth the less readable code for performance.

pedrobaeza (approved):

carlosdauden (approved):

stock_picking_product_barcode_report/wizard/stock_barcode_selection_printing.py
sergio-teruel: Oks, thanks

sergio-teruel (approved):

PR #319, opened 2021-02-06
[13.0][ADD] stock_barcodes_gs1_secondary_unit: New module

sergio-teruel:
cc @Tecnativa
New modulo to store barcode in secondary units and use it with barcode interface


sergio-teruel:
Continue in https://github.com/OCA/stock-logistics-barcode/pull/321

PR #325, opened 2021-03-22
[13.0][MIG] barcodes_generator_location: Migration to 13.0

ernestotejeda:
Cc @Tecnativa TT27167


barcodes_generator_location/__manifest__.py
```diff
@@ -2,13 +2,13 @@
...
+    "version": "13.0.1.0.0",
     "category": "Tools",
-    "author": "LasLabs, " "Odoo Community Association (OCA)",
-    "website": "https://laslabs.com",
+    "author": "LasLabs, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/stock-logistics-barcode",
     "license": "AGPL-3",
-    "depends": ["barcodes_generator_abstract", "stock",],
-    "data": ["views/stock_location.xml",],
+    "depends": ["barcodes_generator_abstract", "stock", "stock_barcodes"],
```
pedrobaeza: This dependency shouldn't be used. It's too much for this.

ernestotejeda: @pedrobaeza  take a look at the changes I made regarding this, please.

chienandalu: Would it be more appropiate to pull an extra `stock_location_barcode` module which both this and `stock_barcodes` depends on to keep this stuff simple?

pedrobaeza: I don't think we should add such module complexity, and more not having control over enterprise module `stock_barcode`, so better to inject it via code if missing.

victoralmau:
It's ready to review because `stock_barcodes` dependency have removed, right?

pedrobaeza:
Please finish it, whatever it takes.

victoralmau (approved):

pedrobaeza:
@chienandalu is it OK for you?

chienandalu:
I'll do the final functional review

chienandalu (approved):
Works as expected :+1:

PR #327, opened 2021-03-25
[13.0][ADD] stock_quant_barcode: New module

sergio-teruel:
cc @Tecnativa TT28846
ping @chienandalu @carlosdauden

PR #329, opened 2021-03-29
[13.0][WIP][ADD] stock_barcodes_quant: New module

sergio-teruel:
cc @Tecnativa TT28870
- [ ] Depends on: https://github.com/OCA/stock-logistics-barcode/pull/327


sergio-teruel:
Included in #369 and #348

PR #338, opened 2021-05-18
[12.0][IMP] stock_barcodes: add html field to track pending products

albariera:
:sparkles: This PR adds a table at the stock_barcodes_read_picking wizard to track pending products in an easier way.
I tried to do it first with an onchange method, but the quantity_done view was not refreshed. To overcome this, I have simulated a table with an html field that does not have this problem. 
If someone comes up with a better solution, it will be welcome. :slightly_smiling_face:

### DIFF
--- a/stock_barcodes/templates/missing_moves_template.xml
+++ b/stock_barcodes/templates/missing_moves_template.xml
@@ -1,0 +1,37 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<odoo>
+    <template id="missing_moves">
+        <div class="table-responsive" t-if="moves">
+        <table class='o_list_view table table-sm table-hover table-striped o_list_view_ungrouped'>
+            <thead>
+            <tr>
+                <th class="o_column_sortable">Product</th>
+                <th class="o_column_sortable">Initial Demand</th>
+                <th class="o_column_sortable">Done</th>
+            </tr>
+            </thead>
+            <tr t-foreach="moves" t-as="move">
+                <td>
+                    <span t-field="move.product_id"/>
+                </td>
+                <td>
+                    <span t-field="move.product_uom_qty"/>
+                </td>
+                <td>
+                    <span t-field="move.quantity_done"/>
+                </td>
+            </tr>
+            <tfoot class="oe_foot">
+                <tr>
+                    <td></td>
+                    <td></td>
+                    <td></td>
+                </tr>
+            </tfoot>
+        </table>
+        </div>
+        <div t-else="">
+            <h3>No pending operations</h3>
+        </div>
+    </template>
+</odoo>
--- a/stock_barcodes/templates/missing_moves_template.xml
+++ b/stock_barcodes/templates/missing_moves_template.xml
@@ -23,9 +23,7 @@
             </tr>
             <tfoot class="oe_foot">
                 <tr>
-                    <td></td>
-                    <td></td>
-                    <td></td>
+                    <td colspan="3"/>
                 </tr>
             </tfoot>
         </table>
--- a/stock_barcodes/wizard/stock_barcodes_read_picking.py
+++ b/stock_barcodes/wizard/stock_barcodes_read_picking.py
@@ -22,6 +22,7 @@ class WizStockBarcodesReadPicking(models.TransientModel):
     )
     pending_moves = fields.Html(
         compute="_compute_pending_move",
+        groups="stock_barcodes.group_track_pending_products_picking_barcode"
     )
     candidate_picking_ids = fields.One2many(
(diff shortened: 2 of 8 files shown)


stock_barcodes/static/src/css/stock.css
```diff
@@ -6,3 +6,5 @@
     background: none !important;
     box-shadow: none !important;
 }
+
+
```
MeetKD: You may remove this extra lines.

MeetKD (commented):
Minor change!

stock_barcodes/static/src/css/stock.css
albariera: Comment attended. Thank you!

etobella (approved):
LGTM

stock_barcodes/templates/missing_moves_template.xml
```diff
@@ -0,0 +1,34 @@
...
+                </td>
+                <td>
+                    <span t-field="move.quantity_done"/>
+                </td>
+            </tr>
+            <tfoot class="oe_foot">
+                <tr>
+                    <td></td>
+                    <td></td>
+                    <td></td>
```
MeetKD: Do we need this empty elements?

albariera: I did it to simulate the odoo table as much as possible. If I am not wrong, using the web inspector I saw that odoo uses this method to simulate the gray line at the bottom of the table.

MeetKD: I am not able to open runbot somehow. Can you please try to remove the empty elements and check if that works ? @pedrobaeza @sergio-teruel Can you please share your inputs here ?

stock_barcodes/wizard/stock_barcodes_read_picking.py
```diff
@@ -40,6 +43,31 @@ class WizStockBarcodesReadPicking(models.TransientModel):
...
+                            "picking": record.picking_id,
+                            "moves": moves
+                        },
+                    )
+                else:
+                    text = self.env["ir.qweb"].render(
+                        "stock_barcodes.no_missing_moves", {
+                            "picking": record.picking_id,
+                        },
+                    )
```
MeetKD: I think we can also improve code here and keep only one template **stock_barcodes.missing_moves** and at template level we can check if moves are there show the table else show the the message **No pending operation** ? What do your think ? @albariera

stock_barcodes/templates/missing_moves_template.xml
albariera: On Odoo, it is done in a similar way, but on js, as you can see here:

https://github.com/odoo/odoo/blob/12.0/addons/web/static/src/js/views/list/list_renderer.js#L394

stock_barcodes/wizard/stock_barcodes_read_picking.py
albariera: You are right. I have already changed the code! Thank you for the review :smiley:

sergio-teruel:
@albariera Can you add an option in res.config.settings to activate this behaviour ??

albariera:
@sergio-teruel Done! :smile:

stock_barcodes/templates/missing_moves_template.xml
sergio-teruel: You can use `<td colspan="3"/>`

stock_barcodes/wizard/stock_barcodes_read_picking.py
```diff
@@ -40,6 +43,25 @@ class WizStockBarcodesReadPicking(models.TransientModel):
         string='Confirmed moves',
     )
 
+    @api.depends(
+        "picking_id", "barcode", "picking_id.move_lines.move_line_ids.qty_done"
+    )
+    def _compute_pending_move(self):
+        for record in self:
```
sergio-teruel, line 51: Compute field only if user has the security group for performance.

albariera: Done :+1:

stock_barcodes/templates/missing_moves_template.xml
albariera: :+1:

sergio-teruel:
@albariera Thanks!!

sergio-teruel (approved):

pedrobaeza:
/ocabot merge minor

Please move it to v13

sergio-teruel:
@albariera can you add for v13 in this PR?
* https://github.com/OCA/stock-logistics-barcode/pull/348

PR #309, opened 2020-11-13
[14.0] mig product multi ean

sebastienbeau:
Migrate from version 12.0

PR #341, opened 2021-05-26
[WIP][13.0] migrate and improve product_multi_ean

jjscarafia:
Migrate product_multi_ean to v13
Improve:

- no more size (port of https://github.com/OCA/stock-logistics-barcode/pull/285/files)
- allow to setup barcodes from product.template when no variants

Further tests still needed.
I didn't find a nicer implementation to allow working from product.template and product.product


pedrobaeza:
Maybe we should call it product_multi_barcode.

jjscarafia:
I agree that product_multi_ean may be better, but I believe we could do that change in v14 pr (https://github.com/OCA/stock-logistics-barcode/pull/309) as already suggested.
What do you think?

pedrobaeza:
If this is a migration, why not do it now?

jjscarafia:
So that PR https://github.com/OCA/stock-logistics-barcode/pull/309 don't need to start again
But no problem for me to do it on v13

I will ping on the other PR

jjscarafia:
Close in favor of https://github.com/OCA/stock-logistics-barcode/pull/345

Issue #342, opened 2021-06-02
[RFC] GS1 Barcodes


etobella:
During the configuration of the barcodes, we have found the following issue:

* A supplier creates a Product, we can call it A
* As they are delivering on boxes of 100, they do create a Barcode for the box, and they use it with code 02 (Product Barcode)
* When you receive the box, you will open it and use it by boxes

Has anyone found this situation? Any ideas how to solve it?

Thanks for your effort

PR #343, opened 2021-06-02
[IMP] stock_barcodes

etobella:
This PR tries to make the following changes:

* Use the Packaging unit if defined, not the Purchase
* Multiply the units if package is defined on GS1 (it was replaced)
* Try to find the Product code as Package code (#342)
* Returns the stripped value, this is important when scanned Code128 codes, as sometimes this kind of characters is added
* Puts the proper destination location using putaway strategies
* Puts the proper origin location using quants

@LoisRForgeFlow

### DIFF
--- a/base_gs1_barcode/models/gs1_barcode.py
+++ b/base_gs1_barcode/models/gs1_barcode.py
@@ -161,7 +161,7 @@ class GS1Barcode(models.Model):
                             ai, position))
 
                 position += len(groups['value'])
-                results[ai] = groups['value'].replace(separator, '')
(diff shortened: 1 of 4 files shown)


LoisRForgeFlow (approved):
I don't see any inconvenient of this change, so LGTM

pedrobaeza:
@sergio-teruel it seems there are several bugs fixed here. Can you check?

etobella:
@sergio-teruel Can you review it please? :pray:

sergio-teruel:
@etobella Can we do a meet?

sergio-teruel (approved):

PR #345, opened 2021-06-15
[13.0] migrate and improve product_multi_ean (renamed)

jjscarafia:
Continue with https://github.com/OCA/stock-logistics-barcode/pull/341 but rename module to product_multi_barcode (by making a cherry-pick of https://github.com/OCA/stock-logistics-barcode/pull/309/


ValentinVinagre:
Hi @jjscarafia ,
What is the status of this pr? Is it already finished and need reviews?
Thanks,

product_multi_barcode/views/product_template_view.xml
```diff
@@ -0,0 +1,25 @@
...
+        <field name="priority">16</field>
+        <field name="inherit_id" ref="product.product_template_only_form_view" />
+        <field name="arch" type="xml">
+            <field name="barcode" position="attributes">
+                <attribute name="readonly">True</attribute>
+            </field>
+            <field name="barcode" position="after">
+                <field
+                    name="barcode_ids"
+                    attrs="{'invisible': [('product_variant_count', '>', 1)]}"
```
ivantodorovich, line 15:
```suggestion
```
Why product_variant_count > 1?
Wouldn't we want it in single variant templates too?

product_multi_barcode/models/product_barcode.py
```diff
@@ -0,0 +1,54 @@
...
+    def _compute_product_tmpl(self):
+        for rec in self.filtered(lambda x: not x.product_tmpl_id and x.product_id):
+            rec.product_tmpl_id = rec.product_id.product_tmpl_id
+
+    @api.depends("product_tmpl_id.product_variant_ids")
+    def _compute_product(self):
+        for rec in self.filtered(
+            lambda x: not x.product_id and x.product_tmpl_id.product_variant_ids
+        ):
+            rec.product_id = rec.product_tmpl_id.product_variant_ids[0]
```
ivantodorovich, line 42:
```suggestion
    product_id = fields.Many2one(
        "product.product",
        string="Product",
        required=True,
    )
    product_tmpl_id = fields.Many2one(
        related="product_id.product_tmpl_id",
        store=True,
    )
```
I take it's to be able to write on `product_tmpl_id` (and update `product_id` accordingly), but what's the point? Barcodes should be linked to a `product.product` anyway.

Without this, creating barcodes from product.template wouldn't work as it is, but.. -> (see the other comment below)

product_multi_barcode/views/product_template_view.xml
```diff
@@ -0,0 +1,25 @@
...
+                >
+                    <tree string="Barcodes" editable="bottom">
+                        <field name="sequence" widget="handle" />
+                        <field name="name" />
+                    </tree>
+                </field>
```
ivantodorovich, line 21: (continuing from previous comment)

Issues I see with this:
If a template has multiple variants, you wouldn't know to which variant the barcode belongs to.
If the template has only 1 variant, having a product_id here is useless, so..

I would try to find a way to:
- Display barcode's `product_id` if `product_variant_count > 1`
- Otherwise, set a default value for barcode's `product_id` to be the template's only variant and hide the field

So, maybe something like this (haven't tested it):

```suggestion
                <field
                    name="barcode_ids"
                    attrs="{'invisible': [('product_variant_count', '&gt;', 1)]}"
                >
                    <tree string="Barcodes" editable="bottom">
                        <field name="sequence" widget="handle" />
                        <field name="product_id"/>
                        <field name="name" />
                    </tree>
                </field>
                <field
                    name="barcode_ids"
                    attrs="{'invisible': [('product_variant_count', '&lt;=', 1)]}"
                    context="{'default_product_id': product_variant_id}"
                >
                    <tree string="Barcodes" editable="bottom">
                        <field name="sequence" widget="handle" />
                        <field name="product_id" invisible="1"/>
                        <field name="name" />
                    </tree>
                </field>
```



If it doesn't work maybe see how odoo's doing it for seller_ids:

https://github.com/odoo/odoo/blob/ee38f8e393e937c5694e6b75cd05b8a35d4f1624/addons/purchase/views/product_views.xml#L14-L17

ivantodorovich (changes requested):

jjscarafia:
> Hi @jjscarafia ,
> What is the status of this pr? Is it already finished and need reviews?
> Thanks,

Yes, I believe so. Ready for review. Thanks

product_multi_barcode/views/product_template_view.xml
jjscarafia: mmm, we only show barcode_ids if only single variant, I've used same criteria of odoo  with [barcode field](https://github.com/odoo/odoo/blob/13.0/addons/product/views/product_template_views.xml#L39)

(it's invisible when more than one variant)

product_multi_barcode/views/product_template_view.xml
jjscarafia: mmm, I get the idea but:
1. I'm not convinced yet that it's better on terms of usability
2. I'm not going to be able to play with the new approach. The change we did  till now was only to improve usability (with some hacks) but keeping the module behavior (if I'm not wrong).

If anyone interested on continuing the work, go ahead!

Thanks!

ivantodorovich (approved):

kevinkhao (approved):
Code review LGTM

PR #351, opened 2021-07-06
[13.0][MIG] product_gtin: Migration to 13.0

victoralmau:
Migration to 13.0

Please @joao-p-marques and @pedrobaeza can you review it?

@Tecnativa TT30724


product_gtin/models/abstract_barcodenumber.py
```diff
@@ -0,0 +1,58 @@
...
+# Copyright 2016 ACSONE SA/NV (<http://acsone.eu>)
+# Copyright 2021 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+import barcodenumber
+from openerp import _, api, models
+from openerp.exceptions import ValidationError
+
+
+class AbstractBarcodenumber(models.AbstractModel):
+    _name = "abstract.barcodenumber"
```
pedrobaeza: Use conventions:

```suggestion
    _name = "barcode.gtin_check.mixin"
```

product_gtin/models/abstract_barcodenumber.py
```diff
@@ -0,0 +1,58 @@
+# Copyright 2004-2011 Tiny SPRL (<http://tiny.be>)
+# Copyright 2010-2011 Camptocamp Austria (<http://www.camptocamp.at>)
+# Copyright 2016 ACSONE SA/NV (<http://acsone.eu>)
+# Copyright 2021 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+import barcodenumber
+from openerp import _, api, models
```
pedrobaeza: Change this to `odoo`

product_gtin/models/product_product.py
```diff
@@ -0,0 +1,17 @@
...
+from openerp import api, models
+
+
+class ProductProduct(models.Model):
+    _name = "product.product"
+    _inherit = ["abstract.barcodenumber", "product.product"]
+    _barcodes = ["barcode"]
+
+    @api.constrains("barcode")
+    def _check_barecode(self):
```
pedrobaeza:
```suggestion
    def _check_barcode(self):
```

product_gtin/models/product_product.py
```diff
@@ -0,0 +1,17 @@
...
+# Copyright 2016 ACSONE SA/NV (<http://acsone.eu>)
+# Copyright 2021 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openerp import api, models
+
+
+class ProductProduct(models.Model):
+    _name = "product.product"
+    _inherit = ["abstract.barcodenumber", "product.product"]
+    _barcodes = ["barcode"]
```
pedrobaeza: This is not serving for anything, but we can give it an utility:

- Move it to the mixin
- Put it a proper variable name:  `_barcode_fields = ["barcode"]` (or if it gets more complicated, do it a single field)
- Move the check to the mixin, and see if you can put on the `@api.constrains` the variable as arguments.
- On the inherited model, you only need to change the variable if not matching the field name in the model.

product_gtin/views/product_view.xml
```diff
@@ -0,0 +1,29 @@
...
+<openerp>
+    <data>
+        <record id="product_template_search_view" model="ir.ui.view">
+            <field name="model">product.template</field>
+            <field name="inherit_id" ref="product.product_template_search_view" />
+            <field name="arch" type="xml">
+                <field name="name" position="attributes">
+                    <attribute name="filter_domain">['|', '|',
+                        ('default_code','ilike',self),('name','ilike',self),
+                        ('barcode','ilike',self)]
```
pedrobaeza: Isn't barcode already searchable in standard?

pedrobaeza: Anyways, this is breaking any other inheritance.

pedrobaeza (changes requested):

product_gtin/views/product_view.xml
victoralmau: It's not really necessary.

victoralmau:
Thanks @pedrobaeza changes done.
Any idea to prevent incompatibility with other addons like `stock_barcodes_gs1`?. Maybe we need to add custom context `skip_barcode_gtin_check=True` in stock_barcodes_gs1 tests.

pedrobaeza:
I think no error should be arisen if the length is different from the GTIN expected ones.

product_gtin/models/product_product.py
```diff
@@ -0,0 +1,11 @@
+# Copyright 2004-2011 Tiny SPRL (<http://tiny.be>)
+# Copyright 2010-2011 Camptocamp Austria (<http://www.camptocamp.at>)
+# Copyright 2016 ACSONE SA/NV (<http://acsone.eu>)
+# Copyright 2021 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openerp import models
```
joao-p-marques:
```suggestion
from odoo import models
```

product_gtin/tests/test_product_gtin_functions.py
```diff
@@ -0,0 +1,123 @@
...
+
+class TestProductGtin(common.TransactionCase):
+    def setUp(self):
+        super().setUp()
+        self.barcode_mixin = self.env["barcode.gtin_check.mixin"]
+        self.product = self.env["product.product"].create({"name": "Test product"})
+
+    # The codes have been tested against
+    # http://www.hipaaspace.com/Medical_Data_Validation/Universal_Product_Code/
+    # UPC_Validation.aspx noqa
```
joao-p-marques: Why this `noqa` comment? Is it doing something important?
If so, could you explain so we can see if it worth fixing instead?

Also, the rest of the comments would probably fit better as a docstring of this method...

product_gtin/tests/test_product_gtin_functions.py
```diff
@@ -0,0 +1,159 @@
...
+        for code in VALID_UPC_CODES:
+            self.assertFalse(self.barcode_mixin._check_ean13(code))
+
+    def test_DICT_CHECK_CODE_TYPE(self):
+        """Check if the dict _DICT_CHECK_CODE_TYPE exists."""
+        self.assertTrue(self.barcode_mixin._DICT_CHECK_CODE_TYPE)
+
+    def test_DICT_CHECK_FUNCTIONS(self):
+        """Check if the dict _DICT_CHECK_FUNCTIONS exists."""
+        self.assertTrue(self.barcode_mixin._DICT_CHECK_FUNCTIONS)
```
joao-p-marques: You could probably concentrate some of these tests in a singe one, without the need for a function name so specific...
```suggestion
    def test_dicts_check_exist(self):
        """Check if the dicts _DICT_CHECK_CODE_TYPE and _DICT_CHECK_FUNCTIONS exist."""
        self.assertTrue(self.barcode_mixin._DICT_CHECK_CODE_TYPE)
        self.assertTrue(self.barcode_mixin._DICT_CHECK_FUNCTIONS)
```

The same applies to some of the other functions that just have 1 or 2 lines, if you think it is appropriate.

joao-p-marques (changes requested):
Code review. Looks good, apart from this small comments :+1:

product_gtin/tests/test_product_gtin_functions.py
victoralmau: It's not really necessary, i have removed it now.

product_gtin/tests/test_product_gtin_functions.py
victoralmau: Thanks, i have improved some functions to simplify validations.

joao-p-marques (approved):
Code and functional review :+1:

CarlosRoca13 (approved):

pedrobaeza (approved):

victoralmau:
It seems that the bot has locked, can you launch it again?

PR #361, opened 2021-08-26
[12.0][FIX] stock_barcodes: Tests fail when Internal reference defined

CarlosRoca13:
As the system shows the display name, when the product has a defined internal reference, the tests will fail.

With this change we make sure that what is displayed on the screen is the same as what is checked in the test.

cc @Tecnativa

ping @pedrobaeza @chienandalu


chienandalu (approved):

pedrobaeza (approved):

pedrobaeza:
Fast-tracking this obvious fix:

/ocabot merge patch

Is it needed in upper versions?

CarlosRoca13:
I will add it, to make sure that this is not going to fail again

PR #363, opened 2021-09-07
[12.0][IMP] stock_barcodes: Allow get lots automatically depends on removal strategy for picking operations

sergio-teruel:
cc @Tecnativa  TT31666

### DIFF
--- a/stock_barcodes/models/res_config_settings.py
+++ b/stock_barcodes/models/res_config_settings.py
@@ -7,9 +7,9 @@ from odoo import fields, models
 class ResConfigSettings(models.TransientModel):
     _inherit = 'res.config.settings'
 
-    stock_barcodes_inventory_auto_lot = fields.Boolean(
-        string='Get lots automatically for inventories',
-        related="company_id.stock_barcodes_inventory_auto_lot",
+    stock_barcodes_auto_lot = fields.Boolean(
+        string='Get lots automatically',
+        related="company_id.stock_barcodes_auto_lot",
         readonly=False,
     )
     group_track_pending_products_picking_barcode_wizard = fields.Boolean(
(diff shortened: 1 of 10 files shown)


stock_barcodes/wizard/stock_barcodes_read.py
```diff
@@ -56,6 +59,12 @@ class WizStockBarcodesRead(models.AbstractModel):
         ('success', 'Barcode read correctly'),
     ], readonly=True)
     message = fields.Char(readonly=True)
+    auto_lot = fields.Boolean(
+        string="Get lots automatically",
+        help="If checked the lot will be set automatically with the same "
+             "removal startegy",
+        default=lambda self: self._default_auto_lot(),
```
carlosdauden:
```suggestion
        default=_default_auto_lot,
```

carlosdauden (approved):

sergio-teruel:
Changes done!!!

stock_barcodes/models/res_config_settings.py
```diff
@@ -22,4 +22,4 @@ class ResConfigSettings(models.TransientModel):
 class ResCompany(models.Model):
     _inherit = "res.company"
 
-    stock_barcodes_inventory_auto_lot = fields.Boolean()
+    stock_barcodes_auto_lot = fields.Boolean()
```
chienandalu, line 25: A migration script should be done, isn't it?

chienandalu (commented):
I couldn't do the auto lot in a delivery picking (inventory works as expected):

Selección_01051

stock_barcodes/models/res_config_settings.py
sergio-teruel: I think that is not mandatory... If user see the check to False only has to set to true in settings....

chienandalu (approved):
Working on production for serveral months without further issue :+1:

pedrobaeza:
/ocabot merge minor

@chienandalu can you fw-port it to v13?

chienandalu:
> @chienandalu can you fw-port it to v13?

Sure

PR #348, opened 2021-06-25
[13.0][IMP][WIP] stock_barcodes: Improvements and restyling

sergio-teruel:
https://github.com/OCA/stock-logistics-barcode/pull/347 superseded

ping @etobella @JordiBForgeFlow @fshah-initos @olgamarcocb
This PR is in wip.. I hope that the next week we work hard to finish...

Improvements:
* Specific customizable menu visible in wizard interface
image
* Configurable options by type of operation
image
* New guided mode
image
* Restyling to adapt to mobile barcode readers

Other related modules:
* https://github.com/OCA/stock-logistics-barcode/pull/349
* https://github.com/OCA/stock-logistics-barcode/pull/327
* https://github.com/OCA/stock-logistics-barcode/pull/329
* https://github.com/OCA/stock-logistics-barcode/pull/350

PR #364, opened 2021-09-28
[13.][IMP] stok_barcodes: add html field to track pending products

albariera:
Cherry-pick from #338

### DIFF
--- a/stock_barcodes/__manifest__.py
+++ b/stock_barcodes/__manifest__.py
@@ -11,6 +11,7 @@
     "depends": ["barcodes", "stock"],
     "data": [
         "security/ir.model.access.csv",
+        "security/security.xml",
         "views/assets.xml",
         "views/res_config_settings_views.xml",
         "views/stock_inventory_views.xml",
--- a/stock_barcodes/__manifest__.py
+++ b/stock_barcodes/__manifest__.py
@@ -20,6 +20,7 @@
         "wizard/stock_barcodes_read_views.xml",
         "wizard/stock_barcodes_read_inventory_views.xml",
         "wizard/stock_barcodes_read_picking_views.xml",
+        "templates/missing_moves_template.xml",
     ],
     "installable": True,
 }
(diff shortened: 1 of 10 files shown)


MiquelRForgeFlow (approved):

LoisRForgeFlow (commented):
Wouldn't it be more useful to list next detailed operations (stock.move.line), this way you will have proper source/dest location, lot information, etc.

What do you think?

albariera:
@LoisRForgeFlow I agree with you. I have added an option to show detailed pending moves, non-detailed pending moves or both in case it is needed. This is how it looks if both are shown:
image
What do you think?

LoisRForgeFlow:
@albariera It looks good! A couple of comments though:

- In detailed operations can you rename "Initial Demand" by "Reserved"?
- How does it look the list in a phone screen? Can you paste an screenshot of that as well?

Thanks!

albariera:
"Intial Demand" already changed to "Reserved" in detailed operations (at non-detailed it remains as "Demand" if you agree)
Here you have the phone view (with web_responsive installed):
image
Thank you!

LoisRForgeFlow (approved):
LGTM, maybe a kanban view would help to visualice all the information in the table of detailed operations (the table overflows in the right). Anyway that can be a future improvement, thanks for considering my suggestions!

sergio-teruel:
All this is working in this PR https://github.com/OCA/stock-logistics-barcode/pull/348

sergio-teruel:
This commit has caused conflicts in the PR

pedrobaeza:
So should we revert it and continue with your PR?

LoisRForgeFlow:
@sergio-teruel @pedrobaeza Sorry, I thought that one was staled. Feel free to revert it, you are the maintainers of the module.

sergio-teruel:
Yes, please.
This changes we are working on this improvements...

pedrobaeza:
Reverted in #373

PR #366, opened 2021-10-06
[IMP] stock_barcodes: Find quant to select the proper location

albariera:
Cherry-pick from https://github.com/OCA/stock-logistics-barcode/commit/6b2ee8de6794fb0f0aecf8fd3fe3d404f26cff03, c5ef249, 
 11a251d99a9e22dc6ef26bdfd359bb2d0ffb69f2 and 0edeb58

### DIFF
--- a/stock_barcodes_gs1/wizard/stock_barcodes_read.py
+++ b/stock_barcodes_gs1/wizard/stock_barcodes_read.py
@@ -29,7 +29,7 @@ class WizStockBarcodesRead(models.AbstractModel):
         self.lot_id = lot
 
     def process_barcode(self, barcode):
-        """ Only has been implemented AI (01, 02, 10, 37), so is possible that
+        """Only has been implemented AI (01, 02, 10, 37), so is possible that
         scanner reads a barcode ok but this one is not precessed.
         """
         try:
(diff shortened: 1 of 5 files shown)


etobella (approved):
Thanks for checking the missing commits :smile:

olgamarcocb (approved):
LGTM 👍

stock_barcodes_gs1/wizard/stock_barcodes_read.py
```diff
@@ -0,0 +1,99 @@
...
+            lot = self._create_lot(barcode_decoded)
+        self.lot_id = lot
+
+    def process_barcode(self, barcode):
+        """Only has been implemented AI (01, 02, 10, 37), so is possible that
+        scanner reads a barcode ok but this one is not precessed.
+        """
+        try:
+            barcode_decoded = self.env["gs1_barcode"].decode(barcode)
+        except Exception:
```
rousseldenis, line 37: Is that generic catch wanted ?

etobella: Well, checking the code it can be replace by ValueError and ValidationError, however, I think it is right to keep it generic. First we try to transform to GS1, if it works we proceed, if it fails, we want to use the standard behavior.

etobella: Well, checking the code it can be replace by ValueError and ValidationError, however, I think it is right to keep it generic. First we try to transform to GS1, if it works we proceed, if it fails, we want to use the standard behavior.

PR #376, opened 2021-11-21
[15.0][MIG] barcodes generator product

sla-subteno-it:


.pre-commit-config.yaml
```diff
@@ -16,7 +16,7 @@ exclude: |
   (LICENSE.*|COPYING.*)
 default_language_version:
   python: python3
-  node: "14.18.0"
+  node: "17.1.0"
```
legalsylvain: Why this change for that module ?

sla-subteno-it: @legalsylvain I have a Mac M1, the 14.18.0 has not package for M1 so, I used the last one.

legalsylvain: I don't think you can change this setting for a single repo. If a setting is incorrect we should ping OCA maintainers to change it globally.
@pedrobaeza any point of view ?

pedrobaeza: Yes, indeed that's something that should be discussed in oca-addons-repo-template in an issue to make all OS compatible, but changing this now without testing other options may crash other configurations except yours. Please revert it or keep it by yourself. Anyways, I have heard that it's not easy to make Odoo work in a M1.

sla-subteno-it: @legalsylvain @pedrobaeza I just fix it, tI reverted the version.

legalsylvain:
@sla-subteno-it could you add me as maintainers for the both abstract and product modules ? I forget to do it when I developed them.
Thanks.

sebastienbeau:
Hi Seb ! Great to see you here
Maybe this fix can interest you https://github.com/OCA/stock-logistics-barcode/pull/381

PR #380, opened 2021-12-28
[14.0] _barcode_scanned invisible set 0

qianxunman:
It cannot working when the _barcode_scanned field set invisible is 1

When i set invisible to 0, it's works, just likes below:
图片

It's test on odoo 14.0


legalsylvain (commented):
just a remark, setting ``invisible="0"`` is unnecessary, just remove the ilne.
regarding the change, no idea. But I find it strange that it doesn't work when the field is invisible.
did you reproduced the bug on runbot ?

eantones (approved):
LGTM!

PR #384, opened 2022-01-28
[14.0][FIX] product_supplierinfo_barcode: barcode field not available on purchase tree view on v14

eantones:
The barcode field is not available on supplier info tree view and since in v14 you have no access to form view then the barcode is not available to fill up.


mariadforgeflow (approved):
LGTM!

LoisRForgeFlow (approved):
LGTM

PR #391, opened 2022-02-23
[14.0][FIX] stock_barcodes: make available the action of reading barcodes.

mariadforgeflow:
The field _barcode_scanned has to be visible in order to trigger the action of reading barcodes.

@JoanMForgeFlow
@LoisRForgeFlow


JoanMForgeFlow (approved):
Now it works,and it can read.
Thanks

LoisRForgeFlow (approved):

PR #396, opened 2022-03-12
ADD module sale_input_barcode

bealdav:
cc @jjscarafia @JordiBForgeFlow @sergio-teruel

### DIFF
--- a/sale_input_barcode/models/sale_order.py
+++ b/sale_input_barcode/models/sale_order.py
@@ -1,0 +1,19 @@
+# copyright 2022 David BEAL @ Akretion
+
+from odoo import models
+
+
+class SaleOrderLine(models.Model):
+    _name = "sale.order.line"
+    _inherit = ["sale.order.line", "product.barcode.line.mixin"]
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    def action_sale_line_barcode(self, barcode):
+        """Create a sale line according barcode information"""
+        self.ensure_one()
+        self.env["sale.order.line"].with_context(
+            order_id=self.id, company_id=self.company_id.id
+        )._process_barcode_on_product_line(barcode)
--- a/sale_input_barcode/tests/__init__.py
+++ b/sale_input_barcode/tests/__init__.py
@@ -1,0 +1,1 @@
+from . import test_module
--- a/sale_input_barcode/__init__.py
+++ b/sale_input_barcode/__init__.py
@@ -1,0 +1,1 @@
+from . import models
--- a/sale_input_barcode/__manifest__.py
+++ b/sale_input_barcode/__manifest__.py
@@ -1,0 +1,23 @@
+# Copyright (C) 2022 Akretion
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+
+{
+    "name": "Sale Input Barcode",
+    "version": "14.0.1.0.0",
+    "category": "Tools",
+    "website": "https://github.com/OCA/stock-logistics-barcode",
+    "author": "Akretion, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "installable": True,
+    "summary": "Add Sale line with barcode",
+    "maintainers": ["bealdav"],
+    "depends": [
+        "sale_management",
+        "barcode_action",
+        "base_gs1_barcode",
+    ],
+    "data": [
+        "views/sale.xml",
+    ],
+    "demo": [],
+}
--- a/sale_input_barcode/models/__init__.py
+++ b/sale_input_barcode/models/__init__.py
@@ -1,0 +1,2 @@
+from . import product_barcode_line_mixin
+from . import sale_order
--- a/sale_input_barcode/models/product_barcode_line_mixin.py
+++ b/sale_input_barcode/models/product_barcode_line_mixin.py
@@ -1,0 +1,73 @@
+# copyright 2022 David BEAL @ Akretion
+
+from odoo import _, models
+from odoo.exceptions import UserError
+
+
+class ProductLineMixin(models.AbstractModel):
+    _name = "product.barcode.line.mixin"
+    _description = "Utilities for any model to deal with product_id field"
+
+    def _process_barcode_on_product_line(self, barcode):
+        """
+        This mixin may be moved in another module to be used in mixed use cases
+        res =>
+        {
+            "01": "03400933816759",  # GTIN
+            "17": "2014-05-31",      # expiry date
+            "10": "B04059A",         # lot number
+            "310": 0.06385,          # weight
+            "15": "2014-05-01",      # sale date
+        }
+        """
+        res = self.env["gs1_barcode"].decode(barcode)
+        if not res.get("01"):
+            raise UserError(
+                _(
+                    "Decoded barcode %s doesn't include a valid segment for GTIN"
+                    % barcode
+                )
+            )
+        product = self.env["product.product"].search([("barcode", "=", res["01"])])
+        if product:
+            if len(product) > 1:
+                raise UserError(
+                    _(
+                        "These products %s share the same barcode.\n"
+                        "Impossible to guess which one to choose."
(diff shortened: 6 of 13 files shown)


sale_input_barcode/models/sale_order.py
```diff
@@ -0,0 +1,96 @@
...
+            elif res.get("17"):
+                lot = lot.filtered(lambda s: s.expiry_date == res["17"])
+                if not lot:
+                    raise UserError(_("Existing lot with the right date"))
+            vals["lot_id"] = lot.id
+            if "order_id" in self._fields:
+                vals["order_id"] = self.env.context.get("order_id")
+        self.create(vals)
+
+    def _should_i_create_unknown_lot(self, barcode):
```
legalsylvain, line 74: maybe a name like ``_handle_unknow_lot`` could be better.

legalsylvain:
Hi @bealdav. thanks for sharing that work.
2 minors remarks : 
- could you split code, 1 file per model ?
- could you rename the function ``_should_i_create_unknown_lot``, by something more generic. (``_handle_unknown_lot`` for exemple ?)

kind regards.

bealdav:
Changes done, thanks for the review @legalsylvain

sale_input_barcode/tests/__init__.py
```diff
@@ -0,0 +1 @@
+from . import test_
```
legalsylvain: a reason to set an underscore at the end of the file ?

bealdav: test_.py contains test for this module. This is the minimal name required else test unit is not triggered.
The module is simple, no need several files. In this case there is no need to have a longer name.
Logging is self explained

Screenshot from 2022-03-14 09-29-06

At least it works as it

legalsylvain: it doesn't work if the name is ``test.py`` ?

sale_input_barcode/models/product_line_mixin.py
```diff
@@ -0,0 +1,73 @@
+# copyright 2022 David BEAL @ Akretion
+
+from odoo import _, models
+from odoo.exceptions import UserError
+
+
+class ProductLineMixin(models.AbstractModel):
+    _name = "product.line.mixin"
```
rousseldenis: @bealdav I would have named it with 'barcode' as it is not a 'product-generic' mixin but 'barcode' oriented.

Moreover, I would have put it in another module as it is not specific to 'sale_input'

sale_input_barcode/tests/test_.py
```diff
@@ -0,0 +1,24 @@
+# © 2022 David BEAL @ Akretion
```
rousseldenis: It's quite uncommon to have that file naming.

See here for convention : https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#id12

sale_input_barcode/tests/test_.py
```diff
@@ -0,0 +1,24 @@
...
+    def test_input_line(self):
+        self.env = self.env(context=dict(self.env.context, tracking_disable=True))
+        barcode = "01034009338167591714050010B04059A\x1d310500638515140501"
+        product = self.env["product.product"].create(
+            {
+                "name": "barcode test",
+                "barcode": barcode[2:16],
+            }
+        )
+        assert product
```
rousseldenis: It's better to use self.assertTrue() as we are in Odoo tests context (and not pure python one).

rousseldenis (changes requested):
Code review. Some comments

sale_input_barcode/models/product_line_mixin.py
bealdav: Ok to change mixin name.

bealdav: For another module why not but if somebody may tells me : which module name, which repo, which use cases'll be implemented.

If no use case, I can add in roadmap section ?

rousseldenis: Yes. I told that because of your model description :

"Utilities for any model to deal with product_id field"

So, if you want to use it elsewhere you need to pull this one.

bealdav: I don't see any evident case but If somebody find it one, it'll be easy to make an extraction in another module.
Thanks to point me this option

sale_input_barcode/tests/__init__.py
bealdav: On some Odoo versions at least. I may test again with `test.py` if you prefer ! ? while it's not useless long name, no problem

sale_input_barcode/tests/test_.py
bealdav: I really don't share your opinion. I'll answer you ...

sale_input_barcode/tests/__init__.py
bealdav: I checked once again it's not triggered if only called `test.py`

sale_input_barcode/tests/test_.py
bealdav: While we do not use https://github.com/camptocamp/pytest-odoo in oca test I understand you ask me to avoid assert

rousseldenis: Indeed.

sale_input_barcode/tests/test_.py
bealdav: While there is only one file included test in 1 module, it's more readable to avoid to have module name in test file name

`openerp_test odoo.addons.sale_input_barcode.tests.test_: Starting Test.test_input_line ...`

vs

`openerp_test odoo.addons.sale_input_barcode.tests.test_sale_input_barcode: Starting Test.test_input_line`

In my opinion the first one is more readable avoiding duplicated information and could be the default one.
Not sure we have to trigger a long debate on this subject. I'm declared as maintainer on this module, I'll maintain it.

sale_input_barcode/tests/test_.py
bealdav: Do you think it could be interesting to promote pytest-odoo towards Oca ?
Are you in favor ?

sale_input_barcode/tests/test_.py
rousseldenis: Not to have a long debate about it and no need to change. But explicit things are better IMHO

sale_input_barcode/tests/test_.py
rousseldenis: We are currently having the debate internally. I personally use it for some weeks. That would imply a lot of changes. One of them is it will break tests flows for module changing some standard behaviours. That will lead to repositories numbers growing as we will need to identify groups of modules more precisely.

We can maybe open an issue on maintainers repo

bealdav: I don't want imply such changes with my own person as I have not the capacity to be involved to solve them.
Thanks for the information.

bealdav:
@rousseldenis I've just rebased

sale_input_barcode/tests/test_.py
legalsylvain: Hello.
well regarding that minor topic, at oca we generally name the file ``test_module.py``.

- https://github.com/OCA/pos/commits/12.0/pos_order_return_traceability/tests
- https://github.com/OCA/pos/tree/12.0/pos_order_mgmt/tests

no occurences of ``test_.py``.

sale_input_barcode/tests/__init__.py
legalsylvain: thanks. I didn't know.

sale_input_barcode/tests/test_.py
bealdav: test_module seems fine to me.

bealdav: I've applied this name

rousseldenis (approved):

francesco-ooops:
@bealdav it's not really clear from readme what's the usage of this module. Can you confirm user should click this button and then scan barcodes? Thanks!

image

bealdav:
Yes

Issue #401, opened 2022-03-22
[14.0] barcode is generated when no barcode rule is selected if a barcode rule has `automatic_generation=True` 


Mat-moran:
## Module
barcodes_generation_product

### automatic generation of barcodes IMO should be bounded to the rule itself
Now when you defined a rule with automatic generation it applies to all the products, even if you do not select this rule during the creation of the product. IMO the automatic barcode generation should only happen if you select an automatic rule in the product.

## To Reproduce
**14.0**:

1. Create a barcode nomenclature
2. Create a new rule for EAN13 barcode with automatic generation
3. Create a new product without selecting the automatic generation rule
4. You should get the new product with a barcode generated.

**Expected behavior**
The automatic generation of the barcode IMO should be bounded the the context of the rule. Only generate an automatic barcode if you select one of the automatic rules you can have in the system.

**Additional context**

legalsylvain:
@lasley : could you take a look on this one, as you implemented the automatic mechanism.

thanks !

Mat-moran:
Hi @lasley sorry for ping you but this matter is urgent for us, are you going to check this? If you can not do that we can give a try and do a PR.

rafaelbn:
@lasley doesn't' work in Odoo anymore at least 3 years ago.

Mat-moran:
Thanks @rafaelbn,
@rafaelbn, @legalsylvain  are you OK if we propose a PR in this direction??

legalsylvain:
>  @rafaelbn, @legalsylvain are you OK if we propose a PR in this direction??

hi @Mat-moran

Indeed ! Please add tests for the automatic generation workflow.

rafaelbn:
My opinion is I'm agree with this direction. I'm not PSC of this repo.

My suggestion is that more @OCA/logistics-maintainers-psc-representative could give their vision.

PR #407, opened 2022-05-05
[ADD][WIP] Change of functionality for automatic generation of bar codes

cramosh:
In this PR the main modification is that the `generate_automate` checkbox is bounded to the rule itself. 
Behaviour before this PR
When you click the automatic checkbox inside a rule it apply to all products in your inventory. It has no sense because:
The checkbox is inside the rule
You can not have two or more product barcode rules in your database

Behaviour after this PR
When you click the automatic checkbox inside a rule, it is bounded to the usage of this rule in the product template, solving the main two problems of the previous behaviour.

In order to keep the old behaviour, a new checkbox `general_rule` is added to put this rule by default when you create a new product. This checkbox can only be checked in one rule with the same `generate_model`.


legalsylvain (changes requested):
>  You can not have two or more product barcode rules in your database

hi @cramosh. At a first sight, this assertion is wrong. having many product barcode rules is totally valid. see for exemple ``barcode_rule_weight_three_dec`` and ``barcode_rule_price_two_dec`` in odoo data.

Mat-moran:
Hi @legalsylvain ,

`barcode_rule_weight_three_dec` and `barcode_rule_price_two_dec` are modules? we do not find tthem.

Please can you clarify it?

legalsylvain:
@Mat-moran Sorry for not being clear.

``barcode_rule_weight_three_dec`` and ``barcode_rule_price_two_dec`` are xml_ids of barcode rule of products created by odoo.

- https://github.com/odoo/odoo/blob/14.0/addons/point_of_sale/data/default_barcode_patterns.xml#L30
- https://github.com/odoo/odoo/blob/14.0/addons/stock/data/default_barcode_patterns.xml#L4

cramosh:
> > You can not have two or more product barcode rules in your database
> 
> hi @cramosh. At a first sight, this assertion is wrong. having many product barcode rules is totally valid. see for exemple `barcode_rule_weight_three_dec` and `barcode_rule_price_two_dec` in odoo data.

What I mean by not being able to have more than one rule is that odoo does not allow you to create more than one automatic rule for barcodes.

On the other hand, having it marked, what the program does is always use that rule, even though I do not have a different rule selected in the product sheet.

This PR is a propose for solved Issue #401

PR #408, opened 2022-05-05
[ADD] Change functionality generate barcode product.

cramosh:
n this PR the main modifications are:

`barcode_rule_id` is not related with the field `product.product` because was inducing wrong behaviour during variants creation when the rule was not set as `automatic`. This functionality has been achieved modifying the `write` function to assure that if you change anything in the `product.template` it is changed in the `product.product`. The main drawback is that the `barcode_rule_id` in the `product.prouduct` is now `readonly`. 
A new action `generate_barcode` is added to generate a barcode in a product with variants when you add more variants after the barcode is generated.

### DIFF
--- a/barcodes_generator_product/models/product_template.py
+++ b/barcodes_generator_product/models/product_template.py
@@ -11,7 +11,7 @@ class ProductTemplate(models.Model):
 
     # Related to display product product information if is_product_variant
     barcode_rule_id = fields.Many2one(
-        related="product_variant_ids.barcode_rule_id",
+        # related="product_variant_ids.barcode_rule_id",
         string="Barcode Rule",
         readonly=False,
         comodel_name="barcode.rule",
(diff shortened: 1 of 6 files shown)


rousseldenis:
@cramosh Still WIP ?

cramosh:
> @cramosh Still WIP ?

Thanks @rousseldenis . This PR is prepare to review and merge.

barcodes_generator_product/models/product_template.py
```diff
@@ -11,7 +11,7 @@ class ProductTemplate(models.Model):
 
     # Related to display product product information if is_product_variant
     barcode_rule_id = fields.Many2one(
-        related="product_variant_ids.barcode_rule_id",
+        # related="product_variant_ids.barcode_rule_id",
```
legalsylvain, line 14: I dont think that removing thé related is a good idea. Could you elaborate the current use case that is failing ?

legalsylvain (changes requested):

PR #412, opened 2022-06-30
[14.0][UPD] product_multi_barcode: Added menu to see all barcodes. Up…

ilyasProgrammer:
…dated product variant easy edit form to see barcodes.

### DIFF
--- a/product_multi_barcode/readme/CONTRIBUTORS.rst
+++ b/product_multi_barcode/readme/CONTRIBUTORS.rst
@@ -5,3 +5,7 @@
 * Héctor Villarreal (Eficent)
 * Sébastien Beau (Akretion)
 * Kevin Khao (Akretion)
+
+* Ooops404 <https://ooops404.com>
+
+  * Ilyas <irazor147@gmail.com>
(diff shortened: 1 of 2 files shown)


francesco-ooops (approved):
The view in product variant is to reflect the possible presence of multiple barcodes.

The menu addition is to allow an easy import of multi-barcodes. If the added dependency on 'stock' is an issue we can create another module just for the menu, but being a module of repo 'stock-logistics-barcode' thought it would not be too much of a problem. 

@OCA/logistics-maintainers let us know what you think :)

elvise (approved):
LGTM!

product_multi_barcode/__manifest__.py
```diff
@@ -15,11 +15,12 @@
     "ADHOC SA, "
     "Odoo Community Association (OCA)",
     "category": "Product Management",
-    "depends": ["product"],
+    "depends": ["product", "stock"],
```
rousseldenis: IMHO adding a new dependency should be avoided.

You can add a new module instead.

ilyasProgrammer: ok

francesco-ooops (approved):
@rousseldenis ok, now this enhancement PR contains only multi-barcode lines in variant view

menu has been moved to #413 

good for merge? :)

rousseldenis (approved):

PR #415, opened 2022-07-08
[13.0][FIX] stock_barcodes: several small adjustments

LoisRForgeFlow:
* When a picking is processed partially it was not
      posible to validate it because the backorder wizard
      was captured and never returned to the user.
* Only update sml locations when there is a location
      to update, otherwise constraints for non null values
      could raise.
* Invalidate cache of todo lines when determining
      next action to do to ensure that line status is
      updated.
* Reword string in button from "Force" to "Done"
      as it was a bit misleading. Actually, in spanish
      translation it was already being translated to
      "Hecho" ("Done").

Extra, to be discussed, in a separate commit for now:

* show quantity even when not editable.

Desktop view:

image

Mobile view:
image

@ForgeFlow

### DIFF
--- a/stock_barcodes/wizard/stock_barcodes_read_picking.py
+++ b/stock_barcodes/wizard/stock_barcodes_read_picking.py
@@ -165,6 +165,10 @@ class WizStockBarcodesReadPicking(models.TransientModel):
             return False
         if not self.todo_line_ids:
             self.fill_todo_records()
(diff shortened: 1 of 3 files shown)


LoisRForgeFlow:
cc @sergio-teruel @etobella

sergio-teruel (approved):

sergio-teruel:
Good for me

etobella (approved):

Issue #416, opened 2022-07-11
[14.0] product_multi_barcode: delete product containing barcode, but barcode is still assigned to product "false"


francesco-ooops:
[VIDEO](https://recordit.co/neQRm18ekC)

## Module
product_multi_barcode

## Describe the bug
Barcode records are not deleted when product they are related to are deleted

## To Reproduce
**Affected versions**:

1. Assign barcode 1 to product A
2. Delete product A
3. Assign barcode 1 to product B
4. Error: The Barcode "barcode 1" already exists for product "False"

**Expected behavior**
Delete barcode record as per standard behavior

PR #418, opened 2022-07-12
[14.0][FIX] product_multi_barcode: Support product deletion.

ilyasProgrammer:


francesco-ooops (approved):
Functional review ok

fixes #416

elvise (approved):
LGTM!

francesco-ooops:
@OCA/logistics-maintainers could you check? thanks!

francesco-ooops:
@OCA/logistics-maintainers could you please take a look?

product_multi_barcode/models/product_barcode.py
```diff
@@ -27,12 +27,14 @@ class ProductBarcode(models.Model):
...
     )
     product_tmpl_id = fields.Many2one(
         comodel_name="product.template",
         compute="_compute_product_tmpl",
         store=True,
         readonly=False,
+        ondelete="cascade",
     )
 
     @api.depends("product_id")
```
rousseldenis, line 40: @ilyasProgrammer I think the problem you mention is because this depends does not triggered the recompute as it is missing product_tmpl_id.

```suggestion
    @api.depends("product_id.product_tmpl_id")
```

But, at least your fix is good as it is readonly False, so, could be changed manually.

rousseldenis (changes requested):

rousseldenis:
@ilyasProgrammer The better for reviewers is to have a good PR title. Please change to something like '[14.0][FIX] product_multi_barcode: Support product deletion'

rousseldenis:
Could you also provide a test that will help avoiding regression ?

ilyasProgrammer:
> Could you also provide a test that will help avoiding regression ?

@rousseldenis What kind of test ?

rousseldenis:
> > Could you also provide a test that will help avoiding regression ?
> 
> @rousseldenis What kind of test ?

No, that's ok, we can assume ondelete works :-)

rousseldenis (approved):

PR #421, opened 2022-07-28
[13.0][FIX] stock_barcodes: Do not distribute lots for non tracking products for inventories

sergio-teruel:
cc @Tecnativa

ping @chienandalu

### DIFF
--- a/stock_barcodes/wizard/stock_barcodes_read_inventory_views.xml
+++ b/stock_barcodes/wizard/stock_barcodes_read_inventory_views.xml
@@ -29,12 +29,6 @@
                                             <field name="location_ids" />
                                         </div>
                                     </div>
-                                    <div class="row">
-                                        <div class="col">
-                                            <strong>Products:</strong>
-                                            <field name="product_ids" />
-                                        </div>
-                                    </div>
                                 </div>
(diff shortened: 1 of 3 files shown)


chienandalu (approved):
:+1:

stock_barcodes/wizard/stock_barcodes_read_inventory_views.xml
```diff
@@ -29,12 +29,6 @@
                                             <field name="location_ids" />
                                         </div>
                                     </div>
-                                    <div class="row">
```
pedrobaeza, line 32: Why removing this?

pedrobaeza (approved):

stock_barcodes/wizard/stock_barcodes_read_inventory_views.xml
chienandalu: Not very practical. When there are many references it takes to much space in the screen

pedrobaeza: Ok, but then do a separate commit with proper commit message explaining that or at least comment it in the same commit

PR #427, opened 2022-08-24
[FIX] product_multi_ean: Archived Product prevents creating new EAN

SirTakobi:
**Steps**

1. Create a product **A**
2. Assign an EAN **test** to product **A**
3. Archive product **A**
4. Create a product **B**
5. Assign an EAN **test** to product **B**

**Actual behavior**
Error raised:
> Error while validating constraint
>
> The EAN13 Barcode "test" already exists for product "A"

**Expected behavior**
No error

**Additional context**
The part
> Error while validating constraint

was there because the constraint raised a `UserError` instead of a `ValidationError`, fixed in a separated commit

A couple of commits have also been spent to be compliant with https://github.com/OCA/oca-addons-repo-template.


SirTakobi:
Superseded by https://github.com/OCA/stock-logistics-barcode/pull/531

PR #430, opened 2022-09-01
[14.0][FIX] stock_barcodes: make actions readable by non-admin users.

LoisRForgeFlow:
leftover from migration.

@ForgeFlow

cc @sergio-teruel

### DIFF
--- a/stock_barcodes/wizard/stock_barcodes_read_picking.py
+++ b/stock_barcodes/wizard/stock_barcodes_read_picking.py
@@ -758,7 +758,9 @@ class WizCandidatePicking(models.TransientModel):
         if isinstance(res, dict):
             # backorder wizard
             return res
-        return self.env.ref("stock_barcodes.action_stock_barcodes_action").read()[0]
+        return self.env["ir.actions.actions"]._for_xml_id(
+            "stock_barcodes.action_stock_barcodes_action"
+        )
 
     def action_open_picking(self):
         picking = self.env["stock.picking"].browse(
--- a/stock_barcodes/wizard/stock_barcodes_read_picking.py
+++ b/stock_barcodes/wizard/stock_barcodes_read_picking.py
@@ -254,7 +254,7 @@ class WizStockBarcodesReadPicking(models.TransientModel):
             "picking_id": picking.id,
             "move_id": candidate_move.id,
             "qty_done": available_qty,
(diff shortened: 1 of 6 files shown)


pedrobaeza (approved):

LoisRForgeFlow:
@pedrobaeza I added one more change, please check the latest commit and update your review if needed.

stock_barcodes/wizard/stock_barcodes_read_picking.py
```diff
@@ -254,7 +254,7 @@ def _prepare_move_line_values(self, candidate_move, available_qty):
             "picking_id": picking.id,
             "move_id": candidate_move.id,
             "qty_done": available_qty,
-            "product_uom_id": self.product_id.uom_po_id.id
+            "product_uom_id": candidate_move.product_uom.id or self.product_id.uom_id.id
```
pedrobaeza, line 257: Shouldn't this depend on the type of picking? cc @sergio-teruel

LoisRForgeFlow: Even if it is a receipt it does not imply it is a purchase. Using the PO UoM looks like the less frequent case here. If we have a move and the move comes from a PO, the move will already have the PO UoM as `product_uom`. Covering more complex cases would need a dependency with purchase I think.

pedrobaeza: OK, I think it's correct to use the move UoM, but then the `or` is never going to be used, as it always has a value. Let's see Sergio to tell us why he put such UoM.

sergio-teruel (approved):

stock_barcodes/wizard/stock_barcodes_read_picking.py
LoisRForgeFlow: There is actually a case in which is going to be used: when you create a new non-planned picking, in that case you start to scan operations and there is no stock move yet until you validate the operation you just scanned.

pedrobaeza: OK, maybe in that case the UoM should depend on the picking type code.

pedrobaeza: Uhm, thinking twice, no, it should always put the stock UoM, not the PO one, so it's correct.

PR #431, opened 2022-09-02
[13.0][FIX] stock_barcodes: do not use PO UoM by default

LoisRForgeFlow:
When creating new stock move lines, we should use the UoM in the
move, if any, or the default UoM of the product.

Backport of #430 

cc @pedrobaeza @sergio-teruel


pedrobaeza (approved):

pedrobaeza:
Absolutely!

Let's fast-track it as already merged in v14:

/ocabot merge patch

PR #432, opened 2022-09-02
[14.0][FIX] stock_barcodes: remove last scan do not revert last action.

LoisRForgeFlow:
In pickings, the scan log was always generated without details of the move lines modified/created. Without this information in the log removing the log was dong nothing else than just that, removing the log; leaving the last operation still reflected in the picking.

Also, when the picking was a draft one, the stock move line and the stock move could have been created by the scan itself, so we
need to remove it in that case, reverting the action would leave the stock move there that could be validated by mistake.

@ForgeFlow 

cc @pedrobaeza @sergio-teruel

### DIFF
--- a/stock_barcodes/wizard/stock_barcodes_read.py
+++ b/stock_barcodes/wizard/stock_barcodes_read.py
@@ -454,8 +454,9 @@ class WizStockBarcodesRead(models.AbstractModel):
             self.product_qty = 1.0
         if not self.check_done_conditions():
             return False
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):

LoisRForgeFlow:
@pedrobaeza @sergio-teruel one more small fix added in the last commit.

sergio-teruel:
Perfect @LoisRForgeFlow !!

PR #445, opened 2022-10-10
[16.0][MIG] base_gs1_barcode

tewtawat:
Standard Migration


rousseldenis:
/ocabot migration base_gs1_barcode

@tewtawat @StefanRijnhart @sergio-teruel @gurneyalex @fshah-initos @JasminSForgeFlow Have you seen that in v15, there is a module for gs1 (barcodes_gs1_nomenclature).

Shouldn't this be updated accordingly ?

simahawk:
> Shouldn't this be updated accordingly ?

I agree. 

These days we are checking `barcodes_gs1_nomenclature` on v15 and we'll probably backport it to v14 (where we need it).

Despite I don't like that much its code (surprise surprise :stuck_out_tongue_closed_eyes: ) I don't see why we should carry on `base_gs1_barcode`.
Honest question: what's the added value?

sergio-teruel:
Yes I know, I have yet to take a look at it and see if we can remove the current base_gs1_barcode module

edescalona:
Hi @sergio-teruel , could you please tell me if it was possible to replace the base_gs1_barcode module with barcodes_gs1_nomenclature? Thank you for your contribution.

sergio-teruel:
Yes... already done in 16.0

edescalona:
Thank you @sergio-teruel  for your response

PR #446, opened 2022-10-10
[16.0][MIG] product_supplierinfo_barcode

tewtawat:
Standard Migration


rousseldenis:
@tewtawat Thanks for this. Could you just add in the description the dependency PR link in order to easy review? Thanks

tewtawat:
@rousseldenis okay but there're no dependency pr for this module. what should i have to do

rousseldenis:
> @rousseldenis okay but there're no dependency pr for this module. what should i have to do

Ok, sorry. :sweat_smile:

rousseldenis (approved):
Code review

marylla (changes requested):
I did a functional review.

I can only set the barcode for a supplier info in its form view when using this menu: purchase > configuration > vendor pricelists.

Unfortunately, it is not possible to add this barcode in product form view in page "purchase" because the tree view for supplier info has a "edit" option to edit it inline and not open a form view. 

I think, we should add the barcode field also in supplier info tree view.

much-sparenberg (approved):
Weirdly it worked now for me too. :) So everything fine now.

marylla (approved):
Weirdly it worked now for me too. :) So everything fine now.

product_supplierinfo_barcode/views/product_views.xml
```diff
@@ -0,0 +1,17 @@
...
+        <field name="arch" type="xml">
+            <field name="categ_id" position="after">
+                <field
+                    name="seller_ids"
+                    string="Supplier code"
+                    filter_domain="['|', '|', ('seller_ids.barcode', 'ilike', self), ('seller_ids.product_code', 'ilike', self), ('seller_ids.product_name', 'ilike', self)]"
+                />
+            </field>
+        </field>
+    </record>
```
enriquemartin, line 16: Why not migrate the view that adds the barcode to the supllierinfo tree in the product form?
Like in version 15: https://github.com/OCA/stock-logistics-barcode/blob/147a3020ebb92fb974f42e115fcc0bbba3b65188/product_supplierinfo_barcode/views/product_views.xml#L18-L27
Tested in local environment and works correctly.

hildickethan (approved):

hildickethan:
@tewtawat could you rebase the branch? to fix any conflict before merge as it seems like the bot had issues doing that

product_supplierinfo_barcode/__manifest__.py
```diff
@@ -0,0 +1,25 @@
...
+    "website": "https://github.com/OCA/stock-logistics-barcode",
+    "author": "TAKOBI, Odoo Community Association (OCA)",
+    "maintainers": ["eLBati"],
+    "license": "LGPL-3",
+    "application": False,
+    "installable": True,
+    "preloadable": True,
+    "depends": [
+        # we need purchase to view seller_ids in product form
+        "purchase",
```
pedrobaeza, line 18: Is this still applicable?

product_supplierinfo_barcode/models/supplierinfo.py
```diff
@@ -0,0 +1,18 @@
...
+
+    barcode = fields.Char(
+        copy=False,
+        help="Article number used by supplier for product identification.",
+    )
+
+    _sql_constraints = [
+        (
+            "barcode_uniq",
+            "unique(barcode)",
```
pedrobaeza, line 15: I think this constraint is too restrictive, as several suppliers may have the same barcode. It should be a combination of unique company, seller and barcode

pedrobaeza (changes requested):
And check the other existing comment

PR #449, opened 2022-10-24
[16.0][MIG] barcodes_generator_abstract (from 14.0 to 16.0)

legalsylvain:
simple migration.

- bump revision
- update fr translation
- use ``@api.model_create_multi``
- remove redundant string on some fields.
- replace obsolete call of ``invalidate_cache`` by ``invalidate_model``
- fix setup, using : external_dependencies_override

/ocabot migration barcodes_generator_abstract

CC : @pierrickbrun


PierrickBrun (approved):

PierrickBrun:
Why has the python_dependencies changed ? Is it because odoo handles it differently now ?

rousseldenis:
> Why has the python_dependencies changed ? Is it because odoo handles it differently now ?

Indeed @legalsylvain, it seems libraries haven't same purpose:

image

barcodes_generator_abstract/models/barcode_rule.py
```diff
@@ -153,4 +150,4 @@ def _clear_cache(self, vals):
         """It clears the caches if certain vals are updated."""
         fields = ("generate_model", "generate_automate")
         if any(k in vals for k in fields):
-            self.invalidate_cache(fields)
+            self.invalidate_model(fields)
```
rousseldenis, line 153: Question: have you tested without this _clear_cache() function ?

legalsylvain:
@PierrickBrun, @rousseldenis : i didn't changed the dependency. See external_dependencies_override

rousseldenis (approved):

pedrobaeza:
I think you need to change the manifest external_dependencies to match the pypi package name.

barcodes_generator_abstract/models/barcode_rule.py
legalsylvain: Hi. 
I don't know why this part has been introduced.
I removed this part and the test still works. ;-)
It has been intro by @lasley in V13 https://github.com/OCA/stock-logistics-barcode/commit/51d9ef002e86d86e1b0226d118e1e0cbd6b0f134

No idea what to do. I remove ?

legalsylvain:
> I think you need to change the manifest external_dependencies to match the pypi package name.

I don't understand. You mean write "python-barcode" in manifest file ?

pedrobaeza:
If that one is the correct pypi package name, yes.

rousseldenis:
> > I think you need to change the manifest external_dependencies to match the pypi package name.
> 
> I don't understand. You mean write "python-barcode" in manifest file ?

Yes, I'm wondering the same. If you remove override and restore the previous manifest. Would it find it ?

legalsylvain:
if manifest contains python-barcode
and setup.odoo_addon=True
**Then**

```
barcodes_generator_abstract/models/barcode_generate_mixin.py:9: [W7936(missing-manifest-dependency), ] Missing external dependency "barcode" from manifest. More info: https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#external-dependencies
```

@pedrobaeza, @rousseldenis : What I have to do ?

pedrobaeza:
That linter seems incorrect. Not sure if it can be fixed. cc @moylop260 

But proper solution right now is to ignore such lint error.

legalsylvain:
> But proper solution right now is to ignore such lint error.

OK. done !

moylop260:
@sbidoul @pedrobaeza 

IMHO the manifest dependency checks could be disabled for new Odoo versions supporting package names instead of import-lib-name

The lints did its works when Odoo imported the module if it has static folder and raised errors

But now,

- The manifest support packages pypi names
- Odoo is not importing the modules if they are not installed
- oca bot is auto-generating the requirements based on manifest
- there is not a way to know the imported package vs the pypi package related

pylint-odoo >=8.0 were deprecated all them

For pylint-odoo <=7.0 they need to be disabled from pylint.cfg

pedrobaeza:
Can you do it in the copier template or point me how to do it?

moylop260:
> Can you do it in the copier template or point me how to do it?

I'm not the best to change copier template

In fact, I'm looking for help in order to enable all the following pylint checks:
 - https://github.com/OCA/oca-addons-repo-template/issues/164

Maybe @yajo or @sbidoul

PR #451, opened 2022-10-25
[13.0] [FIX] stock_barcodes

sanchonuria:
When testing stock barcodes I found some errors:

- Pending moves where used, and it was confusing when using purchase orders, as the first element was not shown
- Changing the kind of field for INs, as in purchases, we don't know the lot, so we should need to use stock.moves not stock.move.line
- Putaway strategy was not being applied

@etobella @LoisRForgeFlow @sergio-teruel

### DIFF
--- a/stock_barcodes/data/stock_barcodes_option.xml
+++ b/stock_barcodes/data/stock_barcodes_option.xml
@@ -118,7 +118,7 @@
             <field name="barcode_guided_mode" />
             <field name="show_pending_moves">True</field>
             <field name="confirmed_moves">False</field>
-            <field name="source_pending_moves">move_line_ids</field>
+            <field name="source_pending_moves">move_lines</field>
             <field name="fill_fields_from_lot">False</field>
         </record>
         <record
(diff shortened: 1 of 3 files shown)


kluna1998 (approved):

sergio-teruel (approved):

sergio-teruel (changes requested):
Hi @sanchonuria 
Some comments... 

> Putaway strategy was not being applied

You can not apply always the putaway strategy when the sml is create, although a product has a putaway strategy the user can move the product to another location.
Maybe the better option is apply the strategy in product onchage with a conditions like this:

```
@api.onchange("product_id")
def onchange_product_id(self):
    super().onchange_product_id()
    if (self.option_group_id.source_pending_moves == "move_line_ids" 
        and self.option_group_id.option_ids.filtered(
            lambda o: o.field_name == "location_dest_id" 
                      and o.filled_default 
                      and not o.to_scan
        )
    ):
        pending_lines = self.pending_move_ids.filtered(lambda ln: ln.product_id == self.product_id and ln.state == 'pending')
        if pending_lines:
            location_dest = pending_lines.mapped("location_dest_id")
            if len(location_dest) == 1:
                self.location_dest_id = location_dest

```

PR #465, opened 2022-11-18
[14.0][IMP] barcodes_generator_abstract, barcodes_generator_product, barcodes_generator_location: add many sequences 

ThiagoMForgeFlow:
The barcodes generator abstract module now allows you to create barcode rules based on many sequences, handling the jump between them. The barcodes generator product and barcodes generator location modules now support assigning barcodes from the rule created by the abstract module.

@ForgeFlow


barcodes_generator_abstract/models/barcode_rule.py
```diff
@@ -10,11 +10,11 @@
     ("no", "No generation"),
     ("manual", "Base set Manually"),
     ("sequence", "Base managed by Sequence"),
+    ("many sequences", "Base managed by Many Sequences"),
```
legalsylvain: Could you replace space by underscore ?

barcodes_generator_location/models/stock_location.py
```diff
@@ -49,7 +49,13 @@ def fields_view_get(
                 placeholder.addprevious(barcode_field)
             # To the `barcode` field in the view (either the new
             # or the existing one), a modifier is added.
-            modifier = {"readonly": [("generate_type", "=", "sequence")]}
+            modifier = {
+                "readonly": [
+                    "|",
+                    ("generate_type", "=", "many sequences"),
```
legalsylvain: In operator doesnt work ?

legalsylvain (changes requested):
Some remarks. I'm not sure about the design proposed but first i'd like to understand the feature. Could you describe a use case for product barcode generation ? (Configuration, user actions, results)

Thanks !

barcodes_generator_abstract/models/barcode_rule.py
ThiagoMForgeFlow: Done, thanks

barcodes_generator_location/models/stock_location.py
ThiagoMForgeFlow: Done, thanks

ThiagoMForgeFlow:
> Some remarks. I'm not sure about the design proposed but first i'd like to understand the feature. Could you describe a use case for product barcode generation ? (Configuration, user actions, results)
> 
> Thanks !

This module is intended for the case in which a company takes ownership of a set of barcodes. But the set is divided into different sequences, and these are to be managed as if it were a single generation rule. For example:

- 2000001000007
- 2000002000006
- 2500001000002
- 2500002000001
- 2500003000000

So now you can create a barcode rule "many_sequences", you select 2 sequences to create, once created you can modify them and put the prefix, the next number and the last number that the sequence allows. In the previous sequences, the following would be modified:

Sequence 1:
-prefix: 20
- next number: 1
-last number: 2
Sequence 2:
-prefix: 25
- next number: 1
-last number: 3

With these settings, the created barcode rule will generate the expected barcodes for the products, managing the jump between the sequences.

PR #405, opened 2022-04-07
[15.0][MIG] stock_barcodes_gs1

JasminSForgeFlow:
Standard Migration

@ForgeFlow

Depends on:
 - https://github.com/OCA/stock-logistics-barcode/pull/404

PR #473, opened 2022-12-06
[14.0] Port barcodes_gs1_nomenclature from odoo 15.0

simahawk:
@legalsylvain @pedrobaeza I'd like to get your feedback on this, especially on guidelines on how to maintain modules grabbed from Odoo SA (eg: apply pre-commit fully or not?)

FTR I squashed all Transifex commits.

Relates to
- https://github.com/OCA/stock-logistics-barcode/pull/445 
- https://github.com/OCA/stock-logistics-barcode/pull/405


simahawk:
There are some changes required on `barcodes` too: I wonder if it wouldn't be better to keep all changes + this new module into OCB instead of here... 

https://github.com/odoo/odoo/commit/1c0dec6d5813fa81a0fec8668ee13d8036cde5c5

legalsylvain:
No idea. Sorry.

pedrobaeza:
It's stated on the guidelines:

https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#backporting-odoo-modules

simahawk:
> It's stated on the guidelines:
> 
> https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#backporting-odoo-modules

Well, I guess I was too tired yesterday to find it... I checked but I was kind of blind :sweat: 

Thanks @pedrobaeza :)

What about [the last point? ](https://github.com/OCA/stock-logistics-barcode/pull/473#issuecomment-1339694797).
It will be easier to keep all changes into OCB unless we want to port here `barcodes` from v15 too and explain users how to make it override the default one.

pedrobaeza:
You should patch the thing in your module or an extra one, but no changes in OCB should be done so that a module is incompatible with Odoo repo and being only with OCB one.

simahawk:
> You should patch the thing in your module or an extra one, but no changes in OCB should be done so that a module is incompatible with Odoo repo and being only with OCB one.

I'll look for a less obtrusive way but w/o those changes in `barcodes` the module here is totally useless.

Maybe a small glue module and make this one depending on it...

pedrobaeza:
Yes, that can be another option, but if it's going to be only used by this one, include that changes here.

simahawk:
Upon further testing/checking we decided to go for another implementation as we don't have to cover all the use cases supported by the modules in question, especially because the changes in the core are way more hard to backport and maintain.

PR #476, opened 2022-12-19
[14.0][MIG] barcodes_generator_partner: Migration to 14.0

jguenat:
standard migration


barcodes_generator_partner/__manifest__.py
```diff
@@ -6,14 +6,12 @@
...
-    "version": "12.0.1.0.0",
+    "version": "14.0.1.0.0",
     "category": "Tools",
     "author": "GRAP," "La Louve," "Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/stock-logistics-barcode",
     "license": "AGPL-3",
-    "depends": [
-        "barcodes_generator_abstract",
-    ],
+    "depends": ["barcodes_generator_abstract", "point_of_sale"],
```
rousseldenis: @jguenat Why adding point_of_sale in dependencies ?

rousseldenis (commented):
IMHO, there is no need to change the dependency

barcodes_generator_partner/__manifest__.py
jguenat: Because the odoo view adding the barcode field is in point_of_sale in V14.
I guess because functionally speaking in odoo's viewpoint you only really need partner barcode in point of sale.

I agree technically it does not make sense because the res.partner barcode field is defined in base.

I guess I could just add the barcode field on the base res.partner view and get rid of this dependency.

What do you think?

jguenat: @rousseldenis 
I've removed the dependency to point_of_sale. The only downfall is that if you install point_of_sale, there will be two barcode fields on the partner form and the one from point_of_sale won't have the 'readonly': [('generate_type', '=', 'sequence')].
But I think we can live with that.
If OK for you like that I'll fixup the last commit.

Issue #463, opened 2022-11-15
[14.0][stock_barcodes] method action_force_done() is triggered reading barcodes


Bilbonet:
When you read a product barcode that is not in the picking. 

1. The wizard shows  the wrong product found.
2. The button with the action action_force_done() is visible.

Then, if you read another barcode the button with the action is triggered. And create a new stock move.

![stock_barkodes](https://user-images.githubusercontent.com/8178160/201987867-daf0fc34-5d86-471e-94f3-d58cbf442417.gif)

rousseldenis:
@sergio-teruel @LoisRForgeFlow @pedrobaeza @chienandalu

pedrobaeza:
Closing as there's an existing PR. Next time, if you plan to do a PR, you can explain the issue in the PR itself instead of opening 2 things.

sergio-teruel:
@Bilbonet CAn you try this on v15?

Bilbonet:
@sergio-teruel @pedrobaeza @rousseldenis 
I've tried on v14 and v15 and now works fine.
I was wrong

PR #477, opened 2022-12-20
[14.0][FIX] stock_barcodes avoid method action_create_move reading product …

Bilbonet:
The situation explained here #463

### DIFF
--- a/stock_barcodes/static/src/js/basic_controller.js
+++ b/stock_barcodes/static/src/js/basic_controller.js
@@ -283,13 +283,8 @@ odoo.define("stock_barcodes.BasicController", function (require) {
                             const $action_confirm = self.$(
                                 "button[name='action_confirm']:visible"
                             );
-                            const $action_confirm_force = self.$(
-                                "button[name='action_force_done']:visible"
-                            );
                             if ($action_confirm.length) {
                                 $action_confirm.click();
-                            } else if ($action_confirm_force.length) {
-                                $action_confirm_force.click();
                             }
                         }
                     }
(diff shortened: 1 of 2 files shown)


Bilbonet:
Can anybody review this?
Thanks

rousseldenis:
@sergio-teruel @LoisRForgeFlow @pedrobaeza @chienandalu

stock_barcodes/static/src/js/basic_controller.js
```diff
@@ -283,13 +283,8 @@ odoo.define("stock_barcodes.BasicController", function (require) {
                             const $action_confirm = self.$(
                                 "button[name='action_confirm']:visible"
                             );
-                            const $action_confirm_force = self.$(
```
pedrobaeza, line 286: Why are you removing these features?

Bilbonet: In the context of doing a picking. 
The function that was called from these actions, forced the creation a product movement when you scanned a wrong product.
As I explain in the PR, I think is better not doing nothing if the product scaned is not in the picking list.

pedrobaeza: I don't think so. We may need to add extra products in the picking. That's an intended feature.

pedrobaeza:
Closing as stated in https://github.com/OCA/stock-logistics-barcode/issues/463#issuecomment-1518604393

stock_barcodes/static/src/js/basic_controller.js
Bilbonet: OK, you're right 
Now it works fine. 
Thanks for the comments

PR #478, opened 2022-12-21
[14.0][FIX]computed field (qty_done) does not update

Bilbonet:
In this case the need is to pick out the products from a sale order. The operative is scan the code bar product and if it's correct picks it up.

As you can see in the image.
Wizard window don't update amount done while you are scanning. I have to server 4 units and I've already scanned two but windows still to 0.
If I click in the line opens de window with the correct information.

imagen

Even it would be good that the line change the color if the amount demanded is different that amount done.

### DIFF
--- a/stock_barcodes/wizard/stock_barcodes_read_todo.py
+++ b/stock_barcodes/wizard/stock_barcodes_read_todo.py
@@ -41,7 +41,7 @@ class WizStockBarcodesReadTodo(models.TransientModel):
     qty_done = fields.Float(
         "Done",
         digits="Product Unit of Measure",
-        readonly=False,
+        readonly=True,
         compute="_compute_qty_done",
         store=True,
     )


hieulucky111 (approved):
Code Review
LGTM

Bilbonet:
This is a simple change to fix a little problem. 
Can any maintainer review this?
Thanks

stock_barcodes/wizard/stock_barcodes_read_todo.py
```diff
@@ -41,7 +41,7 @@ class WizStockBarcodesReadTodo(models.TransientModel):
     qty_done = fields.Float(
         "Done",
         digits="Product Unit of Measure",
-        readonly=False,
+        readonly=True,
```
rousseldenis, line 44: @pedrobaeza Maybe indeed here a readonly=True is justified ?

pedrobaeza: But I don't see how changing the field to readonly makes that the wizard now shows the quantity field updated. That should happen the same being readonly or not.

Bilbonet: @pedrobaeza 
I thought because it's a computed field needed to be readonly = true
I made the change and it worked.

pedrobaeza: No, computed writable fields are readonly=False, and that's something intended for allowing people to change the value. I think you are connecting 2 unrelated things (or at least, not being the proper patch).

Bilbonet: Yes, it's possible. But I've tried right now and the behavior is like I've described. 
In the wizard quantity done is not update in the wizard view.
If I change the field with 'readonly = True', quantity done is update in the wizard view.

Maybe there's a better solution.

In v15 neither update 'cuantity_done' in the wizard view.

PR #493, opened 2023-02-10
[15.0][FIX] stock_barcodes: put in pack call method fails

### LINKED ISSUE #492: [15.0]stock_barcodes defect for Exception - AttributeError: 'stock.picking' has no attribute 'put.in.pack'
## Module
The name of the module that has a bug: stock_barcodes: https://github.com/OCA/stock-logistics-barcode/tree/14.0/stock_barcodes

## Describe the bug
Defect error message when we validate the Receipt Transfer after setting Done quantities from Barcode Scanner.

## To Reproduce
**Affected versions**: 15.0, and probably earlier versions.

1. Configure the Barcode Options with "Auto Put In Pack" [auto_put_in_pack] = true for Picking IN and set it for your Operation Type Receipt.
2. Scan your Receipt using barcode scanner, to set done quantities for products
3. Click "Validate" on the Receipt it will show the defect.

**Expected behavior**
If this configuration is enable from the Barcode Options, but we do not use Packs or Lots, the Users should see an error message.

**Additional context**
stock

Add any other context about the problem here. (e.g. OS, Python version, ...)

sergio-teruel:
cc @Tecnativa

ping @chienandalu @carlosdauden @nicolas-tg

### DIFF
--- a/stock_barcodes/models/stock_picking.py
+++ b/stock_barcodes/models/stock_picking.py
@@ -40,5 +40,5 @@ class StockPicking(models.Model):
             self.picking_type_id.barcode_option_group_id.auto_put_in_pack
             and not self.move_line_ids.mapped("result_package_id")
         ):
-            self.put_in_pack()
+            self.action_put_in_pack()
         return super().button_validate()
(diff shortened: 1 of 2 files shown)


chienandalu (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #494, opened 2023-02-14
[13.0][IMP] stock_barcodes: Remove lots from scan log for operations with check "use_create_lots" checked

sergio-teruel:
cc @Tecnativa TT41588

### DIFF
--- a/stock_barcodes/wizard/stock_barcodes_read_picking.py
+++ b/stock_barcodes/wizard/stock_barcodes_read_picking.py
@@ -5,7 +5,7 @@ import logging
 from odoo import _, api, fields, models
 from odoo.exceptions import ValidationError
 from odoo.fields import first
-from odoo.tools.float_utils import float_compare
+from odoo.tools.float_utils import float_compare, float_is_zero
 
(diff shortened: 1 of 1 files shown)


chienandalu (approved):

stock_barcodes/wizard/stock_barcodes_read_picking.py
```diff
@@ -601,6 +601,10 @@ def _prepare_scan_log_values(self, log_detail=False):
 
     def remove_scanning_log(self, scanning_log):
         for log in scanning_log:
+            if log.picking_id.picking_type_code == "incoming":
+                rounding = log.product_id.uom_po_id.rounding
```
pedrobaeza, line 605: An incoming picking can come from a return of an OUT, so PO UoM may not apply here.

pedrobaeza (commented):
@sergio-teruel

stock_barcodes/wizard/stock_barcodes_read_picking.py
pedrobaeza: @sergio-teruel can you check this?

PR #501, opened 2023-02-23
[14.0] product_multi_barcode: fix muti-company access right issue

sebastienbeau:
Context
In multi company
- add a barcode in company A (for a product that belong to company A)
- then add the same barcode in company B (for a product that belong to company B)

=> you have a access error message instead of a the error message that give the information of the duplicated product barcode.

### DIFF
--- a/product_multi_barcode/models/product_barcode.py
+++ b/product_multi_barcode/models/product_barcode.py
@@ -56,7 +56,19 @@ class ProductBarcode(models.Model):
                 [("id", "!=", record.id), ("name", "=", record.name)]
             )
             if barcodes:
+                # by default barcode 'shared' between all company (no ir.rule)
+                # so we may not have the access right on the product
+                # note: if you do not want to share the barcode between company
+                # you just need to add a custom ir.rule
+                product = barcodes[0].sudo().product_id
                 raise UserError(
-                    _('The Barcode "%s" already exists for product ' '"%s"')
-                    % (record.name, barcodes[0].product_id.name)
+                    _(
+                        'The Barcode "%(barcode_name)s" already exists for '
+                        'product "%(product_name)s" in the company %(company_name)s'
+                    )
+                    % dict(
+                        barcode_name=record.name,
+                        product_name=product.name,
+                        company_name=product.company_id.name,
+                    )
                 )


sebastienbeau:
@kev-roche

Kev-Roche (approved):
LGTM

aliciagaarzo (approved):

StefanRijnhart (approved):

product_multi_barcode/models/product_barcode.py
```diff
@@ -56,7 +56,19 @@ def _check_duplicates(self):
                 [("id", "!=", record.id), ("name", "=", record.name)]
             )
             if barcodes:
+                # by default barcode 'shared' between all company (no ir.rule)
```
rousseldenis, line 59: @sebastienbeau @StefanRijnhart 

Isn't the good solution to put an ir.rule on product.barcode model based on product instead of using sudo ???

rousseldenis:
```suggestion
                <record id="product_barcode_comp_rule" model="ir.rule">
                    <field name="name" >Product barcode multi-company</field>
                    <field name="model_id" ref="model_product_barcode"/>
                    <field name="domain_force"> ['|', ('product_id.company_id', 'in', company_ids), ('product_id.company_id', '=', False)]</field>
                </record>

```

StefanRijnhart: In core Odoo, the uniqueness check on the barcode works across companies, [using sudo](https://github.com/odoo/odoo/blob/16.0/addons/product/models/product_product.py#L178). It seems the most obvious choice for us to keep the same behaviour this module.

rousseldenis: The behaviour remains the same with the ir.rule, but without sudo, no ?

StefanRijnhart: With the rule, by default this module will allow duplicate barcodes as long as not more than one of them is set as the main barcode on the product. That seems inconsistent to me.

rousseldenis: Mmmmh yes, ok, I see.

rousseldenis (approved):

Issue #516, opened 2023-05-23
[14.0] barcodes_generator_product - Automatic generation does not work on existing products


francesco-ooops:
## Module
barcodes_generator_product

## Describe the bug
When adding a rule with automatic generation to a product created before the rule was set up, automatic barcode is not generated

## To Reproduce
**Affected versions**:

- [ ] 14.0
Other versions not checked

1. Enable automatic generation in demo data product rule

![image](https://github.com/OCA/stock-logistics-barcode/assets/55489873/b57a9c10-2835-45c3-952e-8b5c270b7d99)

2. Add rule to existing product
3. Barcode Base is not generated by sequence, and barcode is not generated

**Expected behavior**
If a rule is set to "automatic generation", when rule is added to product barcode should be automatically generated

PR #450, opened 2022-10-24
[16.0][MIG] barcodes_generator_product (from 14 to 16)

legalsylvain:
simple migration 
- bump version
- update fr translation
- add demo data
- use ``model_create_multi``
- improve product form display

product_template_manual_generation

PR #517, opened 2023-05-24
[14.0][FIX] barcodes_generator_product: #516

ilyasProgrammer:
fix https://github.com/OCA/stock-logistics-barcode/issues/516

### DIFF
--- a/barcodes_generator_abstract/models/barcode_generate_mixin.py
+++ b/barcodes_generator_abstract/models/barcode_generate_mixin.py
@@ -29,6 +29,18 @@ class BarcodeGenerateMixin(models.AbstractModel):
         related="barcode_rule_id.generate_type",
     )
 
+    def write(self, vals):
+        res = super().write(vals)
+        if vals.get("barcode_rule_id"):
+            rule = self.env["barcode.rule"].browse(vals["barcode_rule_id"])
+            if rule.generate_automate and rule.generate_type == "sequence":
+                for rec in self:
+                    if not rec.barcode_base:
+                        rec.generate_base()
+                    if not rec.barcode:
+                        rec.generate_barcode()
+        return res
+
     @api.model
     def create(self, vals):
         """It creates a new barcode if automation is active."""
@@ -36,7 +48,7 @@ class BarcodeGenerateMixin(models.AbstractModel):
         if barcode_rule.exists():
             vals.update({"barcode_rule_id": barcode_rule.id})
         record = super().create(vals)
-        if barcode_rule:
+        if barcode_rule and barcode_rule.generate_type == "sequence":
             record.generate_base()
             record.generate_barcode()
         return record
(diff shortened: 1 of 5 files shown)


legalsylvain (changes requested):
Hi. Thanks for the fix. Some remarks :   
- you should call generate barcode only if it is an automated rule.
- could you add test ?
- the current design (overload onchange) doesnt work on import. I think implement on create or write is a better design.
- could you implement the patch in the abstract module ? So it will work for all modules. (Lot / partner / ....)


Thanks !

ilyasProgrammer:
@legalsylvain Thanks for feedback. Please check new version.

barcodes_generator_abstract/models/barcode_generate_mixin.py
```diff
@@ -29,14 +29,25 @@ class BarcodeGenerateMixin(models.AbstractModel):
...
 
+    def write(self, vals):
+        res = super(BarcodeGenerateMixin, self).write(vals)
+        if vals.get("barcode_rule_id"):
+            rule = self.env["barcode.rule"].browse(vals["barcode_rule_id"])
+            if rule.generate_automate and rule.generate_type == "sequence":
+                for rec in self:
+                    if not rec.barcode:
+                        rec.generate_base()
+                        rec.generate_barcode()
```
legalsylvain:
```suggestion
                    if not rec.barcode_base:
                        rec.generate_base()
                     if not rec.barcode:
                        rec.generate_barcode()
```

(if base is defined, regenerate it is not required, don't you think ?)

barcodes_generator_abstract/models/barcode_generate_mixin.py
```diff
@@ -29,14 +29,25 @@ class BarcodeGenerateMixin(models.AbstractModel):
         related="barcode_rule_id.generate_type",
     )
 
+    def write(self, vals):
+        res = super(BarcodeGenerateMixin, self).write(vals)
```
legalsylvain:
```suggestion
        res = super().write(vals)
```

nitpicking : lighter syntax.

legalsylvain (commented):
Thanks a lot ! 
I think the new design is more robust, and covers more cases.

2 minors remarks inline.

legalsylvain (approved):
Thanks a lot ! 

/ocabot merge patch

Note : I migrated recently the modules to 16.0 version [here](https://github.com/OCA/stock-logistics-barcode/pull/450). if you can port this patch, it would be great!

francesco-ooops:
@legalsylvain shall we add you as maintainer for _abstract too?

francesco-ooops (approved):
Functional review ok!

legalsylvain:
I'm defined as the maintainers here : 
- https://github.com/OCA/stock-logistics-barcode/blob/16.0/barcodes_generator_abstract/__manifest__.py#L13
- https://github.com/OCA/stock-logistics-barcode/blob/16.0/barcodes_generator_product/__manifest__.py#L17

so no need to add it.

Don't understand why bot doesn't allow me to merge. 
Waiting for a repo maintainer.

francesco-ooops:
@legalsylvain but you're not for v14 ;)

PR #522, opened 2023-06-06
[IMP][16.0] barcodes_generator_product : various improvments 

legalsylvain:
Note to reviewers : better to review commit by commit.

- [x] [IMP] barcode_generator_product : Prevent user to create ``barcode.rule`` via product forms, to avoid to have barcode.rule not related to ``barcode.nomenclature``. (Such rule will not be accessible after, because there is no menu entry to access to barcode.rule
- [x] [IMP] barcode_generator_abstract : UX : set the field ``barcode_rule.generate_model`` required if generate_type is defined
- [x] [IMP] barcodes_generator_abstract : UX : Do not display ``barcode_rule.generate_automate`` for non 'sequence' generation
- [x] [IMP] barcodes_generator_abstract : UX : Adds a message to specify that sequence creation can be automatic.

image

- [x] [IMP] barcodes_generator_abstract : UX : display barcode.rule new field on barcode.nomenclature form view.

image

- [x] [REF] barcodes_generator_abstract : Refactor. 
    - Do not mix automatic generation and default value concepts
    - allow to have multiple automatic barcode.rule for a same model
    - Simplify a lot the code base
    - remove useless cache clearing

- [x] [REF] barcode_generate_abstract : Add more tests (Coverage +3,61%, see https://app.codecov.io/gh/OCA/stock-logistics-barcode/pull/522)

Before : 

image

After : 

image

- [x] [REF] Improve fields description, adding help
- [x] [REF] barcodes_generator_abstract : Use external_dependencies_override in setup file to correctly define python library


PaulGoubert (approved):
Thnak you Sylvain for those improvements

flalexg:
Hello @legalsylvain. I've been using this module and I detected an error. Taking advantage of the fact that you are improving this module, I will tell you about the error and see if it can be solved here or it is better to open an issue with the fix. 

In the `product_template` file we inherit the create function. In line 59 a `result |= template` is made. The thing is that if we try to make a copy of a product, by having the copy function the decorator `@api.returns('self', lambda value: value.id)`.
As you can see in the attached image, by returning multiple records in the create, we make the copy function not work. I don't know what would be the most effective way to solve this. What do you think?

Captura desde 2023-06-15 18-19-16

legalsylvain:
Hi @flalexg.

Could you take a look on the last commit https://github.com/OCA/stock-logistics-barcode/pull/522/commits/67dab9d43cff37e2220ec76cbe2dd27d41f919de, and realize test ? 

thanks for your feedback.

flalexg:
Hi @legalsylvain, functional test OK now! Thank you!

flalexg (approved):

PaulGoubert:
Hi @all.   Thnak you @flalexg for reporting this issue, and @legalsylvain for the fix

bealdav:
Thanks @legalsylvain I confirm than this fix remove a bug on copy record (product.template) at least.

I don't know how we get that. But it solve it

PaulGoubert (approved):

PR #525, opened 2023-06-16
[FIX] barcodes_generator_abstract : align create algorithm with write … algorigthm

legalsylvain:
Trivial fix introduced by https://github.com/OCA/stock-logistics-barcode/pull/522


legalsylvain:
fast tracking
/ocabot merge patch

PR #526, opened 2023-07-08
[14.0] [REF] simplify logic, get rid of duplicate company field

florentx:


francesco-ooops:
@legalsylvain what do you think of this change?

geomer198 (approved):
Code review ok!

florentx:
Just noticed that "company_id" is not saved on Product creation.
Here the fix with related tests.

product_barcode_constraint_per_company/models/product_product.py
```diff
@@ -2,24 +2,22 @@
...
 
 class ProductProduct(models.Model):
     _inherit = "product.product"
 
-    tmpl_company_id = fields.Many2one(
-        related="product_tmpl_id.company_id",
-        string="Company of the related Template",
-        store=True,
+    # Store related field
+    company_id = fields.Many2one(
```
rousseldenis: I would say, this could be in another module to avoid collisions

florentx: Maybe I don't see where is the risk of collision.

The add-on is only 10 lines of Python : it stores the related field which already exist in standard, and monkey patch the SQL constraint. Nothing more.

rousseldenis: @florentx I wanted to say the functionality of storing company_id on product.product level could be seen as independent feature (but base one for this).

francesco-ooops: @rousseldenis in which repo should that module be published in your opinion?

francesco-ooops:
@florentx can you apply requested modifications?

or would you like for us to pick up your work and bring it to completion? let us know! :)

product_barcode_constraint_per_company/models/product_product.py
francesco-ooops: > @florentx I wanted to say the functionality of storing company_id on product.product level could be seen as independent feature (but base one for this).

@rousseldenis @simahawk would https://github.com/OCA/product-variant work in your opinion?

Aldeigja: > @florentx I wanted to say the functionality of storing company_id on product.product level could be seen as independent feature (but base one for this).

Sorry, but I'm not sure I got your point here, if you mean that the code related to the functionality of storing company_id at product.product level should be put in a separate module, then @florentx  module would then only need to handle  _sql_constraints.
So for only ten lines of code we will create two modules, it seems a bit complicated.

@rousseldenis @simahawk Could you please elaborate on this further?

francesco-ooops: @simahawk ping :)

simahawk: I'd go for product-attribute but as for making it more complex than it is: it CAN happen.
IMO for now we can survive with a TODO and move on like this.

francesco-ooops: @simahawk ok we'll push a simple module with just those lines

legalsylvain (approved):
Lgtm

product_barcode_constraint_per_company/models/product_product.py
francesco-ooops: @florentx if you want to bring this PR to completion, the module suggested by @rousseldenis has been merged here https://github.com/OCA/product-attribute/pull/1405 ;)

legalsylvain (changes requested):
please depend on product_variant_company new module.

thanks !

product_barcode_constraint_per_company/models/product_product.py
francesco-ooops: @florentx please let us know if you have time to finish this PR, we're available to pick it up and bring it to completion if that's ok for you :)

florentx:
Changed to depend on `product_variant_company`

Test fails due to incompatibility with `_check_duplicates` in add-on `product_multi_barcode`, file `product_multi_barcode/models/product_barcode.py`

francesco-ooops:
@florentx will you keep working on this or shall we pick it up?

florentx:
> @florentx will you keep working on this or shall we pick it up?

You can take over this PR and bring it to completion @francesco-ooops . Thanks

PR #528, opened 2023-07-25
[14.0][ADD] sale_barcode: add module

dessanhemrayev:
This module adds the product to the sales order line according to the barcode information provided by the barcode/datamatrix device.


dessanhemrayev:
Create lot and/or select lot if used in combination with sale_order_lot_selection feature is not implemented yet

sale_barcode/models/sale_order.py
```diff
@@ -0,0 +1,20 @@
...
+        product = self._product_barcode_search(barcode)
+        if not product:
+            raise UserError(_("There is no product with such a barcode"))
+        self.order_line = [(0, 0, {"product_id": product.id})]
+        return {"type": "ir.actions.act_window_close"}
+
+    @api.model
+    def _product_barcode_search(self, barcode):
+        """Returns the found product by barcode"""
+        return self.env["product.product"].search([("barcode", "=", barcode)], limit=1)
```
aayartsev: Do we need to write limit=1 when we have same constraint in model?

aayartsev (changes requested):

sale_barcode/models/sale_order.py
dessanhemrayev: Thanks) I'll fix it

aayartsev (approved):

aayartsev (approved):
It looks good

PR #531, opened 2023-08-11
[FIX] product_multi_ean: Archived Product prevents creating new EAN

SirTakobi:
Rebooting https://github.com/OCA/stock-logistics-barcode/pull/427 because it went stale and I can't reopen it.


mrcast (approved):
Functional test, ok

PR #535, opened 2023-10-17
[16.0][FIX] barcodes_generator_abstract: get the real integer code for all coding types

RodrigoBM:
if you use for example the gs1-128 coding, this coding cannot be saved in odoo directly.

> import barcode
> EAN = barcode.get_barcode_class('gs1_128')
> str(EAN('0000'))
> 'ñ0000'

We have to call the function get_fullcode which is a code that we can save in Odoo.

> import barcode
>  EAN = barcode.get_barcode_class('gs1_128')
>  str(EAN('0000').get_fullcode())
> '0000'


legalsylvain (approved):
Not expert of gs1-128. (only use ean13 barcode). But looks OK ! 

thanks for your contribution !

xAdrianCif (approved):
Correct and respectful on the current flow.

Looks great! thanks @RodrigoBM

PR #537, opened 2023-10-21
[14.0][IMP] product_barcode_constraint_per_company: simplify logic

dessanhemrayev:
Change old method
Update oca_dependencies file
Fix tests


product_barcode_constraint_per_company/__manifest__.py
```diff
@@ -1,18 +1,17 @@
...
+# Copyright 2023 Cetmix OÜ
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 {
     "name": "Product Barcode Constraint per Company",
     "version": "14.0.1.0.1",
     "category": "Product",
-    "summary": "Change the product barcode constraint"
-    ", allowing the same barcode for differents companies",
-    "author": "GRAP, Odoo Community Association (OCA)",
```
legalsylvain, line 11: could you preserve authorship, and add Cetmix instead of replacing ? thanks !

dessanhemrayev: okay, thanks

legalsylvain (approved):

product_barcode_constraint_per_company/README.rst
```diff
@@ -1,90 +0,0 @@
-======================================
```
legalsylvain: No need to remove this file. It générate useless diff.

product_barcode_constraint_per_company/__manifest__.py
```diff
@@ -1,18 +1,18 @@
...
+    "summary": """Change the product barcode constraint,
+                  allowing the same barcode for differents companies""",
+    "author": "GRAP, Cetmix, Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/stock-logistics-barcode",
     "license": "AGPL-3",
     "depends": [
-        "product",
+        "product_variant_company",
+        "product_multi_barcode",
```
legalsylvain: This new dependency is more problematic.

legalsylvain (changes requested):
Do not add new dependency to multi barcode module.

dessanhemrayev:
> Do not add new dependency to multi barcode module.

@legalsylvain   Do you think we need to create a bridge module for that, or is there possibly another solution?

legalsylvain:
multi barcode module is a module that is really changing a lot of things. The feature is not desired in some implementation. So in that case, indeed glue module looks a good solution, if you want to make the module product_barcode_constraint_per_company working with that module.

francesco-ooops:
supersedes #526

legalsylvain (commented):
except the deletion of the README.rst file, LGTM. Thanks !

legalsylvain (commented):
nitpicking : still changes in product_barcode_constraint_per_company/README.rst ! 

No need to change this file, it will be generated automatically, when mergin the PR, by the bot.

regards.

product_barcode_constraint_per_company/tests/test_module.py
```diff
@@ -1,37 +1,46 @@
...
+from odoo import _
+from odoo.tests.common import SavepointCase, tagged
 from odoo.tools.misc import mute_logger
 
 
-class TestModule(TransactionCase):
-    def setUp(self):
-        super().setUp()
+@tagged("post_install", "-at_install")
+class TestModule(SavepointCase):
```
geomer198: Please add assertion message for all assert cases and docstring for all tests functions

product_multi_barcode_constraint_per_company/models/product_barcode.py
```diff
@@ -0,0 +1,21 @@
...
+from odoo import api, models
+
+
+class ProductBarcode(models.Model):
+    _inherit = "product.barcode"
+
+    @api.constrains("name")
+    def _check_duplicates(self):
+        for record in self:
+            barcodes = self.search(
```
geomer198: Why you use search function here?
And why you call `super` in cycle?

product_multi_barcode_constraint_per_company/tests/test_product_barcode_constraint.py
```diff
@@ -0,0 +1,84 @@
...
+        # and expect an IntegrityError
+        with mute_logger("odoo.sql_db"), self.assertRaises(
+            IntegrityError,
+            msg="A barcode can only be assigned to one product per company !",
+        ):
+            self._create_product("Product 2", self.company_1)
+
+    def test_create_different_company(self):
+        """Verify creating a product with the same barcode for different companies"""
+        super().test_create_different_company()
```
geomer198: You deactivate parent function. Why do you use `super` here?

product_multi_barcode_constraint_per_company/tests/test_product_barcode_constraint.py
```diff
@@ -0,0 +1,84 @@
...
+            result,
+            types.FunctionType,
+            msg="Expecting a FunctionType result, but got {result_type}".format(
+                result_type=type(result)
+            ),
+        )
+
+    def test_create_same_company(self):
+        """Verifying the Existence of a Product with the Same Barcode within a Single Company"""
+        super().test_create_same_company()
```
geomer198: You deactivate parent function. Why do you use `super` here?

geomer198 (changes requested):
Please fix comments

product_multi_barcode_constraint_per_company/tests/test_product_barcode_constraint.py
dessanhemrayev: Thanks, I'll fix it

product_multi_barcode_constraint_per_company/tests/test_product_barcode_constraint.py
dessanhemrayev: Thanks, I'll fix it

product_multi_barcode_constraint_per_company/models/product_barcode.py
dessanhemrayev: Thanks, I'll fix it

product_multi_barcode_constraint_per_company/models/product_barcode.py
```diff
@@ -0,0 +1,35 @@
...
+from odoo import _, api, models
+from odoo.exceptions import UserError
+
+
+class ProductBarcode(models.Model):
+    _inherit = "product.barcode"
+
+    @api.constrains("name")
+    def _check_duplicates(self):
+        barcodes = self.search(
```
geomer198, line 13: For what you write `search` here?

product_multi_barcode_constraint_per_company/models/product_barcode.py
```diff
@@ -0,0 +1,35 @@
...
+            [("id", "not in", self.ids), ("name", "in", self.mapped("name"))]
+        )
+
+        for barcode in barcodes:
+            for record in self:
```
geomer198: Cycle into cycle it's bad solution, please leave only one cycle.

product_multi_barcode_constraint_per_company/models/product_barcode.py
```diff
@@ -0,0 +1,35 @@
...
+                if (
+                    barcode
+                    and barcode.sudo().product_tmpl_id.company_id.id
+                    == record.product_tmpl_id.company_id.id
+                ):
+                    product = barcode.sudo().product_id
+                    raise UserError(
```
geomer198: Why do you removed `super`?

product_multi_barcode_constraint_per_company/models/product_barcode.py
```diff
@@ -0,0 +1,35 @@
...
+                        _(
+                            'The Barcode "%(barcode_name)s" already exists for '
+                            'product "%(product_name)s" in the company %(company_name)s'
+                        )
+                        % dict(
```
geomer198: `dict()` slowly than `{}`, so better use `{}` in code.

geomer198 (changes requested):
@dessanhemrayev Please, fix my comments.

product_multi_barcode_constraint_per_company/models/product_barcode.py
dessanhemrayev: okay, thanks

product_multi_barcode_constraint_per_company/models/product_barcode.py
dessanhemrayev: okay, I'll fix it

product_multi_barcode_constraint_per_company/models/product_barcode.py
dessanhemrayev: okay, I'll fix it

product_multi_barcode_constraint_per_company/models/product_barcode.py
```diff
@@ -0,0 +1,25 @@
...
+
+from odoo import api, models
+
+
+class ProductBarcode(models.Model):
+    _inherit = "product.barcode"
+
+    @api.constrains("name")
+    def _check_duplicates(self):
+        existing_codes = False
```
geomer198: Please change like this:
```py
@api.constrains("name")
    def _check_duplicates(self):
        existing_codes = self.env["product.barcode"]
        for record in self:
            barcode = self.search(
                [("id", "!=", record.id), ("name", "=", record.name)], limit=1
            )
            if (
                barcode
                and barcode.sudo().product_tmpl_id.company_id.id
                == record.product_tmpl_id.company_id.id
            ):
                existing_codes |= record
                break
        if existing_codes:
            super(ProductBarcode, existing_codes)._check_duplicates()
```

geomer198 (changes requested):
Please fix my comment

geomer198 (approved):
LGTM!

geomer198:
@legalsylvain Could you please check changes?

legalsylvain (approved):
Partial review . (product_barcode_constraint_per_company only)
LGTM. Thanks for work !

francesco-ooops:
@florentx good for you?

francesco-ooops:
@OCA/logistics-maintainers good for merge?

product_multi_barcode_constraint_per_company/__init__.py
```diff
@@ -0,0 +1,15 @@
+from . import models
+
+from odoo.addons.product_barcode_constraint_per_company.tests.test_module import (
+    TestModule,
+)
+import unittest
+
+
+@unittest.skip
```
simahawk: what is this?

product_barcode_constraint_per_company/tests/test_module.py
```diff
@@ -1,37 +1,82 @@
-# Copyright (C) 2019 - Today: GRAP (http://www.grap.coop)
```
simahawk: there's no init file here, these tests won't run at all

product_barcode_constraint_per_company/tests/test_module.py
```diff
@@ -1,37 +1,82 @@
...
+# Copyright 2023 Cetmix OÜ
```
simahawk: 2023 comes after :wink:

product_multi_barcode_constraint_per_company/tests/test_product_barcode_constraint.py
```diff
@@ -0,0 +1,82 @@
...
+)
+
+from .. import void
+
+
+@tagged("post_install", "-at_install")
+class TestProductBarcodeConstraintInherit(TestModule):
+    def test_void(self):
+        """Verify the type of the result of the void method call."""
+        result = void(None)
```
simahawk: I don't get what is this void thing

simahawk (changes requested):

oca_dependencies.txt
```diff
@@ -1 +1,2 @@
 # See https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#oca_dependencies-txt
+product-attribute https://github.com/OCA/product-attribute 14.0
```
simahawk: FTR this file is useless for OCA builds since a while (unless you rely on it for your CI)

product_multi_barcode_constraint_per_company/__init__.py
dessanhemrayev: to skip tests that don't work correctly without a new module

product_barcode_constraint_per_company/tests/test_module.py
dessanhemrayev: I don't understand you, what did you mean here?

oca_dependencies.txt
dessanhemrayev: Thanks, I'll fix it

product_multi_barcode_constraint_per_company/tests/test_product_barcode_constraint.py
dessanhemrayev: Codecov requires that this void be covered with a test

product_barcode_constraint_per_company/tests/test_module.py
dessanhemrayev: Please check carefully, the init file is in the repository. It is not displayed because this file is immutable
image

francesco-ooops:
@simahawk better now?

product_barcode_constraint_per_company/tests/test_module.py
simahawk: You should not change the order. Leave the original (c) line on top

product_barcode_constraint_per_company/tests/test_module.py
simahawk: sorry, I mixed w/ the other init file in this PR :v:

product_multi_barcode_constraint_per_company/__init__.py
simahawk: I still don't understand why you need this. Can you please elaborate and explain why it is needed?
BTW you should not import here any test stuff.

simahawk (changes requested):

product_multi_barcode_constraint_per_company/__init__.py
dessanhemrayev: Without disabling the tests, there is no point in checking the functionality of the modules. 
I disabled the tests of the module `product_barcode_constraint_per_company` because its tests do not work correctly after we changed the module itself, that is, we moved the company_id field to a separate module, and changed the verification method. now, after we have made changes, we run the tests, they crash, the tests are due to the module `product_multi_barcode`. The module `product_multi_barcode` does not provide for the situation that one barcode can be in different companies. Because of this, we had to write a new bridge module. In which all these situations are fixed.

product_barcode_constraint_per_company/tests/test_module.py
dessanhemrayev: Thanks, I fix it)

product_multi_barcode_constraint_per_company/__init__.py
francesco-ooops: @simahawk makes sense now?

francesco-ooops:
@simahawk good to go?

product_barcode_constraint_per_company/tests/test_module.py
```diff
@@ -1,37 +1,81 @@
...
 
-from odoo.tests.common import TransactionCase
+from odoo.tests.common import SavepointCase, tagged
 from odoo.tools.misc import mute_logger
 
 
-class TestModule(TransactionCase):
-    def setUp(self):
-        super().setUp()
+@tagged("post_install", "-at_install")
```
HekkiMelody: **todo**: it's not necessary to tag these tests as `post_install`. Please remove

product_barcode_constraint_per_company/tests/test_module.py
```diff
@@ -1,37 +1,81 @@
-# Copyright (C) 2019 - Today: GRAP (http://www.grap.coop)
```
HekkiMelody: **todo**: Please restore this to its original value

product_multi_barcode_constraint_per_company/models/product_barcode.py
```diff
@@ -0,0 +1,25 @@
...
+            )
+            if (
+                barcode
+                and barcode.sudo().product_tmpl_id.company_id.id
+                == record.product_tmpl_id.company_id.id
+            ):
+                existing_codes |= record
+                break
+        if existing_codes:
+            super(ProductBarcode, existing_codes)._check_duplicates()
```
HekkiMelody: **question**: By default `product.barcode` doesn't have a `company_id`, therefore it's visible to everyone from every company. Does it cause issues if, for example, people can see multiple of the same barcode, but can't necessarily see the related products?

Should barcode also have a `company_id` field (maybe a related field, from `product_tmpl_id.company_id`?)

In the case of the constraint, I think it may be better to just override it. If `company_id` is False, there can be no other barcode with the same name. If `company_id` has a value, it has to be unique per company.  Please add a test for the case where `company_id` is False.

product_multi_barcode_constraint_per_company/README.rst
```diff
@@ -0,0 +1,90 @@
...
+
+Do not contact contributors directly about support or help with technical issues.
+
+Credits
+=======
+
+Authors
+~~~~~~~
+
+* GRAP
```
HekkiMelody: **todo**: Pre-commit should have regenerated this README (and also the `index.html`) automatically, I'm not sure why it didn't do that. Can you check?

product_multi_barcode_constraint_per_company/tests/test_product_barcode_constraint.py
HekkiMelody: I agree with @simahawk , I don't think this is appropriate or necessary

product_multi_barcode_constraint_per_company/__init__.py
HekkiMelody: I don't think this is necessary.

In OCA tests are run only once, on module installation. If your new module breaks the old module's tests, that shouldn't be a problem. It's normal for your module to change the behaviour of a related module, but the tests are executed in installation order.

Everything in this file (except the first line) should be removed IMO.

product_multi_barcode_constraint_per_company/tests/test_product_barcode_constraint.py
```diff
@@ -0,0 +1,82 @@
...
+
+from odoo.addons.product_barcode_constraint_per_company.tests.test_module import (
+    TestModule,
+)
+
+from .. import void
+
+
+@tagged("post_install", "-at_install")
+class TestProductBarcodeConstraintInherit(TestModule):
```
HekkiMelody: If you want to inherit the setup from your other tests, but not the tests themselves, you should separate the setup function into its own class so you can inherit from it separately.

Here's an example: https://github.com/OCA/stock-logistics-workflow/blob/14.0/stock_picking_auto_create_lot/tests/test_stock_picking_auto_create_lot.py#L10

HekkiMelody (changes requested):
This PR is definitely not very clean. I think the general concept is sound, but there are some issues to fix.

product_multi_barcode_constraint_per_company/tests/test_product_barcode_constraint.py
```diff
@@ -0,0 +1,82 @@
...
+            msg="Expecting a FunctionType result, but got {result_type}".format(
+                result_type=type(result)
+            ),
+        )
+
+    def test_create_same_company(self):
+        """Verifying the Existence of a Product with the Same Barcode within a Single Company"""
+        # Create the first product in company_1
+
+        product_1 = self._create_product("Product 1", self.company_1)
```
HekkiMelody: **issue**: These tests should use the new `product.barcode` model more extensively. This includes checking a product that has two barcodes, for example.

product_multi_barcode_constraint_per_company/tests/test_product_barcode_constraint.py
dessanhemrayev: Thanks for the advice, I will definitely use it.

product_multi_barcode_constraint_per_company/README.rst
dessanhemrayev: It may be necessary to update the pre-commit

product_barcode_constraint_per_company/tests/test_module.py
dessanhemrayev: ok, I'll fix it

product_barcode_constraint_per_company/tests/test_module.py
dessanhemrayev: ok, I'll fix it

product_barcode_constraint_per_company/__manifest__.py
dessanhemrayev: ok, I'll fix it)

product_multi_barcode_constraint_per_company/__init__.py
dessanhemrayev: okay, I'll fix it

product_multi_barcode_constraint_per_company/models/product_barcode.py
dessanhemrayev: Thanks for the advice, I will definitely use it.

product_multi_barcode_constraint_per_company/tests/test_product_barcode_constraint.py
dessanhemrayev: okay,  I will definitely add a test for it.

product_barcode_constraint_per_company/tests/test_module.py
```diff
@@ -3,35 +3,62 @@
...
-    def setUp(self):
-        super().setUp()
 
-        self.ResCompany = self.env["res.company"]
-        self.ProductProduct = self.env["product.product"]
-        self.company_1 = self.ResCompany.create({"name": "Company 1"})
-        self.company_2 = self.ResCompany.create({"name": "Company 2"})
-
-    # Test Section
+@tagged("at_install")
```
HekkiMelody: **issue**: I noticed that tests broke because of the order in which modules are installed and tested in the OCA CI.

1. product_multi_barcode
2. product_barcode_constraint_per_company --> tests break here
3. product_multi_barcode_constraint_per_company

The tests break, predictably, because the glue module hasn't been installed yet. Because of this, I think in this specific circumstance it may be acceptable to tag these tests as `post_install`, with a comment explaining why it is being done this way.

product_multi_barcode_constraint_per_company/models/product_barcode.py
```diff
@@ -0,0 +1,25 @@
...
+        for record in self:
+            barcode = self.search(
+                [("id", "!=", record.id), ("name", "=", record.name)], limit=1
+            )
+            if (
+                barcode
+                and barcode.sudo().product_id.company_id.id
+                == record.product_id.company_id.id
+                or not record.product_id.company_id
+            ):
```
HekkiMelody: **chore**: you don't need to access the `id` here. Also, just to really make sure that the order of operations is clear and unambiguous, I think I'd prefer more brackets

```suggestion
                barcode
                and (not record.product_id.company_id or
                    (barcode.sudo().product_id.company_id == record.product_id.company_id)
                )
            ):
```

product_multi_barcode_constraint_per_company/tests/test_product_barcode_constraint.py
```diff
@@ -0,0 +1,148 @@
...
+from odoo.exceptions import UserError
+from odoo.tests import tagged
+from odoo.tools.misc import mute_logger
+
+from odoo.addons.product_barcode_constraint_per_company.tests.common import (
+    CommonProductBarcodeConstraintPerCompany,
+)
+
+
+@tagged("at_install")
```
HekkiMelody: **chore**: this is the default; IMO this line is not necessary

product_multi_barcode_constraint_per_company/tests/test_product_barcode_constraint.py
```diff
@@ -0,0 +1,148 @@
...
+            self.product_test_2.barcode_ids = [
+                (0, 0, {"name": self.valid_barcode2_1}),
+                (0, 0, {"name": self.valid_barcode_1}),
+            ]
+
+        with mute_logger("odoo.sql_db"), self.assertRaises(UserError, msg=message):
+            self.product_test_2.barcode_ids = [
+                (0, 0, {"name": self.valid_barcode_1}),
+                (0, 0, {"name": self.valid_barcode2_1}),
+            ]
```
HekkiMelody: (non-blocking)

In case you didn't know, you can decorate the entire test method with `mute_logger("odoo.sql_db")` so you don't have to specify it every time here.

HekkiMelody (changes requested):
Code review, just a couple of minor issues IMO. Much better than before.

**question(non-blocking)**: not sure about the use of newsfragments in OCA anymore. In this case, what would `2944` refer to?

product_barcode_constraint_per_company/tests/test_module.py
dessanhemrayev: Thank you so much for your help, this advice helped me fix this problem with the tests.

product_multi_barcode_constraint_per_company/tests/test_product_barcode_constraint.py
dessanhemrayev: Thanks, I'll fix it

product_multi_barcode_constraint_per_company/tests/test_product_barcode_constraint.py
dessanhemrayev: I'll fix it

HekkiMelody (approved):
Code review, LGTM

francesco-ooops:
@simahawk good for you?

francesco-ooops:
@simahawk waiting for your blessing :)

francesco-ooops:
@simahawk help me Orsi-Wan Kenobi, you're my only hope

francesco-ooops:
@pedrobaeza can this be merged based on existing reviews?

product_barcode_constraint_per_company/README.rst
```diff
@@ -87,4 +87,4 @@ promote its widespread use.
 
 This module is part of the `OCA/stock-logistics-barcode <https://github.com/OCA/stock-logistics-barcode/tree/14.0/product_barcode_constraint_per_company>`_ project on GitHub.
 
-You are welcome to contribute. To learn how please visit https://odoo-community.org/page/Contribute.
+You are welcome to contribute. To learn how please visit https://odoo-community.org/page/Contribute.
```
pedrobaeza: Remove this change.

product_barcode_constraint_per_company/__manifest__.py
```diff
@@ -6,13 +6,13 @@
     "name": "Product Barcode Constraint per Company",
     "version": "14.0.1.0.1",
     "category": "Product",
-    "summary": "Change the product barcode constraint"
-    ", allowing the same barcode for differents companies",
+    "summary": """Change the product barcode constraint,
+                  allowing the same barcode for differents companies""",
```
pedrobaeza: Don't change this. It was correct. Now, it isn't, as the spaces are included.

product_barcode_constraint_per_company/__manifest__.py
```diff
@@ -6,13 +6,13 @@
...
     "author": "GRAP, Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/stock-logistics-barcode",
     "license": "AGPL-3",
     "depends": [
-        "product",
+        "product_variant_company",
```
pedrobaeza, line 15: You mustn't change dependencies in a stable module.

pedrobaeza (changes requested):

product_barcode_constraint_per_company/__manifest__.py
francesco-ooops: Hi @pedrobaeza , AFAICS this is not a stable module, and the change in dependency was asked here: https://github.com/OCA/stock-logistics-barcode/pull/526#pullrequestreview-1658253604 following https://github.com/OCA/stock-logistics-barcode/pull/526#discussion_r1275978467

pedrobaeza: OK then about this.

product_barcode_constraint_per_company/__manifest__.py
```diff
@@ -6,13 +6,13 @@
     "name": "Product Barcode Constraint per Company",
     "version": "14.0.1.0.1",
```
simahawk: Detail: as you added a new module in the same PR you should change the version manually here, unless you don't care that the new module will have version 1.1.0 too when we'll bump w/ `minor`.

product_multi_barcode_constraint_per_company/models/product_barcode.py
```diff
@@ -0,0 +1,25 @@
...
+    def _check_duplicates(self):
+        existing_codes = self.env["product.barcode"]
+        for record in self:
+            barcode = self.search(
+                [("id", "!=", record.id), ("name", "=", record.name)], limit=1
+            )
+            if barcode and (
+                not record.product_id.company_id
+                or (
+                    barcode.sudo().product_id.company_id == record.product_id.company_id
```
simahawk: I'd move `sudo` to `for record in self.sudo()`

simahawk (commented):
pls check Pedro's remarks

product_barcode_constraint_per_company/__manifest__.py
dessanhemrayev: Thanks for review!)
I'll fix it

product_multi_barcode_constraint_per_company/models/product_barcode.py
dessanhemrayev: Thanks for review!)
I'll fix it

francesco-ooops:
@pedrobaeza @simahawk kind reminder :)

product_barcode_constraint_per_company/README.rst
simahawk: still here

simahawk (approved):
minor remark, pre-approving

dessanhemrayev:
@pedrobaeza  please check, I have corrected your comments)
Thanks

pedrobaeza (approved):
/ocabot merge nobump

PR #541, opened 2023-10-24
[16.0][ADD] product_packaging_multi_barcode_stock_menu: view packaging on barcode list

santostelmo:
Show  `packaging` field in  product barcode tree view


cyrilmanuel (approved):

sebalix (changes requested):
Cannot do that, `product.barcode.packaging_id` field is part of `product_packaging_multi_barcode` module, so this doesn't respect the inheritance.

santostelmo:
> product_packaging_multi_barcode

@sebalix Should I make `product_multi_barcode_stock_menu` dependent of `product_packaging_multi_barcode` ?

sebalix:
@santostelmo no we cannot do that neither sadly. I think the way it has been designed, we have to create a `product_packaging_multi_barcode_stock_menu` module to handle this need.
I know its overkill but I do not see another solution.

In the long term, as both product and packaging are part of Odoo std (`product` module), I'm wondering why `product_multi_barcode` and `product_packaging_multi_barcode` couldn't be merge in one module (you install it and you get all your models coming from  `product` extended with this multi barcodes feature), so we could have only one extra module that make the glue  between `product_multi_barcode` with `stock` (without restricting this to a menu, so a better name would have been `product_multi_barcode_stock` for instance).
But I don't know these modules nor why it has been designed this way, so for now we have to stick with what we have.

cc @simahawk

santostelmo:
> @santostelmo no we cannot do that neither sadly. I think the way it has been designed, we have to create a `product_packaging_multi_barcode_stock_menu` module to handle this need. I know its overkill but I do not see another solution.
> 
> In the long term, as both product and packaging are part of Odoo std (`product` module), I'm wondering why `product_multi_barcode` and `product_packaging_multi_barcode` couldn't be merge in one module (you install it and you get all your models coming from `product` extended with this multi barcodes feature), so we could have only one extra module that make the glue between `product_multi_barcode` with `stock` (without restricting this to a menu, so a better name would have been `product_multi_barcode_stock` for instance). But I don't know these modules nor why it has been designed this way, so for now we have to stick with what we have.
> 
> cc @simahawk

@sebalix I have created the module `product_packaging_multi_barcode_stock_menu`

product_packaging_multi_barcode_stock_menu/__manifest__.py
```diff
@@ -0,0 +1,14 @@
+{
+    "name": "Multiple barcodes menu show product packaging",
+    "version": "16.0.1.0.0",
+    "license": "AGPL-3",
+    "author": "Camptomcamp, Odoo Community Association (OCA)",
+    "category": "Product Management",
+    "depends": ["product_multi_barcode_stock_menu"],
```
sebalix: Need to add `product_packaging_multi_barcode` as well here

product_packaging_multi_barcode/README.rst
```diff
@@ -7,7 +7,7 @@ Multiple barcodes on product packagings
    !! This file is generated by oca-gen-addon-readme !!
    !! changes will be overwritten.                   !!
    !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
-   !! source digest: sha256:cbad2b941a855012dc1379aaf1de51fedce967a4ccc05ec8e8c4923b2e2502a1
+   !! source digest: sha256:fe1979a76bb20948f63a3bd26d33f088a2ff8fafdf1197d1bf468ff1e3ffbe22
```
sebalix: Can't these pre-commit changes on `product_packaging_multi_barcode` be canceled?

sebalix (commented):
Thank you, would like to hear another opinion about this though, as I do not know these modules a lot

QuocDuong1306:
Ping @santostelmo , Could you help to fix this?

product_packaging_multi_barcode_stock_menu/__manifest__.py
santostelmo: @sebalix dependency was added.

santostelmo:
@QuocDuong1306 I fixed the points mentioned above. @sebalix could you please check your change requests ?

sebalix (approved):
LG, can you add (in a separate commit) a ROADMAP.rst entry in `product_packaging_multi_barcode` module to merge it in `product_multi_barcode` for the next migration, and same for `product_packaging_multi_barcode_stock_menu` + `product_multi_barcode_stock_menu` to merge them in a new module `product_multi_barcode_stock`.

At the end we'll get only two modules, one depending exclusively on `product`, the other one being a glue module extending the first by adding a dependency on `stock`.

cyrilmanuel (approved):
Not an expert but LGTM

product_packaging_multi_barcode_stock_menu/readme/ROADMAP.rst
```diff
@@ -0,0 +1 @@
+In future migration that should be merged in a new module `product_multi_barcode_stock`
```
simahawk, line 1: why not renaming this module w/o `_menu` right now? :thinking:

cyrilmanuel: done

cyrilmanuel:
missing script on pre to change module name if its already installed

product_packaging_multi_barcode/__manifest__.py
```diff
@@ -6,7 +6,7 @@
     "license": "AGPL-3",
     "author": "Camptocamp, Odoo Community Association (OCA)",
     "category": "Product Management",
-    "depends": ["product_multi_barcode", "product_multi_barcode_stock_menu"],
+    "depends": ["product_multi_barcode", "product_multi_barcode_stock"],
```
simahawk, line 9: we need a manual version change here as we'll have to merge the PR w/ `nobump`

simahawk (commented):
@ilyasProgrammer ping

simahawk:
@santostelmo this PR deserves a better description explaining the changes :wink:

cyrilmanuel (approved):

cyrilmanuel:
hi @simahawk we can go on with this pr. i've fix error on pre-commit by updating the branch 16 template. 

all good :)

cyrilmanuel:
@OCA/logistics-maintainers could you merge this pr please :)

product_multi_barcode_stock/readme/ROADMAP.rst
```diff
@@ -0,0 +1 @@
+In future migration that should be merged in a new module `product_multi_barcode_stock`
```
jbaudoux: Is this still relevant?
It looks like the modules refactoring has been done

jbaudoux (changes requested):
The renaming of the module `product_multi_barcode_stock_menu` into `product_multi_barcode_stock` doesn't bring value. And you make a new module  `product_packaging_multi_barcode_stock_menu` where you keep `menu` in the name which is inconsistent. That renaming should have been suggested during migration. Nevertheless, that renaming is not usefull.

Drop the renaming of the existing module and adapt the dependency of `product_packaging_multi_barcode_stock_menu`. Then it should be good.

simahawk:
@jbaudoux you should read previous discussions ;)

jbaudoux:
> @jbaudoux you should read previous discussions ;)

I did :p

There is a new ROADMAP file suggesting to rename the module but the module is also renamed which doesn't make sense. All this should be reverted.

PR #542, opened 2023-10-24
[16.0][IMP] product_packaging_multi_barcode: barcode product show packaging

santostelmo:
Improvement to show packaging and create date on the product barcodes

### DIFF
--- a/product_packaging_multi_barcode/views/product_template_view.xml
+++ b/product_packaging_multi_barcode/views/product_template_view.xml
@@ -1,0 +1,20 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="product_template_form_view_multi_barcode" model="ir.ui.view">
+        <field name="name">product.template.form.multi.barcode.inherited</field>
+        <field name="model">product.template</field>
+        <field
+            name="inherit_id"
+            ref="product_multi_barcode.product_template_form_view_multi_barcode"
+        />
+        <field name="arch" type="xml">
+            <xpath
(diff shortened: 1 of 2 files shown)


cyrilmanuel (approved):

product_packaging_multi_barcode/views/product_template_view.xml
```diff
@@ -0,0 +1,20 @@
...
+            name="inherit_id"
+            ref="product_multi_barcode.product_template_form_view_multi_barcode"
+        />
+        <field name="arch" type="xml">
+            <xpath
+                expr="//field[@name='barcode_ids']//field[@name='name']"
+                position="after"
+            >
+                <field name="packaging_id" />
+                <field name="create_date" />
```
sebalix: Why `create_date` here?
OK if there is a generic need to add it here, perhaps put it as `optional="hide"`.
But if this is only to fulfill a specific need, we'll have to put this in a specific module.

leemannd (approved):

product_packaging_multi_barcode/views/product_template_view.xml
santostelmo: Our client needs to know the create date of the barecode. Not sure if it can be something generic. So I think we could provide that as an option `optional="hide"`

sebalix (approved):

product_packaging_multi_barcode/views/product_template_view.xml
santostelmo: Changed

cyrilmanuel:
@simahawk ping :)

PR #544, opened 2023-10-26
product_multi_barcode: add missing index

sebastienbeau:
Add the missing index, that can cause perf issue when having a huge database


sebastienbeau:
@Kev-Roche @bealdav

bealdav (approved):

rvalyi (approved):

PR #545, opened 2023-10-26
product_multi_barcode: remove useless costly dependency

sebastienbeau:
When having a product with a lot of variant
if you try to remove some product.attribute.value
it will remove the related product.product
and everytime a product.product is remove
this will trigger the computed field
On my db removing a product.product take 10s without this change and 3s after

A test have been added to check that this change to not have impact when creating a product_template with a barcode

@kev-Roche


legalsylvain:
Thanks !

Please take a look on precommit.

rousseldenis (approved):
LGTM. @sebastienbeau Could you rebase as soon as #577 is merged?

PR #551, opened 2023-10-31
[15.0][IMP] stock_barcodes: Filled all fields if they is defined in options as filled default.

sergio-teruel:
-- Determine scan step after determine todo action

cc @Tecnativa

ping @carlosdauden @CarlosRoca13 @dalonsod

PR #557, opened 2023-11-11
[15.0][IMP] stock_barcodes_gs1: Add hook process gs1 parsed values by other modules

sergio-teruel:
cc @Tecnativa

Depends on:
- [x] https://github.com/OCA/stock-logistics-barcode/pull/551

ping @carlosdauden @CarlosRoca13 @dalonsod

### DIFF
--- a/stock_barcodes_gs1/tests/test_stock_barcodes_gs1.py
+++ b/stock_barcodes_gs1/tests/test_stock_barcodes_gs1.py
@@ -3,11 +3,12 @@
 
 from odoo.tests import common
 
+from odoo.addons.mail.tests.common import MailCommon
 from odoo.addons.stock_barcodes.tests.test_stock_barcodes import TestStockBarcodes
 
 
 @common.tagged("post_install", "-at_install")
-class TestStockBarcodesGS1(TestStockBarcodes):
+class TestStockBarcodesGS1(TestStockBarcodes, MailCommon):
     # pylint: disable=W8121
     @classmethod
     def setUpClass(cls):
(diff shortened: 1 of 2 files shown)


chienandalu (approved):

rousseldenis:
@sergio-teruel Could you check tests?

rousseldenis:
@sergio-teruel What's the status of this ? FYI, @alexis-via has done an improvement too for barcode lib: https://github.com/OCA/stock-logistics-barcode/pull/566

sergio-teruel:
It is this module that we no longer use base_gs1_barcode. Now we use the barcodes_gs1_nomenclature core module of Odoo

sergio-teruel:
I have to review the tests

sergio-teruel:
Tests fixed

CarlosRoca13 (approved):

alexis-via:
Did you test the quality of the code of Odoo in the new native module barcodes_gs1_nomenclature ?
I just made a test on a medicine box (Amoxicilline) I have at home with Odoo v16:
barcode : 0103400935251862211824243996541210LT902217241130
stdnum.gs1_128.info() returns {'01': '03400935251862', '21': '1824243996541210LT90', '22': '17241130'}
self.env.company.nomenclature_id.filtered("is_gs1_nomenclature").parse_barcode() returns None

I have the feeling that, because the native GS1 nomenclature doesn't define AI 22, it returns None... (I didn't analyse the code yet).

alexis-via:
An analysis of the code of barcode_gs1_nomenclature confirms this and some further tests with my Amoxicilline box confirms this : it the GS1 barcode contains an AI that is NOT defined in the GS1 nomenclature in Odoo, self.env.company.nomenclature_id.filtered("is_gs1_nomenclature").parse_barcode() returns None


This is caused by this line of code:
https://github.com/odoo/odoo/blob/16.0/addons/barcodes_gs1_nomenclature/models/barcode_nomenclature.py#L123
If you replace "return None" by "return results", it will work.
Another solution is to define AI 22 in the GS1 barcode nomenclature in Odoo => it will read the barcode properly. The problem is that there are so many AI that it's difficult to define them all on the GS1 barcode nomenclature in Odoo. But, if you don't do that, it will fail if you receive a product with an AI that is not defined.

PR #495, opened 2023-02-16
[15.0][IMP] stock_barcodes: some improvements

sergio-teruel:
cc @Tecnativa

- Take into account package fields when scan barcodes.
- Add prepare method to set wizard values from picking to be more inheritable.
- Allow display detailed operations on barcode wizard.
- Display partner on barcode wizard.
- Do not clean values forced in an action done.
- Add package to moves todo group key.
- Display picking notes in wizard barcode.
- Display improvements.
- Show same operation type ready pickings tree after validate picking.
- ✴️ **Allow make inventories** ✴️
- Check serial product tracking for inventory when a quant is created.
- Improve heritability of write on stock move lines.
- Better inheritance of qty focus.

.... And more coming soon...   💥 💥

ping @dalonsod @CarlosRoca13 @chienandalu

PR #560, opened 2023-11-20
[14.0][IMP] stock_barcodes: some improvements

dalonsod:
Backport of #495


dalonsod:
@sergio-teruel could you review?

Regarding to commit history, I've tried my best trying to preserve it when porting commits and later making required adjustments for v14.

Thanks in advance!!

elvise (approved):
Functional 👍

sergio-teruel (approved):

stock_barcodes/wizard/stock_barcodes_read_inventory.py
```diff
@@ -1,93 +1,95 @@
...
-                ),
+    def _compute_inventory_quant_ids(self):
+        for wiz in self:
+            quants = self.env["stock.quant"].search(
+                [
+                    ("user_id", "=", self.env.user.id),
+                    "|",
+                    ("inventory_quantity_set", "=", True),
+                    ("inventory_date", "<=", fields.Date.context_today(self)),
+                ],
```
alexis-via, line 29: These 3 fields user_id, inventory_quantity_set and inventory_date don't exist on stock.quant on odoo v14 !!! These fields exists in v15 but not in v14.

alexis-via:
In fact, it's not just a matter of those 3 fields: in v15, stock.inventory and stock.inventory.line have been removed, so there are major differences between an inventory in v15 and an inventory in v14. In this PR, the code has been backported without change from v15 to v14 for the inventory ; I think we'll have to completely roll-back the inventory part of this PR.

dalonsod:
Yes, we know it... we'll make PR today or tomorrow, if we don't find more problems than it. Thanks!!

PR #563, opened 2023-12-06
[IMP]product_packaging_multi_barcode: Display barcode packaging

cyrilmanuel:
Display barcode of packaging product or product if not present.


product_packaging_multi_barcode/models/product_barcode.py
```diff
@@ -14,6 +14,31 @@ class ProductBarcode(models.Model):
         ondelete="cascade",
     )
 
+    display_product_id = fields.Many2one(
```
rousseldenis, line 17: @cyrilmanuel Thanks for this. I would have added it to another module instead.

cyrilmanuel: @rousseldenis  thanks, do you think it shall be added into a new module ? or you have an idea on which one ?

cyrilmanuel: @rousseldenis gentle ping :  any idea or can we merge it like this ?

rousseldenis: @cyrilmanuel Ok, didn't see the packaging_id definition. We can move forward.

rousseldenis (approved):

rousseldenis:
@cyrilmanuel Could you attract a reviewer here ?

wouitmil (approved):

gurneyalex (approved):
Maybe reformat the commit line so that it fits (https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message)

sebalix:
@cyrilmanuel better to not include the Odoo version number in the commit message, as commits can be ported to other versions this is misleading and doesn't help commit comparison afterwards

PR #566, opened 2023-12-15
[14.0] stock_barcodes_gs1: use python-stdnum instead of base_gs1_barcode

alexis-via:
stock_barcodes_gs1: Use the python lib stdnum.gs1_128 instead of the OCA module base_gs1_barcode. The stdnum lib is already used by Odoo, so it's not an additionnal lib. Add debug logs


rousseldenis:
@alexis-via Could you check tests ?

PR #570, opened 2023-12-29
[14.0][FIX] barcodes_generator_product_multi_barcode: fix inheritance

HekkiMelody:
This commit fixes an issue where the barcode generation was being overwritten not just for products, but for every model that inherited from the `barcode.generate.mixin` (users, locations, packages)

Additionally, it also fixes a SingletonError in case `generate_barcode()` was called with a multi recordset

### DIFF
--- a/barcodes_generator_product_multi_barcode/models/barcode_generate_mixin.py
+++ b/barcodes_generator_product_multi_barcode/models/barcode_generate_mixin.py
@@ -1,28 +1,23 @@
 import barcode
 
-from odoo import api, fields, models
+from odoo import models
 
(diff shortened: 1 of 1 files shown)


ivs-cetmix (approved):
LGTM

dessanhemrayev (commented):
Please, change file name)
`barcode_generate_mixin.py` to `product.py`

HekkiMelody:
> Please, change file name) `barcode_generate_mixin.py` to `product.py`

That was correct, but I didn't get the chance to do this fix :D

PR #581, opened 2024-01-29
[14.0][ADD] sale_barcode: add module

dessanhemrayev:
This module adds the product to the sales order line according to the barcode information provided by the barcode/datamatrix device.


dessanhemrayev:
@francesco-ooops please check)

bealdav:
Hi @dessanhemrayev thanks for your contribution.

Have you seen this module https://github.com/OCA/stock-logistics-barcode/tree/14.0/sale_input_barcode

It seems it's about the same features.
Have you considered it to use it avoid to confusion with modules with the same scope ?

What are the missing features ? Should these features can be added in existing module or maybe your module could depends on it ?

francesco-ooops:
Hi @bealdav , I'm not technical but here's the analysis that was done when developing this module: 

- the main request was to make [sale_input_barcode](https://github.com/OCA/stock-logistics-barcode/tree/14.0/sale_input_barcode) work with any type of barcode assigned to product, eg: 

Barcode 5156000000030 was returning "Could not decode barcode: Unknown Application identifier at position 0" error

Here's Cetmix's team analysis:

> For some reason sale_input_barcode depends on [base_gs1_barcode](https://github.com/OCA/stock-logistics-barcode/tree/14.0/base_gs1_barcode) module for  barcode decoding, which specifies "decode a GS1-128/GS1-Datamatrix string to dictionary of values with Application Identifiers as keys"
> 
> Issue you caught occurs when used the [decode method](https://github.com/OCA/stock-logistics-barcode/blob/9cfa08c699f4d31181d759604ed66031df89f218/base_gs1_barcode/models/gs1_barcode.py#L180) from this module.
> 
> On the other hand [stock_barcodes](https://github.com/OCA/stock-logistics-barcode/tree/14.0/stock_barcodes) module depends on Odoo's standard "barcode" module and can handle any barcodes.
> 
> I believe to solve the issue you mentioned we need to completely change the way the sale_input_barcode module  handles barcodes,
> this is such a deep refactoring, how do you think should we made it as IMP in this sale_input_barcode or create a new one?

So we went for this new module, @bealdav if you could review or provide your input that would be great, thanks!

bealdav:
Thanks @francesco-ooops for this clarification.

I think there is a way to merge features of this PR in sale_input_barcode.
I mean the plan is to remove base_gs1_barcode dependency.

If this last module is installed then the current behavior must be kept. If not, the behavior of the current PR is added.

Choice in this behavior must be defined in `_process_barcode_on_product_line`

I don't know if my explanation is ok for you ?

I like in your module these features 

- update sale line when enter the same barcode
- feedback to the user about barcode scanned

Thanks a lot

PR #587, opened 2024-02-28
[16.0] [MIG] stock_barcodes

FranzPoize:


angelmoya:
Functional testing:

Set Barcode options:
Odoo-Picking-IN-options

Set this barcode option on incoming picking type.

Configura one product tracking as serial number.

Purchase this product to create a incoming picking.

Go to incoming picking and get this error trying to open barcode interface:

Traceback (most recent call last):
  File "/opt/odoo/odoo/http.py", line 1633, in _serve_db
    return service_model.retrying(self._serve_ir_http, self.env)
  File "/opt/odoo/odoo/service/model.py", line 133, in retrying
    result = func()
  File "/opt/odoo/odoo/http.py", line 1660, in _serve_ir_http
    response = self.dispatcher.dispatch(rule.endpoint, args)
  File "/opt/odoo/odoo/http.py", line 1864, in dispatch
    result = self.request.registry['ir.http']._dispatch(endpoint)
  File "/opt/odoo/odoo/addons/base/models/ir_http.py", line 154, in _dispatch
    result = endpoint(**request.params)
  File "/opt/odoo/odoo/http.py", line 697, in route_wrapper
    result = endpoint(self, *args, **params_ok)
  File "/opt/odoo/addons/web/controllers/dataset.py", line 46, in call_button
    action = self._call_kw(model, method, args, kwargs)
  File "/opt/odoo/addons/web/controllers/dataset.py", line 33, in _call_kw
    return call_kw(request.env[model], method, args, kwargs)
  File "/opt/odoo/odoo/api.py", line 468, in call_kw
    result = _call_kw_multi(method, model, args, kwargs)
  File "/opt/odoo/odoo/api.py", line 453, in _call_kw_multi
    result = method(recs, *args, **kwargs)
  File "/mnt/data/odoo-addons-dir/stock_barcodes/models/stock_picking.py", line 35, in action_barcode_scan
    wiz.determine_todo_action()
  File "/mnt/data/odoo-addons-dir/stock_barcodes/wizard/stock_barcodes_read_picking.py", line 188, in determine_todo_action
    self.fill_todo_records()
  File "/mnt/data/odoo-addons-dir/stock_barcodes/wizard/stock_barcodes_read_picking.py", line 181, in fill_todo_records
    self.env["wiz.stock.barcodes.read.todo"].fill_records(self, [move_lines])
  File "/mnt/data/odoo-addons-dir/stock_barcodes/wizard/stock_barcodes_read_todo.py", line 163, in fill_records
    todo_vals[key] = self._prepare_fill_record_values(
  File "/mnt/data/odoo-addons-dir/stock_barcodes/wizard/stock_barcodes_read_todo.py", line 88, in _prepare_fill_record_values
    "product_uom_qty": line.product_uom_qty,
AttributeError: 'stock.move.line' object has no attribute 'product_uom_qty'

stock_barcodes/__manifest__.py
```diff
@@ -3,12 +3,12 @@
...
     "name": "Stock Barcodes",
     "summary": "It provides read barcode on stock operations.",
-    "version": "15.0.2.0.0",
+    "version": "16.0.1.0.0",
     "author": "Tecnativa, " "Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/stock-logistics-barcode",
     "license": "AGPL-3",
     "category": "Extra Tools",
-    "depends": ["barcodes", "stock", "web_widget_numeric_step"],
+    "depends": ["barcodes", "stock", "web_widget_numeric_step", "web"],
```
chienandalu: `web_widget_numeric_step` already depends on web

stock_barcodes/wizard/stock_barcodes_read_picking_views.xml
```diff
@@ -125,121 +126,116 @@
...
+                                                <div class="col-6">
+                                                    <strong>
+                                                        <span>
+                                                            <field name="product_id" />
+                                                        </span>
+                                                    </strong>
+                                                </div>
+                                                <div class="col-2">
+                                                    <span
+                                                        class="pull-right text-right text-muted"
```
chienandalu: Review your bootstrap classes as they should be migrated to BS5: https://getbootstrap.com/docs/5.0/migration/

You've got some tooling in `openupgradelib` to analyze your view code: https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade_160.py

stock_barcodes/static/src/barcodes_models_utils.esm.js
```diff
@@ -0,0 +1,19 @@
+/** @odoo-module */
+/* Copyright 2022 Tecnativa - Alexandre D. Díaz
+ * License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). */
+
+const barcodeModels = [
```
chienandalu:
```suggestion
export const barcodeModels = [
```

stock_barcodes/static/src/barcodes_models_utils.esm.js
```diff
@@ -0,0 +1,19 @@
...
+    "wiz.candidate.picking",
+    "wiz.stock.barcodes.new.lot",
+    "wiz.stock.barcodes.read",
+    "wiz.stock.barcodes.read.inventory",
+    "wiz.stock.barcodes.read.picking",
+    "wiz.stock.barcodes.read.todo",
+];
+
+export function isAllowedBarcodeModel(modelName) {
+    return barcodeModels.indexOf(modelName) !== -1;
```
chienandalu: A little bit more expressive:
```suggestion
    return barcodeModels.includes(modelName);

File: stock_barcodes/static/src/barcodes_models_utils.esm.js
Comment (chienandalu) [inline stock_barcodes/static/src/barcodes_models_utils.esm.js:1]: Organize these source files at leat in functionality folders so it's more clear where to look what

File: stock_barcodes/static/src/barcodes_models_utils.esm.js [hunk 5]
```diff
@@ -0,0 +1,19 @@
...
+    "stock.picking.type",
+    "wiz.candidate.picking",
+    "wiz.stock.barcodes.new.lot",
+    "wiz.stock.barcodes.read",
+    "wiz.stock.barcodes.read.inventory",
+    "wiz.stock.barcodes.read.picking",
+    "wiz.stock.barcodes.read.todo",
+];
+
+export function isAllowedBarcodeModel(modelName) {
```
chienandalu: This also had a docstring

stock_barcodes/static/src/boolean_toggle.esm.js
```diff
@@ -0,0 +1,17 @@
+/** @odoo-module */
+/* Copyright 2018-2019 Sergio Teruel <sergio.teruel@tecnativa.com>.
+ * License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). */
+
+import {BooleanToggleField} from "@web/views/fields/boolean_toggle/boolean_toggle_field";
+import {registry} from "@web/core/registry";
+
+class BarcodeBooleanToggleField extends BooleanToggleField {
+    onChange(newValue) {
```
chienandalu: Missing documentation from the original here as well where it was stated why this was needed...

stock_barcodes/static/src/form_view.esm.js
```diff
@@ -0,0 +1,15 @@
+/** @odoo-module */
+/* Copyright 2021 Tecnativa - Alexandre D. Díaz
+ * License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). */
+
+import {FormController} from "@web/views/form/form_controller";
+import {patch} from "@web/core/utils/patch";
+
+patch(FormController.prototype, "Allow display.controlPanel overriding", {
+    setup() {
```
chienandalu: Docs missing here as well

stock_barcodes/static/src/kanban_renderer.esm.js
```diff
@@ -0,0 +1,163 @@
...
+            return;
+        }
+        const groups = isGrouped
+            ? [...area.querySelectorAll(".o_kanban_group")]
+            : [area];
+        let cards = [...groups]
+            .map((group) => [...group.querySelectorAll(".o_kanban_record")])
+            .filter((group) => group.length);
+
+        if (isAllowedBarcodeModel(this.props.list.resModel)) {
```
chienandalu: Why do yo need to copy the whole method if you're returning in the middle of it? Maybe you have to repeat some of it if it's encapsulated, but in the end you should try to return `super`. Just evaluate this first, and repeat as less as possible

stock_barcodes/static/src/kanban_renderer.esm.js
```diff
@@ -0,0 +1,163 @@
...
+import {patch} from "@web/core/utils/patch";
+import {useBus} from "@web/core/utils/hooks";
+import {useHotkey} from "@web/core/hotkeys/hotkey_hook";
+import {useRef} from "@odoo/owl";
+
+patch(KanbanRenderer.prototype, "add hotkey", {
+    setup() {
+        const rootRef = useRef("root");
+        useHotkey(
+            "Enter",
```
chienandalu: What's the purpose of this?

chienandalu (changes requested):
Hi @FranzPoize Thanks for your work here.

I didn't go deep into the python logic, although it looks correct at first sight.

As a general comment, I'm finding hard to review the changes in the javascript logic:

- There are no diffs
- There's missing documentation from the former code and for the new it isn't explained.
- There are overrides that copy the original method (a thus the inheritance is broken for those)

Some more specific comments:

FranzPoize:
Thank you @chienandalu for your review

> * There are no diffs

Yes unfortunately the differences in behavior between the legacy front and owl are truly unreconcialable...
First of all the file name are changed to use the new odoo-module.
Also for example most of the logic for hotkey management was in kanban renderer. In the new frontend there can be many kanban renderer instances which would have registered their own global management of hotkey this needed to be completely refactored.
> * There's missing documentation from the former code and for the new it isn't explained.

I hope it's fixed in the latest push
> * There are overrides that copy the original method (a thus the inheritance is broken for those)

Yes that is not inevitable but is the least ugly way to do what needs to be done I'll go in more detail in the review comment

chienandalu (commented):
Thanks @FranzPoize take also in consideration that there are lots of recent changes in the 15.0 that should be included as a base to the migration as well...

stock_barcodes/static/src/kanban_renderer.esm.js
FranzPoize: The code is not returning in the `if` statement it's recomputing the `cards` variable.
I could call super and return its value but I would still have duplicate that code because nothing is returned from the parent function. I still need to compute the candidate `cards` still filter them and focus the correct one.

If you have a way to do it better I would need more information because I'm not sure how to go about it.

stock_barcodes/static/src/barcodes_models_utils.esm.js
FranzPoize: Yes thank you !

stock_barcodes/static/src/barcodes_models_utils.esm.js
FranzPoize: Thank you ! Changed

stock_barcodes/static/src/barcodes_models_utils.esm.js
FranzPoize: Tried to. Tell me if you thinks that's better

stock_barcodes/static/src/barcodes_models_utils.esm.js
FranzPoize: Added

stock_barcodes/static/src/boolean_toggle.esm.js
FranzPoize: Added !

stock_barcodes/static/src/form_view.esm.js
FranzPoize: Added !

stock_barcodes/static/src/kanban_renderer.esm.js
FranzPoize: Not sure what you're referring to ? the rootRef or the useHotkey ?

chienandalu (changes requested):
Also testing a manual entry, I found this error

```
File "/opt/odoo/auto/addons/stock_barcodes/wizard/stock_barcodes_read.py", line 744, in action_confirm
    res = self.action_done()
  File "/opt/odoo/auto/addons/stock_barcodes/wizard/stock_barcodes_read_picking.py", line 269, in action_done
    ).action_done()
  File "/opt/odoo/auto/addons/stock_barcodes/wizard/stock_barcodes_read.py", line 547, in action_done
    if not self.check_done_conditions():
  File "/opt/odoo/auto/addons/stock_barcodes/wizard/stock_barcodes_read_picking.py", line 670, in check_done_conditions
    and float_compare(
  File "/opt/odoo/custom/src/odoo/odoo/tools/float_utils.py", line 155, in float_compare
    rounding_factor = _float_check_precision(precision_digits=precision_digits,
  File "/opt/odoo/custom/src/odoo/odoo/tools/float_utils.py", line 29, in _float_check_precision
    assert precision_rounding is None or precision_rounding > 0,\
AssertionError: precision_rounding must be positive, got 0.0

The above server error caused the following client error:
RPC_ERROR: Odoo Server Error
    at makeErrorFromResponse (http://localhost:16069/web/assets/3930-e51fdb0/web.assets_backend.min.js:993:163)
    at XMLHttpRequest.<anonymous> (http://localhost:16069/web/assets/3930-e51fdb0/web.assets_backend.min.js:1001:13)

**Comment by FranzPoize** [comment]
Yes this PR includes all the changes up to 28th of february not sure I'll have the bandwidth to add the more recent changes if somebody can help with that I'd be glad.

I'll fix the codecov issue

**Comment by chienandalu** [comment]
> Yes this PR includes all the changes up to 28th of february not sure I'll have the bandwidth to add the more recent changes if somebody can help with that I'd be glad.
> 
> I'll fix the codecov issue

We can do this: finish your ongoing fixes and then we can supersed the migration from the 15.0 branch and putting on top your migration commit fixing conflicts and finishing whatever needs to be fixed from there. What do you think?

File: stock_barcodes/wizard/stock_barcodes_read.py [hunk 10]
```diff
@@ -0,0 +1,745 @@
...
+
+    def _prepare_lot_vals(self):
+        return {
+            "name": self.lot_name,
+            "product_id": self.product_id.id,
+            "company_id": self.env.company.id,
+        }
+
+    def _create_new_lot(self):
+        StockProductionLot = self.env["stock.production.lot"]
```
carolinafernandez-tecnativa, line 822: Hi, you need to change this model to stock.lot

        StockProductionLot = self.env["stock.lot"]

chienandalu:
Hi, @FranzPoize what do you think about what I proposed?:

> We can do this: finish your ongoing fixes and then we can supersed the migration from the 15.0 branch and putting on top your migration commit fixing conflicts and finishing whatever needs to be fixed from there. What do you think?

FranzPoize:
@chienandalu 
I'm having a real trouble figuring how the code works.

When you use the barcode button on a picking type. It calls `action_barcode_scan` on the `stock_picking_type` models there is 3 methods that are called:
* `create`
* `determine_todo_action`
* `fill_pending_moves`

1. First called is `create` this call `set_candidate_picking` if there is a `picking_id` but there is no `picking_id` in the parameters (even when we click on the new button on a picking type). It then calls `action_show_step` which sets the correct step because most `filled_default` are still correctly set (by `_prepare_barcode_wiz_vals`)

2. Second is `determine_todo_action`, if the barcode option is guided, this calls `fill_todo_records` if there is no `todo_line_ids` (and there isn't) which calls `_get_stock_move_lines`. Since there is no `picking_id`, `_get_stock_move_lines` has to return an empty list. In this method all the default are filled but since there is no `todo_line_ids` all those value are set to empty sets. This then calls `action_show_step` again which remove the properly set step because the default value are now all empty.

3. Third a call to `fill_pending_moves` this call `fill_todo_records` again which still does nothing.

So during `action_barcode_scan`:
* `create` sets the proper default values
* `action_show_step` is called twice once it sets a step (during) `create` and the second time it undoes what it did because `determine_todo_action` unset all the default values
* `fill_todo_records` is called twice and twice it does nothing

This means no step at the end in some case (which are real convoluted but not so much).

* Set the barcode option as guided
* Leave no step with a step number of 0 (if a step with a 0 selected is `action_show_step` it will trigger the line`self.step = options_required[:1].step`because `not 0` is `True`).
* And click on the barcode of the `picking_type`

Or simpler use the demo Picking internals barcode option on internal transfer and click on the barcode on the internal transfer picking type.

FranzPoize:
Found the commit that adds a new call to `action_show_step` in `determine_todo_action` 50b0c67107c2c618ceff60752222305df10659b5 

This seems to break the option/step finding since if there is no `todo_line_ids` ,which there isn't in all case where you click on a barcode button on a picking type.

My question is it suppose to work without a picking_id in the `stock_barcodes_read_picking` ? There is a picking_ids field but it is not used anywhere in the code.

@chienandalu @sergio-teruel

pedrobaeza:
To be taken into account: #605

FranzPoize:
The frontend has been completely redone this should be resolved in 16.0

Le jeu. 2 mai 2024 à 12:56, Pedro M. Baeza ***@***.***> a
écrit :

sergio-teruel:
To be taken into account: https://github.com/OCA/stock-logistics-barcode/pull/604

sergio-teruel:
> @chienandalu I'm having a real trouble figuring how the code works.
> 
> When you use the barcode button on a picking type. It calls `action_barcode_scan` on the `stock_picking_type` models there is 3 methods that are called:
> 
> * `create`
> * `determine_todo_action`
> * `fill_pending_moves`
> 
> 1. First called is `create` this call `set_candidate_picking` if there is a `picking_id` but there is no `picking_id` in the parameters (even when we click on the new button on a picking type). It then calls `action_show_step` which sets the correct step because most `filled_default` are still correctly set (by `_prepare_barcode_wiz_vals`)
> 2. Second is `determine_todo_action`, if the barcode option is guided, this calls `fill_todo_records` if there is no `todo_line_ids` (and there isn't) which calls `_get_stock_move_lines`. Since there is no `picking_id`, `_get_stock_move_lines` has to return an empty list. In this method all the default are filled but since there is no `todo_line_ids` all those value are set to empty sets. This then calls `action_show_step` again which remove the properly set step because the default value are now all empty.
> 3. Third a call to `fill_pending_moves` this call `fill_todo_records` again which still does nothing.
> 
> So during `action_barcode_scan`:
> 
> * `create` sets the proper default values
> * `action_show_step` is called twice once it sets a step (during) `create` and the second time it undoes what it did because `determine_todo_action` unset all the default values
> * `fill_todo_records` is called twice and twice it does nothing
> 
> This means no step at the end in some case (which are real convoluted but not so much).
> 
> * Set the barcode option as guided
> * Leave no step with a step number of 0 (if a step with a 0 selected is `action_show_step` it will trigger the line`self.step = options_required[:1].step`because `not 0` is `True`).
> * And click on the barcode of the `picking_type`
> 
> Or simpler use the demo Picking internals barcode option on internal transfer and click on the barcode on the internal transfer picking type.

The next week in Spanish OCA days we will work about it...

pedrobaeza:
Superseded by #610, that was the result of the Spanish OCA days,

PR #588, opened 2024-03-04
[15.0][IMP] stock_picking_product_barcode_report: Allow force qty using context

CarlosRoca13:
cc @Tecnativa TT48162


stock_picking_product_barcode_report/wizard/stock_barcode_selection_printing.py
```diff
@@ -121,9 +121,10 @@ def _get_label_qty(self, move_line):
 
     @api.model
     def _prepare_data_from_move_line(self, move_line):
+        qty = self.env.context.get("force_quantity_line") or move_line.qty_done
```
carlosdauden: Do you want use move_line.qty_done if force_quantity_line is zero?
```suggestion
        qty = self.env.context.get("force_quantity_line", move_line.qty_done)
```

sergio-teruel (approved):

carlosdauden (approved):

PR #600, opened 2024-04-05
[15.0][IMP] stock_barcodes: Skip backorder creation for pickings with all moves read or ignore pending quantities

sergio-teruel:
cc @tecnativa 

### DIFF
--- a/stock_barcodes/models/stock_picking.py
+++ b/stock_barcodes/models/stock_picking.py
@@ -1,6 +1,7 @@
 # Copyright 2019 Sergio Teruel <sergio.teruel@tecnativa.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 from odoo import models
+from odoo.tools.float_utils import float_compare
 
 
 class StockPicking(models.Model):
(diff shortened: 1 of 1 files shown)


carlosdauden (approved):

chienandalu (approved):

stock_barcodes/models/stock_picking.py
```diff
@@ -46,7 +47,38 @@ def button_validate(self):
...
+                    < 0
+                ):
+                    # In normal conditions backorder will be created
+                    create_backorder = True
+                    if not move.move_line_ids or any(
+                        sml.state in ["pending"] for sml in move.move_line_ids
+                    ):
+                        # If any move are not processed we can not skip backorder
+                        skip_backorder = False
+                        break
```
pedrobaeza, line 72: Any line will impact in the whole picking?

pedrobaeza (commented):
I miss regression tests for all these specific cases.

stock_barcodes/models/stock_picking.py
pedrobaeza: @sergio-teruel can you answer this? The problem I see is that this aborts the evaluation of the rest of the picking. Is that correct?

pedrobaeza: ping @sergio-teruel

sergio-teruel:
Yes, I know!!

pedrobaeza:
OK, it seems weird...

/ocabot merge patch

Please include it in the v16 migration

PR #601, opened 2024-04-11
[15.0][IMP] stock_barcodes: Add option group to allow to select if you want order by location or location_dest_id

sergio-teruel:
cc @Tecnativa TT48125

With this PR user can select the field to order the pending moves, location or location_dest. Is usefull for internal operations

ping @chienandalu @CarlosRoca13

### DIFF
--- a/stock_barcodes/models/stock_barcodes_option.py
+++ b/stock_barcodes/models/stock_barcodes_option.py
@@ -97,6 +97,12 @@ class StockBarcodesOptionGroup(models.Model):
     use_location_dest_putaway = fields.Boolean(
         string="Use location dest. putaway",
     )
+    location_field_to_sort = fields.Selection(
+        selection=[
+            ("location_id", "Origin Location"),
+            ("location_dest_id", "Destination Location"),
+        ]
+    )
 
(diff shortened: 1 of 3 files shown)


chienandalu (approved):
:+1:

CarlosRoca13 (approved):

stock_barcodes/models/stock_barcodes_option.py
```diff
@@ -97,6 +97,12 @@ class StockBarcodesOptionGroup(models.Model):
     use_location_dest_putaway = fields.Boolean(
         string="Use location dest. putaway",
     )
+    location_field_to_order = fields.Selection(
```
pedrobaeza: Although it can be considered correct, it's better to use `sort` to differentiate this from order as document (sales order, purchase order...). Is it too much to change it?

```suggestion
    location_field_to_sort = fields.Selection(
```

sergio-teruel:
Changes done!!! Thanks

pedrobaeza (approved):
Thanks for changing it!

/ocabot merge minor

PR #605, opened 2024-05-02
[15.0][FIX] stock_barcodes: Error when using arrows on Inventory Overview

CarlosRoca13:
To reproduce the error use arrows of the keyboard on the Inventory Overview. An error will be thrown because the function _onDocumentKeyDown is undefined.

By doing this change, when the function is undefined it will work like normally it does.

cc @Tecnativa TT49029


sergio-teruel (approved):

sergio-teruel:
ping @chienandalu

pedrobaeza (approved):
/ocabot merge patch

PR #607, opened 2024-05-06
[ADD] barcode_validation: Base barcode validation

EmilioPascual:
Odoo allows set illegal barcode for some barcode type but an ugly error can happen when you print report with this illegal barcode.

This a base module.

@yajo @shide @rafaelbn please review

MT-5927 @moduon


barcode_validation/models/barcode_validation_mixin.py
```diff
@@ -0,0 +1,36 @@
...
+            validation_barcode_type,
+        ) in validation_barcode_settings.items():
+            if validation_barcode_type:
+                try:
+                    self.env["ir.actions.report"].barcode(barcode_type, barcode)
+                except ValueError as e:
+                    raise ValidationError(
+                        _(
+                            "The barcode {barcode} is not a valid {barcode_type} barcode."
+                        ).format(barcode=barcode, barcode_type=barcode_type)
```
yajo, line 35: issue: as a rule of thumb, don't use f-strings in translations.

Probably it's harmless in this case, but in other cases, like passing in a recordset as an argument, it could open a vulnerability door where you could hack odoo just through mis-using weblate.

Using %-format strings is always safe, and besides it's supported by `_()` directly:

```suggestion
                        _(
                            "The barcode %(barcode)s is not a valid %(barcode_type)s barcode."
                            barcode=barcode,
                            barcode_type=barcode_type,
                        )
```

barcode_validation/readme/CONTEXT.md
```diff
@@ -0,0 +1,8 @@
+This module was developed because Odoo allows set illegal barcode for some barcode type but an error can happen when you print report with this illegal barcode.
+
+It will be useful for you if you use barcodes in products, lots, packagings, ...
+
+If you need this module for those reasons, these might interest you too:
+
+- product_varcode_validation
```
yajo, line 7:
```suggestion
- product_barcode_validation
```

barcode_validation/readme/DESCRIPTION.md
```diff
@@ -0,0 +1 @@
+This module extends the functionality of Odoo and to allow you validate barcode, so when print report with barcode, if it's illegal barcode for some barcode type, you will not have a ugly error.
```
yajo, line 1:
```suggestion
This module extends the functionality of Odoo and to allow you validate barcode, so when print report with barcode, if it's illegal barcode for some barcode type, you will not have an ugly error.
```

barcode_validation/readme/USAGE.md
```diff
@@ -0,0 +1,19 @@
+This module is an abstract module. You can configure barcode validation, but to enable this feature, you need to install an extra module for a given model. This repository provide 'product_barcode_validation' and 'stock_barcode_validation' to validate barcodes for products, stock lots, product packagings, ...
```
yajo, line 1: thought: is there really any use of barcode validation in a lower stack than the `product` module? Couldn't you just merge `barcode_validation` into `product_barcode_validation` and consider the latter one the base for others?

yajo (changes requested):

barcode_validation/models/barcode_validation_mixin.py
```diff
@@ -0,0 +1,36 @@
...
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl-3.0)
+
+from odoo import _, models
+from odoo.exceptions import ValidationError
+
+
+class BarcodeValidation(models.AbstractModel):
+    _name = "barcode.validation.mixin"
+    _description = "Barcode Validation"
+
```
yajo, line 11: suggestion: I don't know if the answer to https://github.com/OCA/stock-logistics-barcode/pull/607#discussion_r1591886581 will imply that you don't need a mixin anymore. But in case you still need it, I think the mixin is incomplete. I'd add this:

```suggestion
    _barcode_field = "barcode"  # Override if needed when applying mixin
    
    @api.constrains(lambda self: (self._barcode_field,))
    def _check_barcode(self):
        for record in self:
            record._validate_barcode(record[self._barcode_field])

    @api.model
```

barcode_validation/models/barcode_validation_mixin.py
```diff
@@ -0,0 +1,36 @@
...
+        return {
+            "UPCA": self.env.company.barcode_validation_upca,
+            "EAN8": self.env.company.barcode_validation_ean8,
+            "EAN13": self.env.company.barcode_validation_ean13,
+            "Code128": self.env.company.barcode_validation_code128,
+            "DataMatrix": self.env.company.barcode_validation_datamatrix,
+            "QR": self.env.company.barcode_validation_qrcode,
+        }
+
+    def _validate_barcode(self, barcode):
```
yajo, line 22:
```suggestion
    @api.model
    def _validate_barcode(self, barcode):
```

barcode_validation/tests/models.py
```diff
@@ -0,0 +1,17 @@
...
+    _name = "test.model"
+    _inherit = ["barcode.validation.mixin"]
+
+    name = fields.Char(required=True)
+    barcode = fields.Char(copy=False)
+
+    @api.constrains("barcode")
+    def check_barcode_validation(self):
+        for record in self.filtered("barcode"):
+            record._validate_barcode(record.barcode)
```
yajo, line 17: suggestion: following the other comments in this review, you won't need all this code.

```suggestion
```

yajo (changes requested):
Another small refactor too.

barcode_validation/models/barcode_validation_mixin.py
```diff
@@ -0,0 +1,36 @@
...
+from odoo import _, models
+from odoo.exceptions import ValidationError
+
+
+class BarcodeValidation(models.AbstractModel):
+    _name = "barcode.validation.mixin"
+    _description = "Barcode Validation"
+
+    def _get_validation_barcode_settings(self):
+        return {
```
rousseldenis, line 13: IMHO, I think having these individual behaviours is not really useful. As you want to display a better error message than the Odoo one (for all barcodes).

IMHO, if you want to introduce other barcode types validation, why not overriding the barcode() function to validate it (and return a ValueError too) ?

barcode_validation/models/barcode_validation_mixin.py
```diff
@@ -0,0 +1,36 @@
...
+
+from odoo import _, models
+from odoo.exceptions import ValidationError
+
+
+class BarcodeValidation(models.AbstractModel):
+    _name = "barcode.validation.mixin"
+    _description = "Barcode Validation"
+
+    def _get_validation_barcode_settings(self):
```
Shide, line 12:
```suggestion
    @api.model
    def _get_validation_barcode_settings(self):
```

barcode_validation/models/barcode_validation_mixin.py
Shide, line 22:
```suggestion
    @api.model
    def _validate_barcode(self, barcode):
```

Shide (changes requested):
I have doubts in this module. 
Maybe in some models you want to store EAN13 and other kind of barcodes on another.

Both should pass, but if you set up EAN13 validation and UPCA in packages, you couldn't check both of these.

I think this module should use the Nomenclature to check if the barcode fits with any rule.
`barcode.nomenclature` model has `parse_barcode` method that gives you information about what is this Barcode. 
So I think getting the nomenclature (maybe in a manual form) and try to check what is going to discover Nomenclature when is scanned could be a great option too.
Another model with results could be great too to check what Barcodes has problems or its model doesn't fit the Nomenclature related object.

barcode_validation/models/barcode_validation_mixin.py
yajo: I think the point is also to display the error early: while entering the barcode. Otherwise, when you go to the warehouse and try to print, it becomes a bottleneck for the delivery process, because printing a big wave report with many barcodes won't tell which one fails.

PR #611, opened 2024-05-10
[IMP] sale_input_barcode: scan directly on sale.order

renda-dev:
- Add the possibility to directly scan barcodes on the `sale.order` form
- If you scan the same product twice, it will adjust the product quantity instead of creating a new line

### DIFF
--- a/sale_input_barcode/README.rst
+++ b/sale_input_barcode/README.rst
@@ -47,8 +47,16 @@ Usage
 #. Press the button in the header with the QR Code icon
 #. Use a barcode scanner to scan a barcode.
 
+Alternatively,
+#. Navigate to a Sales Order.
+#. Activate "Edit Mode"
+#. Use a barcode scanner to scan a barcode.
+
 If a product is found using the barcode,
 a new line with that product will be added to the Sales Order.
+
+If a product is already inside the sale order lines,
+the quantity will be updated.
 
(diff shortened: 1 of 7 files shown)


francesco-ooops (approved):
Functional ok!

bealdav:
I have to dive in.

francesco-ooops:
Hi @bealdav , could you take a look at this? It makes the module much more usable as per features described in the first comments

HekkiMelody (approved):
Code review, LGTM

francesco-ooops:
@OCA/logistics-maintainers could you please take a look? thank you!

bealdav:
> If you scan the same product twice, it will adjust the product quantity instead of creating a new line

It seems a good idea to have an option to adjust quantity when the same product is added but I think it should be an option because you may want to create a new line (the initial object of the module when I created it)

The way you did it can the default behavior but I suppose that we can propose the previous to do it with a context in the wizard ?
You may also implement an option in the company or in the ir.config.parameter.

Could you implement both behavior ?

Thanks a lot

cc @matthieusaison

SylweKra:
Superseeded by https://github.com/OCA/stock-logistics-barcode/pull/669

gurneyalex:
superseeded PR -> closing

PR #593, opened 2024-03-20
[17.0][MIG] barcodes_generator_abstract: Migration to 17.0

niboo-jva:

PR #615, opened 2024-05-25
[17.0][MIG] barcodes_generator_abstract: Migration to 17.0

isaako34:
Migrate the module barcodes_generator_abstract to V17.

This PR supersedes the [PR#593](https://github.com/OCA/stock-logistics-barcode/pull/593)


isaako34:
@luisg123v Could you please review.

requirements.txt
```diff
@@ -1,2 +1,2 @@
 # generated from manifests external_dependencies
-barcode
+python-barcode
```
legalsylvain, line 2: @sbidoul in the new v17 package design, what is the good way to map barcode to python-barcode ?

sbidoul: Since v13 (or even 12, I don't remember), if you want https://pypi.org/project/python-barcode/, you need to declare `python-barcode` in the manifest `external-dependencies`. There is no mapping to do. The options that could be set in setup.py can still be set in pyproject.toml but I don't think that's ever needed in practice since v13.

legalsylvain: Here is the context : 
1) if we write ```python-barcode``` we have the following error, when trying to install the module

```
  File "env-17.0/src/odoo/odoo/addons/base/models/ir_module.py", line 356, in check_external_dependencies
    raise UserError(msg)
odoo.exceptions.UserError: Unable to install module "barcodes_generator_abstract" because an external dependency is not met: Python library not installed: python-barcode

```
- https://github.com/odoo/odoo/blob/17.0/odoo/addons/base/models/ir_module.py#L351
- https://github.com/odoo/odoo/blob/17.0/odoo/modules/module.py#L465

2) If we write ```barcode```, we install a bad library so, we don't have required function. See : https://github.com/OCA/stock-logistics-barcode/pull/615/commits/1e225b7018213499caf2de2e34b6d9c9699ca599


> The options that could be set in setup.py can still be set in pyproject.toml but I don't think that's ever needed in practice.

The options was used for this module in [V12](https://github.com/OCA/stock-logistics-barcode/blob/12.0/setup/barcodes_generator_abstract/setup.py#L6), [V16](https://github.com/OCA/stock-logistics-barcode/blob/16.0/setup/barcodes_generator_abstract/setup.py#L6), etc...

See previous discussion : https://github.com/OCA/stock-logistics-barcode/pull/258

Do you know the right way to do it via pyproject.toml ?

sbidoul: @legalsylvain I don't understand. This PR has python-barcode in external-dependencies, and the tests are green. All looks good, no?

legalsylvain: oh. Indeed, I installed python-barcode locally and now all is good. In previous Odoo release, the replacement was mandatory. I don't understand why now all is OK. but it's great ! 

sorry for the noice. thanks for your answer. Have a nice week-end.

sbidoul: Great. Note it was necessary until v12, but since then we can put the pypi name in external-dependencies.

luisg123v: Yes. Actually, using the PyPI package is the recommended way ever since (see https://github.com/odoo/odoo/pull/25549), not the other way around. BTW, I just noticed that change was implemented by @sbidoul himself :)

barcodes_generator_abstract/views/view_barcode_rule.xml
```diff
@@ -14,40 +14,28 @@ License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
             <xpath expr="//form/group" position="after">
                 <group
                     string="Barcode Generation"
-                    attrs="{'invisible': ['|', ('encoding', '=', 'any'), ('padding', '=', 0)]}"
+                    invisible="encoding == 'any' or padding == 0"
```
luisg123v: `or not padding`

test-requirements.txt
```diff
@@ -0,0 +1 @@
+odoo_test_helper
```
luisg123v, line 1: Remove this commit and move this change to the migration commit. There should be only one commit addressing migration changes.

barcodes_generator_abstract/__manifest__.py
```diff
@@ -21,5 +21,5 @@
         "views/menu.xml",
     ],
     "demo": ["demo/res_users.xml"],
-    "external_dependencies": {"python": ["barcode"]},
+    "external_dependencies": {"python": ["python-barcode"]},
```
luisg123v, line 24: Check if this change works in v16. If so (most likely case), apply this in that version. If no, integrate into the migration commit.

luisg123v (changes requested):

isaako34:
@luisg123v  ready for review

luisg123v (approved):
LGTM :+1: 

@legalsylvain what do you think?

legalsylvain (approved):
Diff review. LGTM. No test.

Thanks !

luisg123v:
@moylop260 could you review/merge, please?

PR #618, opened 2024-05-28
[15.0][FIX] stock_barcodes: The field state from stock move line compared is not correct

sergio-teruel:
cc @Tecnativa TT49399


carlosdauden (approved):

pedrobaeza (approved):
/ocabot merge patch

I suppose this is to be put on the 16.0 PR

PR #619, opened 2024-05-29
[16.0][IMP] barcodes_generator_abstract: Avoid double iteration on base generation

dalonsofl:
### Module

barcodes_generator_abstract

### Describe the bug

The base is changed when it is generated by having the ‘Automatic generation’ marked in the barcode nomenclature.

### To Reproduce

1. Configure the Automatic generation in the barcode nomenclature
2. Create / Edit a product barcode rule
3. Press on "Generate base"

(image from runboat OCA)

image

image

image

image

**Affected versions:**

16.0

**Additional context**

The problem is that when going through this function inside the [barcodes_generator_product](https://github.com/OCA/stock-logistics-barcode/tree/16.0/barcodes_generator_product) it goes back through the generate_base() and a recalculation of the base incorrectly, as it already has one before. With this change it would be solved

### DIFF
--- a/barcodes_generator_abstract/tests/test_barcodes_generator_abstract.py
+++ b/barcodes_generator_abstract/tests/test_barcodes_generator_abstract.py
@@ -82,6 +82,8 @@ class TestBarcodesGeneratorAbstract(TransactionCase, FakeModelLoader):
         self.user_fake.generate_barcode()
         self.assertEqual(self.user_fake.barcode, "2000001000007")
 
+        self.user_fake.barcode_base = False
+
         self.user_fake.generate_base()
         self.assertEqual(self.user_fake.barcode_base, 2)
         self.user_fake.generate_barcode()
--- a/barcodes_generator_abstract/models/barcode_generate_mixin.py
+++ b/barcodes_generator_abstract/models/barcode_generate_mixin.py
@@ -67,7 +67,7 @@ class BarcodeGenerateMixin(models.AbstractModel):
                         " 'Generate Type' set to 'Base managed by Sequence'"
                     )
                 )
-            else:
+            elif not item.barcode_base:
                 item.barcode_base = item.barcode_rule_id.sequence_id.next_by_id()
 
     def generate_barcode(self):


barcodes_generator_abstract/tests/test_barcodes_generator_abstract.py
```diff
@@ -82,11 +82,6 @@ def test_generate_sequence_sequence(self):
         self.user_fake.generate_barcode()
         self.assertEqual(self.user_fake.barcode, "2000001000007")
 
-        self.user_fake.generate_base()
-        self.assertEqual(self.user_fake.barcode_base, 2)
-        self.user_fake.generate_barcode()
-        self.assertEqual(self.user_fake.barcode, "2000002000006")
-
```
legalsylvain: Why do you remove this test. It looks valid ?

legalsylvain (changes requested):

barcodes_generator_abstract/tests/test_barcodes_generator_abstract.py
dalonsofl: Hi! Because it will be the same base and barcode. With this change if the product already has a base, it will not generate a new one

aliciagaarzo (approved):
LGTM!

barcodes_generator_abstract/tests/test_barcodes_generator_abstract.py
legalsylvain: If the code is called two times, It should be fixed. But this change is not desired.

dalonsofl: I don't undestand really well, it make sense that it doesn't change because you can't change the base two times in UI

legalsylvain: Maybe I did'nt understood the point. I'll investigate a little. 
Thanks for your time.

amarcosg (approved):
Code review LGTM!

barcodes_generator_abstract/tests/test_barcodes_generator_abstract.py
pedrobaeza: Any news, @legalsylvain ?

legalsylvain: I just investigated a little. the problem exists.

I permit me to fix it with https://github.com/OCA/stock-logistics-barcode/pull/651.

@dalonsofl : could you review it ? 

thanks !

dalonsofl: Hi @legalsylvain !

I checked the PR and it looks great!

I change my test like yours and it seems to work great too.

I get the same value with your PR as mine, so both of them are correct.

What do you think with the new changes?

Thanks!!

legalsylvain: Hi @dalonsofl. Thanks for reviewing the other PR. 
Well, I prefer to have explicit text that explain what does the generate_base() function, and to avoid to double check if barcode_base is defined. 
I so merged the other PR. 

thanks for raising the issue.

dalonsofl: I agree, the result is the same, so if you think it's better to have comments on generate_base() , I'm fine with that.

legalsylvain: I merged the other PR, you can close this one.

thanks for the investigation.

PR #625, opened 2024-06-17
[15.0][IMP] stock_barcodes: Reset lot fields if no lot found.Better play sounds

sergio-teruel:
cc @Tecnativa 

ping @carlosdauden @chienandalu

PR #627, opened 2024-07-03
[15.0][IMP] stock_barcodes: Add pending move with difference between demand and reserved

carlosdauden:
@Tecnativa TT49738


carlosdauden:
Added to #625

PR #629, opened 2024-07-09
[16.0][ADD] product_packaging_barcode_constraint_per_company

kevinkhao:


product_packaging_barcode_constraint_per_company/models/product_product.py
```diff
@@ -0,0 +1,18 @@
...
+from odoo import api, models
+
+
+class ProductProduct(models.Model):
+
+    _inherit = "product.product"
+
+    @api.model
+    def _search(self, domain, *args, **kwargs):
+        company_ids = self._context.get("search_only_in_pkg_company")
```
rousseldenis:
```suggestion
        company_ids = self.env.context.get("search_only_in_pkg_company")
```

product_packaging_barcode_constraint_per_company/tests/test_constraints.py
```diff
@@ -0,0 +1,70 @@
...
+        return self.env["product.packaging"].create(
+            {
+                "name": "test pkg",
+                "product_id": product.id,
+                "barcode": BARCODE_CONSTANT,
+                "company_id": company_id,
+            }
+        )
+
+    def setUp(self):
```
rousseldenis: Could you switch to class method ?

rousseldenis (changes requested):
Code review. Seems great. Some remarks

product_packaging_barcode_constraint_per_company/tests/test_constraints.py
kevinkhao: > Could you switch to class method ?

Thank you for feedback. Done

PR #630, opened 2024-07-25
[15.0][IMP] stock_barcodes_gs1_secondary_unit: Set todo secondary qty from move…

carlosdauden:
… when all qty is available and only one move line.
If the conditions are not satisfied, the quantity is calculated on the basis of the conversion factor.

@Tecnativa TT50148
ping @sergio-teruel @chienandalu


sergio-teruel (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #610, opened 2024-05-08
[16.0][MIG] stock_barcodes: Migration to 16.0

chienandalu:
Followup to #587

- Drop scan read logs feature, which use was deprecated already in the
  previous versions.
- Pass bootstrap migration scripts.
- Fix views issues.
- Other code fixes.

cc @Tecnativa

@FranzPoize I've been doing the migration with @sergio-teruel, who's the main developer of the module in the spanish OCA days. We started from your migration head.

At this point the module is completely functional, although we're still tweaking the views to fit the previous state.

PR #634, opened 2024-08-01
[16.0][MIG] stock_barcodes_gs1: Migration to version 16.0

sergio-teruel:
cc @Tecnativa TT47630

ping @carlosdauden @CarlosRoca13 

Depends on:
- [x] https://github.com/OCA/stock-logistics-barcode/pull/610


pedrobaeza:
You can add the temporary reference to the dependent PR.

/ocabot migration stock_barcodes_gs1

sergio-teruel:
Done

pedrobaeza:
I don't see the temporary commit:

https://github.com/OCA/maintainer-tools/wiki/Use-temporary-reference(s)-to-another-pull-request(s)

Please also squash administrative commits:

https://github.com/OCA/maintainer-tools/wiki/Merge-commits-in-pull-requests

alexis-via (approved):

stock_barcodes_gs1/models/barcode_nomenclature.py
```diff
@@ -7,5 +7,6 @@ class BarcodeNomenclature(models.Model):
     def parse_gs1_rule_pattern(self, match, rule):
         # Allow use weight ai as units directly for products with unit category uom
         result = super().parse_gs1_rule_pattern(match, rule)
-        result["use_weight_as_unit"] = True if rule.use_weight_as_unit else False
+        if result:
+            result["use_weight_as_unit"] = True if rule.use_weight_as_unit else False
```
dreispt, line 11: result["use_weight_as_unit"] = bool(rule.use_weight_as_unit)

dreispt (approved):

stock_barcodes_gs1/tests/test_stock_barcodes_gs1.py
```diff
@@ -0,0 +1,284 @@
...
+    # pylint: disable=W8121
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+
+        gs1_nomenclature = cls.env.ref(
+            "barcodes_gs1_nomenclature.default_gs1_nomenclature"
+        )
+        # Odoo creash with default separator (Alt029|#|\x1D) so remove it from settings
+        gs1_nomenclature.gs1_separator_fnc1 = False
```
kluna1998, line 21: I think you can't initialize the `gs1_separator_fnc1` field as `False` because it's a `Char` field. I think initializing it with an empty string (`""`) should work.

alexis-via:
I just found a bug in stock_barcodes / stock_barcodes_gs1 which is fully reproductible: when you receive a product tracked by serial number with a GS1 barcode such as (02)product_code(21)serialnumber **in multiple quantities**:
1) in the scan interface: when you scan the GS1 barcode of the first product with serial SN001=> works well, no error
2) when you scan the GS1 barcode of the second product (same product_id as the first scan with a different serial SN002), you get a red banner "SN002 (S/N Already in picking)"

When you analyse the code: the error is located here: https://github.com/OCA/stock-logistics-barcode/blob/16.0/stock_barcodes/wizard/stock_barcodes_read_picking.py#L533
When you see the condition that leads to the error, you see that it uses "sml.lot_id == self.lot_id or sml.lot_name == self.lot_id.name". When you scan the second barcode, self.lot_id still points to the first serial.
When you look at the code of stock_barcode_gs1, on that line of code https://github.com/Tecnativa/stock-logistics-barcode/blob/16.0-mig-stock_barcodes_gs1/stock_barcodes_gs1/wizard/stock_barcodes_read.py#L59 you see that the code sets "self.lot_name", but not "self.lot_id".

I feel that the code of stock_barcodes considers that, when entering the method _process_stock_move_line(), self.lot_id should be set to the scanned badcode, whereas the code of stock_barcode_gs1 considers that it just has to set self.lot_name. Or there is a problem linked to the fact that self.lot_id is set later in the execution order. I'm not an expert of the code of stock_barcodes, so my analysis may not be the right one.

alexis-via:
To fix the bug on products tracked by serial when receiving multiple quantities of the same product, I changed the code in stock_barcodes AND stock_barcodes_gs1:
- my changes in stock_barcodes: https://pastecode.io/s/7mgvz4h0
- my changes in stock_barcodes_gs1: https://pastecode.io/s/0bjh3c9j

The idea behind the change in stock_barcodes_gs1 is to reset self.lot_id when reading the GS1 barcode and setting self.lot_name, so that self.lot_id doesn't point to the previously scanned barcode any more.
In stock_barcodes, I made 2 changes:
- use self.lot_name instead of self.lot_id.name => needed to fix the bug
- use float_compare() => not needed to fix the bug, but it may avoid future bugs !

As I am not familiar with the codebase of stock_barcodes, my changes may not be the rights ones and it may break other scenarios. That's why I just put out those diffs and I didn't propose a PR.
But I use those diffs in production now, because I was obliged to fix the bug ASAP.

edescalona:
Hi @sergio-teruel , could you please correct @kluna1998  suggestion?

And check why the runboat is failing?

Otherwise, it seems fine to me. Thanks for your contribution.

edescalona (approved):

sergio-teruel:
Hi... 

I will review this week sorry!!!

ivs-cetmix:
Hi @sergio-teruel thank you for your contribution! any chances to fix the tests?

sergio-teruel:
Hi @ivs-cetmix Not Yet!!! Sorry

stock_barcodes_gs1/wizard/stock_barcodes_read.py
```diff
@@ -0,0 +1,192 @@
...
+        return False
+
+    def process_barcode(self, barcode):
+        nomenclature = self.env.company.nomenclature_id.filtered(
+            "is_gs1_nomenclature"
+        ) or self.env.ref("barcodes_gs1_nomenclature.default_gs1_nomenclature")
+        gs1_list = None
+        if nomenclature.gs1_separator_fnc1 in barcode or not self._ean_barcode_valid(
+            barcode
+        ):
```
AungKoKoLin1997, line 149: It crashes on any scan when the nomenclature has no FNC1 separator: gs1_separator_fnc1 is a Char, so empty means False, and False in barcode raises `TypeError: 'in <string>' requires string as left operand, not bool`.
The module's own test fails for this reason:` setUpClass sets gs1_separator_fnc1 = False`.


```suggestion
        if (
            nomenclature.gs1_separator_fnc1
            and nomenclature.gs1_separator_fnc1 in barcode
        ) or not self._ean_barcode_valid(barcode):
```

PR #628, opened 2024-07-05
[15.0][IMP] stock_barcodes: Add pending move with difference between demand and reserved

carlosdauden:
@Tecnativa TT49738

PR #642, opened 2024-08-30
[15.0][IMP] stock_barcodes_*: Move todo record methods to picking read wizard

sergio-teruel:
cc @Tecnativa 

ping @carlosdauden


carlosdauden:
Included in https://github.com/OCA/stock-logistics-barcode/pull/628

PR #651, opened 2024-10-10
[FIX] barcode_generator_abstract: do not compute a new barcode_base, if the item has already a barcode_base defined.

legalsylvain:
alternative to #619 with better coverage.

- add a comment to explicit how generate_base() works.
- adapt test.

### DIFF
--- a/barcodes_generator_abstract/models/barcode_generate_mixin.py
+++ b/barcodes_generator_abstract/models/barcode_generate_mixin.py
@@ -37,10 +37,8 @@ class BarcodeGenerateMixin(models.AbstractModel):
         for rec in records:
             rule = rec.barcode_rule_id
             if rule and rule.generate_automate and rule.generate_type == "sequence":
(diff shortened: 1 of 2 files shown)


dalonsofl:
Hi @legalsylvain !

There is a problem with these changes, 

image

Selección_036

Selección_037

(images from runboat)

If the barcode rule is changed, the previously generated barcode disappears and a new one is created, it does not keep the existing ones.

It happens because the if that checks if a barcode already exists has been deleted, could you please check it?

Lines:
- https://github.com/OCA/stock-logistics-barcode/pull/651/files#diff-f50419781cc0e7e9172d09cf83bef00972791e23a60c5d8579585367c4453301L42
- https://github.com/OCA/stock-logistics-barcode/pull/651/files#diff-f50419781cc0e7e9172d09cf83bef00972791e23a60c5d8579585367c4453301L56

Thanks!

PR #654, opened 2024-11-05
[16.0][ADD] delivery_package_gs1_barcode_report

thomaspaulb:
This module creates a label report that can be printed via an outgoing stock.picking and contains several GS1 barcodes describing the content of the package - weight, no of items, SSCC tracking code if applicable.

It looks like this:

image


delivery_package_gs1_barcode_report/readme/USAGE.rst
```diff
@@ -0,0 +1,4 @@
+#. Create a new transfer with a product with a serial number.
+#. Assign packages to the individual products in the transfer
+#. On the transfer, click Print SSCC barcode label.
+
```
thomaspaulb, line 4: During testing I noted that in order to be able to use this module, the Odoo setting of Settings > Operations > Packages needs to be enabled. Maybe you can write this in the installation instructions.

delivery_package_gs1_barcode_report/readme/USAGE.rst
```diff
@@ -0,0 +1,4 @@
+#. Create a new transfer with a product with a serial number.
+#. Assign packages to the individual products in the transfer
+#. On the transfer, click Print SSCC barcode label.
```
thomaspaulb, line 3: Add: a PDF report will show with GS1 barcodes describing the weight, SSCC code, and the amount of items in the package.

delivery_package_gs1_barcode_report/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+Print a label for outgoing stock packages, including a GS1 barcode.
```
thomaspaulb: Actually it's multiple GS1 barcodes - describe which ones are included and what they are for.

delivery_package_gs1_barcode_report/data/ir_sequence.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+
+    <record id="stock.seq_quant_package" model="ir.sequence">
+        <field name="use_python_code">True</field>
+        <field
+            name="python_code"
+        >"387193340307" + number_padded + str(get_barcode_check_digit("387193340307" + number_padded + "x"))</field>
```
thomaspaulb: Here, you are specifying a code that's specific to (probably) the person that you made it for. But, people need to be able to choose it for themselves.

I suggest to split up the module in two parts.

1. A module `sequence_sscc` that contains this part and of which the main use is to help people to automatically assign SSCC codes to a package, from a certain range that they have bought.
2. The current module, which includes the package number in the barcodes IF it is a valid SSCC code, and otherwise just omit the barcode.

--> In this way, you will have the both separated and people who for example input their SSCC code manually, can use the module without this functionality of auto generating. Also, you will lose the dependency on `sequence_python` and the module will become easier to review and to use.

delivery_package_gs1_barcode_report/reports/stock_report_views.xml
```diff
@@ -0,0 +1,27 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+
+        <record id="action_report_picking_packages" model="ir.actions.report">
+            <field name="name">SSCC Labels</field>
```
thomaspaulb: The name of this report should be something like "Package label including GS1 barcodes" because there's not only an SSCC code on there but also other GS1 barcodes that specify the weight etc.

delivery_package_gs1_barcode_report/reports/stock_report_views.xml
```diff
@@ -0,0 +1,27 @@
...
+            <field name="model">stock.picking</field>
+            <field name="report_type">qweb-pdf</field>
+            <field
+            name="report_name"
```
thomaspaulb, line 9: Why is this a new report and did you not add the barcodes to the regular Package label PDF.

delivery_package_gs1_barcode_report/reports/stock_report_views.xml
thomaspaulb: Made it into GS1 package label.

delivery_package_gs1_barcode_report/reports/stock_report_views.xml
thomaspaulb: Not addressing this one for now, the module is now usable.

PR #657, opened 2024-11-25
[16.0][FIX] stock_barcodes: play error sound when requested

alexis-via:
Without this fix, stock_barcodes v16 would always play the bell sound even in case of error.


alexis-via:
@sergio-teruel A small fix following your v16 migration of this great module

legalsylvain (approved):

chienandalu (approved):
LGTM :+1: Thanks :)

PR #660, opened 2024-11-29
[FIX] stock_barcodes: Move qty_available field from read base to read picking

sergio-teruel:
cc @Tecnativa TT51889


sergio-teruel:
Close this PR because I have added it directly in this PR https://github.com/OCA/stock-logistics-barcode/pull/628

PR #663, opened 2024-12-09
[16.0][IMP] stock_barcodes: use trash icons instead of recycle icons in inventory

alexis-via:
BEFORE

stock_barcodes_before

AFTER

stock_barcodes-after

### DIFF
--- a/stock_barcodes/wizard/stock_barcodes_read_inventory_views.xml
+++ b/stock_barcodes/wizard/stock_barcodes_read_inventory_views.xml
@@ -84,7 +84,7 @@
                                 name="action_barcode_inventory_quant_unlink"
                                 type="object"
                                 class="btn"
-                                icon="fa-recycle"
+                                icon="fa-trash"
                                 title="Remove quant"
                                 context="{'wiz_barcode_id': parent.id}"
                             />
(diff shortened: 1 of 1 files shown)


sergio-teruel:
Well!! I not used this icon because the quant is not really deleted, only set to zero...
Maybe it could be changed..

@carlosdauden, what do you think about it?

rvalyi (dismissed):

legalsylvain:
Agree with alexis, that the current "recycle" icon doesn't make sense.

> Well!! I not used this icon because the quant is not really deleted, only set to zero...

Indeed. In odoo, trash is for deletion.

if the meaning is "reset" something like this icone is maybe more suitable : https://fontawesome.com/v6/icons/rotate-left?f=classic&s=solid

alexis-via:
@sergio-teruel Your remark is true from a technical point of view. My PR proposes to adopt the user's point of view: the user scans something by mistake and wants to cancel : with a trash icon, he removes his scan and then can scan again.

All the users that I trained on stock_barcodes told me that the recycle icon was not intuitive.

sergio-teruel:
hi @alexis-via Ok.. perfect!!

sergio-teruel (approved):

carlosdauden (approved):

carlosdauden:
Personally I like the icon proposed by @legalsylvain better, but what we all agree on is that the fa-recycle is not appropriate. :wink:

pedrobaeza:
Then switch to the other icon before merging, Alexis.

carlosdauden:
Merge with fa-trash to be closer to the user's approach

pedrobaeza:
OK, I merge as is:

/ocabot merge minor

PR #664, opened 2024-12-13
[14.0] stock_barcodes: enable filter on products for inventory

alexis-via:
Before this PR:

When working on an inventory filtered on products, there is no error message when a users scans a product which is not part of the product filter and a new inventory line is created (in the backoffice, it is not possible to create a new inventory line for a product which is not part of the product filter due to a domain, cf https://github.com/odoo/odoo/blob/14.0/addons/stock/models/stock_inventory.py#L342)

With this PR: 
When you scan a product which is not part of the product filter of the inventory, you get an error sound and this message:

error_bad_product


alexis-via:
Anyone disagrees with this fix ?

PR #695, opened 2025-07-22
[17.0][FIX] stock_picking_product_barcode_report: Fix parameter name in barcode generation (type → barcode_type)

Anxo82:
Before the fix, barcode images were not rendered correctly in reports when using the gs1_128 format. This was caused by a mismatch between the parameter name used in the report template (type) and the one expected by the controller (barcode_type):
image

This commit updates the QWeb template to use barcode_type consistently and modifies the controller to accept the correct parameter. 

@HaraldPanten @Jaimermaccione 

[T-8006]

### DIFF
--- a/stock_picking_product_barcode_report/controllers/main.py
+++ b/stock_picking_product_barcode_report/controllers/main.py
@@ -13,14 +13,14 @@ class ReportController(ReportController):
     @http.route()
     def report_barcode(
         self,
-        type,  # pylint: disable=redefined-builtin
+        barcode_type,
         value,
         width=600,
         height=100,
         humanreadable=0,
         quiet=1,
     ):
-        if type == "gs1_128":
+        if barcode_type == "gs1_128":
             Gs1_128 = barcode.get_barcode_class("gs1_128")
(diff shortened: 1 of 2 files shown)


ValentinVinagre (approved):
LGTM 👍🏻

ValentinVinagre:
@pedrobaeza

stock_picking_product_barcode_report/controllers/main.py
```diff
@@ -13,14 +13,14 @@ class ReportController(ReportController):
     @http.route()
     def report_barcode(
         self,
-        type,  # pylint: disable=redefined-builtin
+        barcode_type,  # pylint: disable=redefined-builtin
```
pedrobaeza: You can remove the pylint disable.

Anxo82: Done! Thanks

pedrobaeza (approved):
/ocabot merge patch

ValentinVinagre:
@Anxo82 Could you move it to the next versions if necessary? Thanks 😸

PR #696, opened 2025-07-22
[16.0][FIX] stock_picking_product_barcode_report: Fix parameter name in barcode generation (type → barcode_type)

Anxo82:
Before the fix, barcode images were not rendered correctly in reports when using the gs1_128 format. This was caused by a mismatch between the parameter name used in the report template (type) and the one expected by the controller (barcode_type):
image

This commit updates the QWeb template to use barcode_type consistently and modifies the controller to accept the correct parameter.

Cherry-picked from commit c8aff03b49aaa670724c6b266e106fc5176b505e (17.0).

@ValentinVinagre @HaraldPanten @Jaimermaccione

### DIFF
--- a/stock_picking_product_barcode_report/controllers/main.py
+++ b/stock_picking_product_barcode_report/controllers/main.py
@@ -13,14 +13,14 @@ class ReportController(ReportController):
     @http.route()
     def report_barcode(
         self,
-        type,  # pylint: disable=redefined-builtin
+        barcode_type,
         value,
         width=600,
         height=100,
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):
Same problem:

https://github.com/odoo/odoo/blob/9728a1754424b9e5bfa93ec5d53a5179fa9d7d05/addons/web/controllers/report.py#L56

ValentinVinagre (approved):

rvalyi (approved):

PR #697, opened 2025-08-11
[15.0][IMP] stock_picking_product_barcode_report: Lang has no sense for custom labels

CarlosRoca13:
cc @Tecnativa TT57599

ping @sergio-teruel @carlosdauden


sergio-teruel (approved):

pedrobaeza:
pre-commit is failing, and why the lang is not going to apply to that part?

CarlosRoca13:
It doesn’t make sense because these are labels that are written at the moment, to be printed with the content set.
The problem with pre-commit is related to this error, which I don’t know how to solve: `ModuleNotFoundError: No module named '_black_version'.`

pedrobaeza (approved):
OK, understood. Let's see if that was a temporary problem:

/ocabot merge patch

PR #702, opened 2025-08-28
[15.0][FIX] stock_picking_product_barcode_report: Error printing from quants…

carlosdauden:
… because doc model is interpreted as stock.picking.line.print
Seems that first ir.action.report definition set the model for all report_file calls

@Tecnativa

Ping @carlos-lopez-tecnativa @CarlosRoca13


carlos-lopez-tecnativa (approved):
LGTM

CarlosRoca13 (approved):

Issue #704, opened 2025-08-29
[16.0] stock_barcodes: README screenshots


houzefa-abba:
README screenshots fail to show up:

* on github: https://github.com/OCA/stock-logistics-barcode/tree/16.0/stock_barcodes
  
<img width="577" height="285" alt="Image" src="https://github.com/user-attachments/assets/cd4948ce-81d7-407f-a999-e0cd08b8652c" />

* in app description within odoo > Apps > Stock Barcodes
  
<img width="455" height="235" alt="Image" src="https://github.com/user-attachments/assets/8a37146a-ec5b-4de6-b6ed-c69e84128458" />

cc @edescalona

edescalona:
Hi @houzefa-abba , corrected in PR https://github.com/OCA/stock-logistics-barcode/pull/706, thanks for your comments.

Issue #703, opened 2025-08-29
[16.0] stock_barcodes: Multi-picking validation fails


houzefa-abba:
Hi `stock.picking::button_validate` is multi-callable but crashes on https://github.com/OCA/stock-logistics-barcode/pull/668/files#diff-1579de390be71a744b4cbafaf849b6b6869c2932324669eb840c177a3c4002c0R68-R69 (`if self.state` check) when called with multiple records at once.

@edescalona Could you look into this please?
Thanks

**Repro steps**

There are various ways to reproduce, here is a simple one:

* Transfer list (stock.picking) > Create 2 transfers with:
    * Operation Type = YourCompany: Receipts
    * 1 product [E-COM06] Corner Desk Right Sit
* Both transfers > Button "Mark as Todo"
* Both transfers > Set a done quantity = 1, **do not validate**
* Transfer list > Select both transfers > Action > Validate

edescalona:
Hi @houzefa-abba , corrected in PR https://github.com/OCA/stock-logistics-barcode/pull/706, thanks for your comments.

PR #706, opened 2025-08-29
[16.0][IMP] stock_barcodes

edescalona:
@BinhexTeam

Making changes to issues:
https://github.com/OCA/stock-logistics-barcode/issues/703
https://github.com/OCA/stock-logistics-barcode/issues/704


stock_barcodes/models/stock_picking.py
```diff
@@ -65,7 +65,10 @@ def button_validate(self):
         if res is True and self.env.context.get("show_picking_type_action_tree", False):
             res = self[:1].picking_type_id.get_action_picking_tree_ready()
 
-        if self.state == "done":
+        if (
```
pedrobaeza, line 68: Instead of this, you have to make it multi-record aware.

houzefa-abba (approved):
Thanks for quick fixes @edescalona, just tried on runboat, this does solve both issues :sunglasses: 

Some comments for your information:

* Usually we would do 2 separate PRs as these are 2 different issues, but since these stem from the same initial global PR, I guess sticking with 1 is fine too
* No need to update global README.rst & static/description in your commit, an OCA bot does it automagically :)
* If you add "Fixes #xxx" in your commit message, the commit & PR will be directly linked to related github issues

None of these are blocking IMO unless more experienced reviewers say otherwise

pedrobaeza:
OK, let's merge this then as is for now.

/ocabot merge patch

PR #710, opened 2025-08-31
[16.0][FIX] stock_barcodes: Restore default behaviour to accumulate qty_done on each scan. Added new option in options model

sergio-teruel:
cc @Tecnativa 

Please @edescalona You con not change the default behaviour as you want... always you can add an option to options group to modelling it
Maybe now, the context key 'no_increase_qty_done' is not needed.. **Can you confirm it??**

ping @carlosdauden @Christian-RB @edescalona


pedrobaeza (approved):

edescalona (approved):
LGTM

PR #711, opened 2025-09-03
[16.0][FIX] stock_barcodes: Singleton error when user validate more than one picking

sergio-teruel:
@Tecnativa

ping @edescalona @carlosdauden


stock_barcodes/models/stock_picking.py
```diff
@@ -67,10 +67,9 @@ def button_validate(self):
 
         if (
             self.env.context.get("stock_barcodes_validate_picking", False)
-            and self.state == "done"
+            and all(p.state == 'done' for p in self)
```
pedrobaeza: pre-commit:

```suggestion
            and all(p.state == "done" for p in self)
```

pedrobaeza (approved):

carlosdauden (approved):

PR #716, opened 2025-09-09
[18.0][MIG] base_gs1_barcode: Migration to version 18.0

eduezerouali-tecnativa:
cc @Tecnativa TT57625


arielbarreiros96:
Hello everyone @eduezerouali-tecnativa @pedrobaeza I've been tracking this module history and seems like in v16 was about to die (#445) being replaced by barcodes_gs1_nomenclature, is there any new reason to bring it back to life

eduezerouali-tecnativa:
Hi @arielbarreiros96, you are right since v16 we use  `barcodes_gs1_nomenclature`. There is no reason to bring it back. Was my fault, never work with barcodes and i did not research enough. I will close the PR and thank you for your comment.

PR #718, opened 2025-09-16
[FIX] stock-logistics-barcode: on the standard super, we shouldn't execute extra code

etobella:
If we do a partial delivery, it cannot be validated.

@edescalona @Christian-RB

### DIFF
--- a/stock_barcodes/models/stock_picking.py
+++ b/stock_barcodes/models/stock_picking.py
@@ -56,11 +56,6 @@ class StockPicking(models.Model):
                 StockPicking, self.with_context(skip_backorder=True)
             ).button_validate()
         else:
-            pickings_to_backorder = self._check_backorder()
-            if pickings_to_backorder:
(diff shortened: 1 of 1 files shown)


edescalona:
Hi @etobella , why don't you think it's appropriate to show the partial delivery wizard when validating the picking?

Perhaps through configuration, allowing the user to decide?

However, the [17.0] stock barcodes module is updated with the changes from [16.0] stock_barcodes, with the latest improvements from @sergio-teruel

etobella:
Create a picking and try to make a partial delivery. It will not work. That is handled in the super. Test it please

edescalona:
Sorry @etobella  I was at 17.0, thanks for the review

edescalona (approved):
LGTM

sergio-teruel (approved):
Thanks!!

pedrobaeza (approved):
/ocabot merge patch

PR #725, opened 2025-10-06
[18.0][MIG] stock_barcodes: Migration to v18

sergio-teruel:
cc @Tecnativa 

ping @carlosdauden 

cc @Christian-RB @florian-dacosta

PR #729, opened 2025-10-13
[18.0][MIG] stock_barcodes_gs1: Migration to 18.0

sergio-teruel:
@Tecnativa

Depends on:

- [ ] https://github.com/OCA/stock-logistics-barcode/pull/725

cc @Christian-RB @florian-dacosta


stock_barcodes_gs1/wizard/stock_barcodes_read.py
```diff
@@ -0,0 +1,196 @@
...
+        nomenclature = self.env.company.nomenclature_id.filtered(
+            "is_gs1_nomenclature"
+        ) or self.env.ref("barcodes_gs1_nomenclature.default_gs1_nomenclature")
+        gs1_list = None
+        if (
+            nomenclature.gs1_separator_fnc1
+            and nomenclature.gs1_separator_fnc1 in barcode
+            or not self._ean_barcode_valid(barcode)
+        ):
+            gs1_list = nomenclature.parse_barcode(barcode)
```
deeniiz, line 154: `parse_barcode()` can also raise (`ValidationError`/`ValueError`) when a non-GS1 code happens to match a date rule (e.g. `1799999912`), so the user gets an error dialog instead of falling back to `super()`. Shouldn't we wrap it like core does in `_preprocess_gs1_search_args()`?

PR #730, opened 2025-10-13
[18.0][MIG] stock_barcodes_gs1_expiry: Migration to 18.0

sergio-teruel:
@Tecnativa

Depends on 

- [ ] https://github.com/OCA/stock-logistics-barcode/pull/729

cc @Christian-RB @florian-dacosta


stock_barcodes_gs1_expiry/wizard/stock_barcodes_read.py
```diff
@@ -0,0 +1,84 @@
...
+        return super().action_clean_values()
+
+    def process_lot_before_done(self):
+        if (
+            self.use_date
+            or self.expiration_date
+            or self.env.context.get("force_create_lot")
+        ):
+            if (
+                not self.lot_id
```
deeniiz, line 76: After a scan that creates a lot, `lot_id` remains set on the wizard. When the next scanned lot does not exist yet, it is never created and its quantity is assigned to the previous lot
```suggestion
                  (not self.lot_id or self.lot_id.name != self.lot_name)
  ```

deeniiz (changes requested):

PR #733, opened 2025-11-23
[16.0][FIX] stock_barcodes: fix opening client action permissions

giarve:
permission error when reading client action if user not holding admin permissions

### DIFF
--- a/stock_barcodes/wizard/stock_barcodes_read.py
+++ b/stock_barcodes/wizard/stock_barcodes_read.py
@@ -688,12 +688,14 @@ class WizStockBarcodesRead(models.AbstractModel):
 
     def open_actions(self):
         self.display_menu = True
-        return self.env.ref(
+        return self.env["ir.actions.actions"]._for_xml_id(
             "stock_barcodes.action_stock_barcodes_action_client"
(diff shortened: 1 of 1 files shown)


stock_barcodes/wizard/stock_barcodes_read.py
```diff
@@ -688,12 +688,14 @@ def reset_qty(self):
...
     def open_actions(self):
         self.display_menu = True
-        return self.env.ref(
-            "stock_barcodes.action_stock_barcodes_action_client"
-        ).read()[0]
+        return (
+            self.env.ref("stock_barcodes.action_stock_barcodes_action_client")
+            .sudo()
+            .read()[0]
+        )
```
pedrobaeza: Instead of this, use `_for_xml_id`, that is specially done for these cases:

```suggestion
        return self.env["ir.actions.act_window"]._for_xml_id(
            "stock_barcodes.action_stock_barcodes_action_client"
        )
```

and the same for the other.

pedrobaeza (approved):
Thank you.

Clear patch, so fast-tracking it:

/ocabot merge patch

Issue #736, opened 2025-12-16
[16.0] stock_barcodes: many warnings


alexis-via:
When you install the module stock_barcodes on odoo 16, there are several warnings:

```
2025-12-16 23:42:44,180 112156 WARNING o6_barcodes odoo.addons.base.models.ir_model: Two fields (candidate_picking_id, picking_id) of wiz.stock.barcodes.read.picking() have the same label: Picking. [Modules: stock_barcodes and stock_barcodes]
```

```
2025-12-16 23:42:44,308 112156 INFO o6_barcodes odoo.modules.loading: loading stock_barcodes/wizard/stock_barcodes_read_views.xml 
2025-12-16 23:42:44,316 112156 WARNING o6_barcodes odoo.addons.base.models.ir_ui_view: A <span> with fa class (fa fa-barcode fa-2x mt-2 mx-3) must have title in its tag, parents, descendants or have text
View error context:
{'file': '/home/alexis/new_boite/dev/16/stock-logistics-barcode/stock_barcodes/wizard/stock_barcodes_read_views.xml',
 'line': 11,
 'name': 'stock.barcodes.read.form',
 'view': ir.ui.view(641,),
 'view.model': 'wiz.stock.barcodes.read',
 'view.parent': ir.ui.view(),
 'xmlid': 'view_stock_barcodes_read_form'} 
2025-12-16 23:42:44,323 112156 WARNING o6_barcodes odoo.addons.base.models.ir_ui_view: A <span> with fa class (fa fa-barcode fa-2x mt-2 mx-3) must have title in its tag, parents, descendants or have text
View error context:
{'file': '/home/alexis/new_boite/dev/16/stock-logistics-barcode/stock_barcodes/wizard/stock_barcodes_read_views.xml',
 'line': 10,
 'name': 'stock.barcodes.read.packaging.form',
 'view': ir.ui.view(642,),
 'view.model': 'wiz.stock.barcodes.read',
 'view.parent': ir.ui.view(641,),
 'xmlid': 'view_stock_barcodes_read_packaging_form'}
```

```
2025-12-16 23:42:44,334 112156 INFO o6_barcodes odoo.modules.loading: loading stock_barcodes/wizard/stock_barcodes_read_picking_views.xml 
[...]

PR #737, opened 2025-12-22
[16.0][FIX] stock_barcodes

edescalona:
@BinhexTeam

Remove warnings (https://github.com/OCA/stock-logistics-barcode/issues/736)


alexis-via (approved):

stock_barcodes/security/ir.model.access.csv
```diff
@@ -9,3 +9,4 @@ access_stock_barcodes_option,access_stock_barcodes_option,model_stock_barcodes_o
 access_stock_barcodes_read_picking,access_stock_barcodes_read_picking,model_wiz_stock_barcodes_read_picking,base.group_user,1,1,1,1
 access_wiz_stock_barcodes_read_todo,access_wiz_stock_barcodes_read_todo,model_wiz_stock_barcodes_read_todo,base.group_user,1,1,1,1
 access_wiz_stock_barcodes_read_inventory,access_wiz_stock_barcodes_read_inventory,model_wiz_stock_barcodes_read_inventory,base.group_user,1,1,1,1
+access_report_stock_barcodes_report_barcode_actions,access_report_stock_barcodes_report_barcode_actions,model_report_stock_barcodes_report_barcode_actions,base.group_user,1,1,1,1
```
pedrobaeza: This is a too broad permission. Please be a bit more specific.

edescalona:
ping @pedrobaeza

stock_barcodes/security/ir.model.access.csv
```diff
@@ -9,3 +9,5 @@ access_stock_barcodes_option,access_stock_barcodes_option,model_stock_barcodes_o
 access_stock_barcodes_read_picking,access_stock_barcodes_read_picking,model_wiz_stock_barcodes_read_picking,base.group_user,1,1,1,1
 access_wiz_stock_barcodes_read_todo,access_wiz_stock_barcodes_read_todo,model_wiz_stock_barcodes_read_todo,base.group_user,1,1,1,1
 access_wiz_stock_barcodes_read_inventory,access_wiz_stock_barcodes_read_inventory,model_wiz_stock_barcodes_read_inventory,base.group_user,1,1,1,1
+access_report_stock_barcodes_report_barcode_actions_manager,access_report_stock_barcodes_report_barcode_actions_manager,model_report_stock_barcodes_report_barcode_actions,base.group_system,1,1,1,1
```
pedrobaeza: Why aren't you using stock groups for this?

edescalona: Hi @pedrobaeza For no specific reason, just following the ones already in the initial version, but can I change them if I think it's necessary?

pedrobaeza: Well, the idea of fine tuning the permissions is to avoid leaks to users that shouldn't see that information. That's why you should adjust it to the proper ones instead going to general wildcards.

edescalona: Okay, so I'll change all the permissions

pedrobaeza: Yeah, I think it would be the best option, but if not wanting side effects or a lot of changes, you can just do it on the new ones.

edescalona: Hi @pedrobaeza , I changed all the permissions to match those of the stock module. I don't think it's a problem, and as you mentioned, it's more specific to your objective.

pedrobaeza (approved):
Thanks for the changes.

/ocabot merge patch

PR #743, opened 2026-02-23
[16.0][FIX] barcodes_generator_abstract: preserve literal N and D in barcode patterns

sergiobstoj:
When a barcode pattern contains literal N or D characters outside the `{}` placeholder (e.g. "AMZNCC0024670952....."), the previous code replaced ALL occurrences of N and D in the entire pattern with "0", corrupting the prefix.

This was because the old code first stripped braces and then did a global `.replace("N", "0").replace("D", "0")` on the full string.

Fix: use `re.sub` to only replace N and D inside `{}` groups, leaving the rest of the pattern untouched.


legalsylvain (approved):
Thanks !

Could you port this patch upstream ?

/ocabot merge patch

AdrianaSaiz (approved):
LGTM

sergiobstoj:
@legalsylvain I'll port the changes asap. I believe the conflict with pre-commit is due to problems in setuptools-odoo (.pre-commit-config)

sergiobstoj:
@legalsylvain can you check this now please? checks have been fixed!

legalsylvain:
/ocabot merge patch

thanks !

PR #748, opened 2026-02-27
[18.0][FIX] barcodes_generator_abstract: preserve literal N and D in barcode patterns

sergiobstoj:
fw-port from https://github.com/OCA/stock-logistics-barcode/pull/743


legalsylvain:
@sergiobstoj could you take a look on red CI ?

PR #751, opened 2026-04-22
[18.0][ADD] web_ir_actions_client_scan: New module to be able to generate actions to scan

CarlosRoca13:
cc @Tecnativa TT61751

ping @victoralmau @juancarlosonate-tecnativa


victoralmau (approved):

juancarlosonate-tecnativa (approved):

victoralmau:
This is ready to be merged. What do you think, @pedrobaeza ?

barcode_action_scan/readme/USAGE.md
```diff
@@ -0,0 +1,17 @@
+You can integrate the scanner into your module by defining a client action:
+
+```xml
+<record id="action_scan_example" model="ir.actions.client">
+    <field name="name">Scan Code</field>
+    <field name="tag">barcode_action_scan.scan</field>
+    <field name="res_model">exmaple_model</field>
```
pedrobaeza:
```suggestion
    <field name="res_model">example_model</field>
```

barcode_action_scan/readme/USAGE.md
```diff
@@ -0,0 +1,17 @@
...
+    <field name="params" eval="&quot;{ 'method': 'exmaple_method', 'field': 'exmaple_field' }&quot;"/>
```
pedrobaeza:
```suggestion
    <field name="params" eval="&quot;{ 'method': 'example_method', 'field': 'example_field' }&quot;"/>
```

barcode_action_scan/readme/DESCRIPTION.md
```diff
@@ -0,0 +1,3 @@
+This Odoo module adds a custom barcode scanning action within the web backend.
+Its purpose is to allow the user to scan a code (for example, using a camera
+or a scanner) and have that value automatically processed on the server.
```
pedrobaeza: I don't have it clear what `automatically processed on the server` means.

barcode_action_scan/readme/USAGE.md
```diff
@@ -0,0 +1,17 @@
...
+</record>
+```
+Then trigger it from a button:
+```xml
+<button name="%(action_scan_example)d"
+        type="action"
+        string="Scan Code"
+        class="btn-primary"/>
+```
```
pedrobaeza: If I'm understanding this correctly, after having the popup and it gets a barcode scanned, it passes such barcode to the calling model and you can intercept it through the barcode onchange, or what? Please document it.

pedrobaeza (changes requested):
Sorry for being so picky with the module names, but it's for trying to keep the consistency. In this case, due to the aim of it, the proper name would be `web_ir_actions_client_scan` or similar, in the same mood than other `web_ir_actions...` in OCA/web. This time, we can keep it in this vertical repository as being related though.

CarlosRoca13:
Changes done @pedrobaeza

web_ir_actions_client_scan/readme/USAGE.md
```diff
@@ -0,0 +1,20 @@
...
+```
+Then trigger it from a button:
+```xml
+<button name="%(action_scan_example)d"
+        type="action"
+        string="Scan Code"
+        class="btn-primary"/>
+```
+The action opens a popup where you will have the option to scan a barcode.
+After completing the scan, the method defined in the action parameters
```
pedrobaeza, line 24: Thanks for the extra information. I see that the README still spells incorrectly `example`. And another doubt, what the field parameter does? I think you should explain that parameters upper.

CarlosRoca13:
Added

web_ir_actions_client_scan/readme/USAGE.md
pedrobaeza: Sorry, I still don't understand the use of the field method. You say: "if there are two fields that can apply...", but what "apply" means. Being a method, you can write to any field you want. What is the signature of the method by the way?

victoralmau: I think we should be careful about making changes if these methods and this module are already in use with some clients. “Removing” the field might be a good idea—okay—but let's add it to the ROADMPA to do it in the future and avoid breaking everything and having to make changes to modules that already use this.

pedrobaeza: Well, I'm not trying to change anything, but to understand how it works, and anyway, it should be properly documented. But if the field is something awkward, I would change it now that it's in PR, and at the end, it's only to modify the method calls.

victoralmau: Okay... we just need to change the methods that are already using this module and `ir.actions.client` records.

CarlosRoca13:
Field parameter removed as it is not necessary

pedrobaeza:
Thanks for the changes. I'm still missing the signature of the method in the documentation, or at least a mention of the arguments. I suppose that now is `def method(self, scanned_barcode)` or similar, isn't it?

CarlosRoca13:
Added

pedrobaeza (commented):
Thank you very much!

I'm now understanding that you were previously storing the scanned value in the field indicated, and calling the method, so I consider this a much better approach because:

- It does things in one transaction, instead of one save + the call of the method.
- It doesn't require an intermediate field to store the value. You may not want to store that scanned value, but to use it for another thing.

One final thing, and sorry for being so picky, but the nomenclature should be the correct always for avoiding misunderstandings. I see that you have put as `value_scanned`, but this is not a value (or well, the word is not specific enough), but a barcode, so it should be IMO `scanned_barcode`, or `barcode_scanned`, or simply `barcode`.

CarlosRoca13:
It doesn't matter because it is a positional argument, bu I will change it in the doc

CarlosRoca13:
Changed

pedrobaeza (approved):
/ocabot merge nobump

PR #755, opened 2026-05-15
[18.0][MIG] stock_barcodes_picking_batch: Migration to 18.0 (continuation of #727)

IJOL:
## Continuation of #727

Builds on @sergio-teruel's #727 and depends on the companion PR for
`stock_barcodes` (sister PR opened from this same fork). All commits
from #727 are preserved (rebased on top of the fixed `stock_barcodes`
branch). On top of that, 1 fix commit takes the test suite from red to
**0 failed, 0 error(s) of 3 tests** with no skipped tests.

### `[18.0][FIX] stock_barcodes_picking_batch: align tests with v18 semantics`

- **test_wiz_scan_picking_batch_values**: `display_name` may be False
  on a freshly created transient wizard until `_compute_display_name`
  runs; guard the assertion (matches the sibling test in
  `stock_barcodes`).
- **test_picking_batch_wizard_scan_product**: the class-level
  `wiz_scan_picking_batch` is built in `setUpClass` BEFORE
  `picking_batch.action_assign()`, so its cached pending moves do not
  reflect the freshly reserved sml. Re-build the wizard with
  `action_barcode_scan()`, exactly the same pattern the sibling test
  `test_picking_batch_wizard_scan_more_product_than_needed` was already
  using. Add `no_increase_qty_done=True` so each scan sets qty_picked
  to 1 instead of accumulating to the move's full available_qty.
  Assert `qty_picked` instead of `quantity` (with the OCA extension
  `stock_move_line_qty_picked`, quantity is only synced at
  action_done).
- **test_picking_batch_wizard_scan_more_product_than_needed**: remove
  `barcode_option_group_out.show_pending_moves = True`; that field was
  removed during the v18 migration of stock_barcodes (see
  `migrations/18.0.1.0.0/pre-migration.py` renaming the column to
  None).
- Regenerate `static/description/index.html` via
  `oca-gen-addon-readme`.

### Pre-commit

`pre-commit run -a` is green.

cc @sergio-teruel @carlosdauden @CarlosRoca13 @flachicasidoo


carlosdauden (changes requested):
Please, if you’d like to suggest any changes, do so on the existing pull request, as we’re currently working on it and this could cause conflicts.

PR #758, opened 2026-07-09
[IMP] barcode_generator_product_variant: has_missing_barcode archive

HekkiMelody:
This commit fixes an issue where archiving or unarchiving a variant wouldn't update the `has_missing_barcodes` field on the product template and wouldn't hide or show the related button accordingly


HekkiMelody:
Pre-commit failure and unrelated test failure are both addressed in https://github.com/OCA/stock-logistics-barcode/pull/759

PR #763, opened 2026-07-14
[18.0][MIG] stock_barcodes_automatic_entry: Migration to 18.0

deeniiz:
Standard migration to 18.0.

  * In 14.0 included `FormController._barcodeActiveScanned`; since 16.0 this lives in
    the `BarcodeHandlerField` widget, so the module now patches
    `BarcodeHandlerField.prototype.onBarcodeScanned` via `@web/core/utils/patch`.
  * The automatic click waits deterministically for the scan to be fully
    processed: one turn of the record model mutex (where core queues the
    `record.update()`), then until the view buttons are re-enabled — 18.0
    disables all form buttons while a button action runs, and `click()` on a
    disabled button is silently ignored.
  * Usage example in the readme: `invisible="1"` → `d-none` .


deeniiz:
@eantones, can you please review it?

PR #766, opened 2026-08-09
[16.0][IMP] stock_barcodes: document configuration and add field help

### LINKED ISSUE #757: [stock_barcodes]: Missing usage documentation
Hello ! 

The `stock_barcodes` module currently has no usage documentation. The [README](https://github.com/OCA/stock-logistics-barcode/blob/16.0/stock_barcodes/README.rst) does not reallyexplain how to use the module after installation and how to configure it.

It would be grateful to have some documentation about the barcodes options and actions configurable in the logistic module. Actually, the helper on each field is not really meaningful and it would be easier to not go in the code understanding how each field is working.

I found out this demo video : https://www.youtube.com/watch?v=ifg4ILqbNE8, and tried to understand by myself how does the module work but it was not really helpful

Would it be possible to have some documentation ?

Thanks ;)

takahii03:
Closes #757

## Context

The reporter of #757 asked for documentation of the barcode options and actions,
and pointed out that the field helps were not meaningful enough to avoid reading
the source. This PR addresses the configuration side of that request.

`stock_barcodes` is configured entirely through data — an option group describes
which fields the interface asks for, in which order, and how it behaves while
scanning — but nothing in the module explained that.

* The module had no `CONFIGURE` readme fragment, so the generated README jumped
  straight from the installation notes to usage, with no description of the
  option groups, the "Steps to scan" list or the barcode actions.
* The configuration models carried almost no field help. `stock.barcodes.option`
  had none at all on its nine fields, so understanding what *Step*, *Sequence*,
  *To scan*, *Forced* or *Filled default* actually do required reading
  `process_barcode()` in the wizard.

## What this PR does

**`readme/CONFIGURE.rst` (new)**

* Where the configuration lives: *Barcode options*, the two option-group fields
  of an operation type, and the location barcode field this module adds back.
* The six option groups installed as default data, with their code and intended
  use, plus a note that they are `noupdate="1"` and that copying a group is
  safer than editing a default one.
* The meaning of each column of the *Steps to scan* list, including how the
  current step is recomputed after each scan and why the sequence inside a step
  matters, illustrated with the *Picking OUT options* group.
* The behavior settings of an option group that are worth knowing before a first
  configuration, and the barcode actions of the *Barcodes* menu.

**Field help**

Added help to the fields of `stock.barcodes.option`, `stock.barcodes.option.group`
and `stock.barcodes.action` that had none. The help of the option-group `code`
field states explicitly that it is not only a label — `IN`, `OUT` and `REL` are
compared in the wizard to change behavior — since renaming it silently breaks
those branches.

**`readme/USAGE.rst` (fix)**

The items of the two "Actions" blocks used `#` as a list marker instead of the
`#.` expected by reStructuredText, so they rendered as plain paragraphs starting
with a hash instead of numbered steps. The embedded images were also indented
one column short of the item body, which closed the list and turned each of them
into a block quote, restarting the numbering at one on every following item.

## Notes for reviewers

Three fields are left without help **on purpose**, because nothing reads them and
their meaning could not be documented without guessing:

* `stock.barcodes.option.message`
* `stock.barcodes.action.key_shortcut` — neither set nor read anywhere
* `stock.barcodes.action.key_char_shortcut` — the value the kanban template sets
  from it is never used, and the shortcuts of the interface are hard coded

No behavior change: this PR only touches help strings and readme fragments.

### DIFF
--- a/stock_barcodes/models/stock_barcodes_action.py
+++ b/stock_barcodes/models/stock_barcodes_action.py
@@ -21,15 +21,32 @@ class StockBarcodesAction(models.Model):
 
     name = fields.Char(translate=True)
     active = fields.Boolean(default=True)
-    sequence = fields.Integer(default=100)
-    action_window_id = fields.Many2one(
-        comodel_name="ir.actions.act_window", string="Action window"
+    sequence = fields.Integer(
+        default=100, help="Order of the tile in the Barcodes menu"
     )
-    context = fields.Char()
+    action_window_id = fields.Many2one(
+        comodel_name="ir.actions.act_window",
+        string="Action window",
+        help="Window action executed when the tile is pressed or its barcode "
+        "is scanned",
+    )
+    context = fields.Char(
+        help="Optional context added to the window action, e.g. "
+        "{'search_default_code': 'incoming'} to open it with a filter of its "
+        "search view already applied.\n"
+        "It must not start or end with a space",
+    )
     key_shortcut = fields.Integer()
     key_char_shortcut = fields.Char()
-    icon_class = fields.Char()
-    barcode = fields.Char()
+    icon_class = fields.Char(
+        help="CSS class of the icon shown on the tile, e.g. a Font Awesome "
+        "class such as fa-truck"
+    )
+    barcode = fields.Char(
+        help="Barcode that triggers this action when scanned from the barcode "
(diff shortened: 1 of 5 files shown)


pcastelovigo (approved):
thank you very much

PR #769, opened 2026-09-15
[16.0][FIX] stock_barcodes: count badge domain, and package visibility on the inventory wizard

AungKoKoLin1997:
## The problem

The count badge on the barcode action kanban is meant to show how many records the
action's context would select. It never filters anything: `_count_elements()` always
returns the **total** number of records of the destination model, whatever the context.

With the default actions, Picking IN, Picking OUT and Picking INTERNAL all point at
`stock.picking.type` with different `search_default_code` values, so all three show the
same badge: the total number of operation types.

Two independent defects both empty the domain, which is why the first one alone is not
enough to explain the behaviour.

**1. The field check is made against a string.** `action_window_id.res_model` is a `Char`
holding the model *name*, not the model:

```python
if hasattr(
    self.action_window_id.res_model,
    FIELDS_NAME.get(field_name, field_name),
):
```

`hasattr("stock.picking.type", "code")` is always `False`, so every leaf is turned into
`False` and discarded.

**2. A guard that can never hold.** Even a leaf that survived step 1 is dropped here:

```python
search_count = (
    list(filter(lambda x: x, domain))
    if all(val_d is True for val_d in domain)
    else []
)
```

`domain` holds tuples and empty tuples, never the literal `True`, so `all(... is True ...)`
is `False` as soon as there is a leaf, and `search_count` is reset to `[]`.

`search_count([])` then counts every record of the destination model.

## Why the existing test did not catch this

The test asserted the broken behaviour, and passed by coincidence:

```python
self.barcode_action_valid.context = "{'search_default_code': 1}"
self.assertEqual(self.barcode_action_valid._count_elements(), 6)
```

That `6` was never "operation types whose code is 1". It is `search_count([])` on
`stock.picking.type` — every operation type that exists — and a minimal install happens to
create exactly six on `stock.warehouse0`. Two unrelated numbers that coincide.

The giveaway is that the assertion is insensitive to its own input: set the context to
`search_default_code: 'incoming'`, to a filter name, or to a key naming no field at all,
and the result is still `6`. It asserted a fact about the database, not about the method.

This also explains why CI is green on this repo but the test fails elsewhere. The quantity
actually asserted grows whenever any other module adds an operation type (`mrp`,
subcontracting, dropshipping...). On this repo only the local dependencies are installed,
so it sits at 6 indefinitely; on a deployment with more modules installed the same
unmodified code gives 12, and the test fails with `12 != 6`.

So the test has to change together with the fix: it is pinned to the behaviour being
fixed. `search_default_code: 1` matches no operation type, so the expected count becomes
`0`. On its own that would leave the valid action asserting the same `0` as an action with
no `action_window_id` at all, so a second assertion counts `search_default_code: 'incoming'`
against the operation types that code selects. Counting everything and counting nothing
both fail it, and it does not depend on how many operation types the installed modules
create.

## The fix

- Look the field up in the destination model's `_fields` instead of using `hasattr()` on
  the model name.
- Drop the `all(val_d is True ...)` guard so the built leaves survive.
- A key listed in `FIELDS_NAME` names a **search view filter**, not a field. In
  `{'search_default_barcode_options': 1}` the `1` is the flag enabling the
  `barcode_options` filter, not an id, so the leaf must mirror that filter's domain
  (`barcode_option_group_id != False`) rather than compare the field to the flag.

## Checking the result

On a database where one incoming operation type has a barcode option group, before the
fix the badge showed the total count of operation types; after it, the badge matches what
the action actually opens:

```
BADGE:                   1
WHAT THE ACTION OPENS:   1
```

@qrtl QT7194

---

## Second fix: the Package field is shown when the Packages feature is disabled

The source package field appears on the inventory reading screen even when
*Inventory > Settings > Packages* (`stock.group_tracking_lot`) has never been enabled.

The base wizard view guards the group correctly
(`stock_barcodes/wizard/stock_barcodes_read_views.xml`):

```xml
<group name="quant_package" groups="stock.group_tracking_lot" col="2">
```

The inventory wizard replaces that whole node to hide the *result* package, and
`position="replace"` swaps the node together with its attributes, so the guard goes with
it (`stock_barcodes/wizard/stock_barcodes_read_inventory_views.xml`):

```xml
<group name="quant_package" position="replace">
    <div class="mt4">
        <strong class="d-none d-sm-block">Package</strong>
        ...
```

Putting `groups` back on the replacement node restores the original behaviour: Odoo's view
postprocessing removes any node carrying a `groups` attribute for users outside the group,
so nothing changes for installations that do use packages.

The Python side was already safe — `process_barcode_package_id()` and
`process_barcode_result_package_id()` both return `False` when the group is missing, so a
scan could never resolve a package. This is display-only.


stock_barcodes/models/stock_barcodes_action.py
```diff
@@ -117,11 +117,7 @@ def _map_context_values(x):
                     map(lambda x: _map_context_values(x), context_values)
                 )
             ]
-            search_count = (
-                list(filter(lambda x: x, domain))
-                if all(val_d is True for val_d in domain)
-                else []
-            )
+            search_count = list(filter(lambda x: x, domain))
```
nobuQuartile, line 120:
```suggestion
            search_count = [x for x in domain if x]
```
This is a more common way, isn't it?

nobuQuartile (approved):
LGTM: Code review and functionl review
Minor point

PR #771, opened 2026-09-17
[16.0][ADD] barcodes_gtin_flexible_match: enables to match a GTIN barcode whatever the variant used

nicolas-delbovier-acsone:
This module enhances Odoo's product barcode search mechanism to seamlessly handle differences between GTIN-13 and GTIN-14 (EAN-13 with a leading zero) formats. 

### Context

In real-world logistics and retail environments, physical barcode labels or scanner hardware frequently introduce or omit leading zeros (e.g., turning a 13-digit EAN `5412345678901` into a 14-digit GTIN `05412345678901`).

cc @jbaudoux


barcodes_gtin_flexible_match/models/product_product.py
```diff
@@ -0,0 +1,60 @@
...
+    def _expand_barcode_domain_term(self, domain_term):
+        """Helper to transform barcode equality domains to 'in' queries with GTIN variants."""
+        if not (isinstance(domain_term, (list, tuple)) and len(domain_term) == 3):
+            return domain_term
+
+        field, operator, value = domain_term
+        if field != "barcode":
+            return domain_term
+
+        if operator == "=" and isinstance(value, str):
```
lmignon: @nicolas-delbovier-acsone I would also support the `like` and `ilike` operators. Otherwise, searches using the search bar in the list view will not benefit from this expansion mechanism. This probably implies an other way to extend the domain....

Image

Image

lmignon (changes requested):

barcodes_gtin_flexible_match/models/product_product.py
nicolas-delbovier-acsone: You are not searching for "barcode" in the picture above... I think this is searching on the name. I did the test on "barcode" on my end and it worked. Can you confirm? (you have to create a custom filter then)

nicolas-delbovier-acsone: I was unaware of the name search mechanism my bad, I'll fix this

jbaudoux (changes requested):
I don't like this approach at all where you mofidy the searched value. It's misleading. Nobody expects what you search for to be modified. It's like if you would modify the product search to also match product packagings having that barcode. It should be the caller of the search method to build properly what you are looking for.

Btw, I don't think your solution works for a domain like ("product.barcode", "=", barcode)]. And you should also cover product.packaging. And if you have multi barcode, you also need such.

Sorry but leave this complexity to the caller.

